PR #2, opened 2021-03-30
[RFC] RFC Process Update

hogepodge:


hsaputra:
Are we not sending the copy of topics from discuss.tvm.apache.org to dev@ mailing list anymore?

I could not find the email copy for the TVM discuss that linked with this PR: https://discuss.tvm.apache.org/t/rfc-update-rfc-process/9033

hogepodge:
It looks like the forum is being archived. I don't know why that particular topic was missed.

hsaputra:
Ah, I saw your latest comment from the TVM archiver to dev@ list. Maybe I accidentally delete the previous emails. I think we are still good. Thanks, Chris

tqchen:
cc @apache/tvm-committers

README.md
```diff
@@ -1,6 +1,64 @@
...
+  having no work done for a period of time defined by project leaders. The RFC
+  will be moved to the docs/rfc/postponed directory
+	- **Resuming an Postponed RFC**: Work on a postponed RFC may be resumed by a
+      new responsible party at any time after another discussion and pull request
+      review process to move the RFC from docs/rfc/abandoned to docs/rfc/active.
+
+## References
+[references]: #references
+
+[RFC Discussion Post](https://discuss.tvm.apache.org/t/rfc-update-rfc-process/9033)
```
tqchen, line 64: Moving forward, we need to use the docs under tha RFC as source of truth, so do not need to refer to the original discuss post.

README.md
```diff
@@ -1,6 +1,64 @@
...
+
+## RFC Workflow
+[rfc-workflow]: #rfc-workflow
+
+- **Community Discussion**: A need or issue is brought to the discussion forum.
+  During this phase, the developer and user community can discuss the need for
+  and requirements of the RFC
+- **Pull Request**: After or concurrent with the conversation on the discussion
+  forum, a pull request is created using the format prescreibed by the
+  [RFC Template](https://github.com/apache/tvm-rfcs/blob/c436b06818cb65f72cac821616670b22a2473619/0000-template.md).
```
tqchen, line 30: point to main

README.md
```diff
@@ -1,6 +1,64 @@
 # TVM RFCs
```
tqchen, line 1: Please add ASF header to the markdown, we might consider introducing similar lint checks to this repo later

tqchen (changes requested):

README.md
```diff
@@ -1,6 +1,64 @@
...
 
-This repoisitory is an evolving repo containing the new RFC process for TVM, more changes to come in next few days.
+## What is an RFC?
+[what-is-an-rfc]: #what-is-an-rfc
 
+An RFC is a “Request for Change” to the TVM project. It is a design document
```
areusch, line 6: it would be good to specify a way to identify RFCs. e.g. Python uses PEP numbers, what is our equivalent? A strawman: `RFC <discuss forum topic #>`.

areusch (changes requested):
doesn't have to be part of this change, but very supportive of adding a linter to this e.g. via GH actions or script + docker container. it would also be great to think about a policy as to how to link to things. For example, when linking to other RFCs, should links go to this repo or to the discuss forum?

Related is my comment on identifying RFCs. If RFCs can be easily identified by e.g. a number or name, then linking to them in this repo is also trivial.

README.md
jroesch: The template provides explicit RFC numbers. I think going back and making the change completely retroactively will be hard but we should bring all RFCs into the repo with explicit integer IDs and names and refer them them that way.

README.md
```diff
@@ -1,6 +1,64 @@
...
+      regardless of if they are accepted or postponed. 
+	- A successful RFC will include an overview with the problem the RFC is
+      attempting to address, a proposed solution that describes the design and
+      implementation strategy, and a timeline for completion. Optional sections can
+      include (but are not limited to) alternatives that were considered, security
+      considerations, and open problems that the RFC does not solve.
+	- It is expected that RFCs will change, as part of the feedback process and
+      as new implementation details arise. Changes to the RFC should not be squashed
+      or force pushed in order to retain change and discussion history.
+- **Tracking Issue**: Upon merging a RFC, a tracking issue will be created where
```
comaniac, line 45: If an RFC involves multiple PRs, it makes more sense to create the issue with the milestone and timeline tracking before merging the first PR.

README.md
areusch: I agree, but I think it's still possible to reference forum posts from them? though RFCs should state the rationale behind their decisions, it may still be helpful to refer to the original discussion

README.md
areusch: ah sorry, my bad. having a number here makes sense.

but, my question is: how do you pick that number? I think that would mean that someone proposing an RFC would then have to go look at all the outstanding PRs to discover $next_rfc_number?

README.md
hogepodge: My thinking was that the RFC in the repository is the source of truth, but that there's value to linking back to the original discussion. There's a balance though, I didn't link back to the community meeting. I'm happy to follow whatever consensus is.

README.md
hogepodge: Monotonically increasing by 1, regardless of if the RFC is implemented or not. Someone would have to search outstanding PRs to find the largest value. A linter could help with this.

README.md
hogepodge: Do you mean the implementation or the RFC itself? The order or operations being Merge RFC, Create Tracking Issue, PRs can be tracked in tracking issue. I'd be happy if you updated the language if there are development features and GH features we want to take advantage of.

README.md
hogepodge: Is there a link I can use as a reference?

tqchen: https://raw.githubusercontent.com/apache/tvm/main/README.md

README.md
```diff
@@ -1,6 +1,64 @@
...
+development. A secondary audience is general users and developers who are
+interested in how and why a feature was designed and implemented.
+
+## RFC Workflow
+[rfc-workflow]: #rfc-workflow
+
+- **Community Discussion**: A need or issue is brought to the discussion forum.
+  During this phase, the developer and user community can discuss the need for
+  and requirements of the RFC
+- **Pull Request**: After or concurrent with the conversation on the discussion
```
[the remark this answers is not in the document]
hogepodge: We could do a couple of things:
A) remove the phrasing "concurrent"
B) state [draft] should be added if there's discussion on the forum.

I think I prefer A to B, for sake of clarity and simplicity.

README.md
tqchen: My original thinking was mostly the other way around, the rfc doc suppose to distill everything available in the forum(so people dont have to lazily refer back), but we can for sure link to discussions when we open RFC PRs etc.

README.md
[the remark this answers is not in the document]
comaniac: I prefer A as well.

README.md
comaniac: Ah sorry I think I misunderstood it before. I supposed "Merging RFC" meant merging the first PR of the RFC. I realized that "Merging RFC" here means the RFC is committed and the author can start filing PRs. Then it makes sense to have the flow like "merging RFC" -> "file a tracking issue" -> "file PRs and update the issue".

README.md
areusch: I guess another option is that when the PR is ready to land, there should be a merge conflict if someone submitted another RFC PR since you initially opened the PR. so at merge time, you could renumber.

as Jared points out, we may need to undertake a migration process, and given the number of past RFCs it seems likely that the migration process would overlap with brand new RFCs being submitted. this means that RFC numbers may not monotonically increase with the order of their adoption. so i'm wondering if making the RFC numbers start at 1 and then having to map to discussion forum topic ID is really buying us anything. should we just adopt the topic id from the forum into the RFC ID? then it would be simple to move back and forth.

the PR ID is yet another id that may confuse ppl, but I don't see a way around that.

this also brings up a related question. now we are discussing this on the PR rather than in the forum thread. admittedly, the discussion is about a process related to the RFC repo itself....but we are still splitting the discussion. Someone who just views the RFC for RFCs thread may not see this.

Do we have policy about where discussion should happen? splitting discussion into two parts seems bad, but code-review naturally lends to discussion because of the inline-comment feature.

hogepodge: I'm flexible on numbering. The standard across many open source projects (including Rust and Kubernetes, for example), is to just increase by 1 and resolve conflicts as they arise. One problem that might come out of linking to the discussion post or pull request number is it ties the repository to technology that the project might not be using in the future.

The workflow seems to be:
* Initial discussion on Discuss, with a decision to formalize as a pull request.
* Pull request, which has more of a final review before merging.
* Upon merging, merged RFC becomes source of truth.

README.md
hogepodge: Raised the question in the discussion forum. My preference is to just link back. Links are cheap, and it saves someone from having to run a search to find old context or reference reasoning behind particular decisions.

README.md
hogepodge: Regarding old RFCs, I'm not sure there's a will to grab old RFCs and post them back up here. If we do, for documentation and other reasons, the two options for dealing with it are:

A1: Start current RFCs as some larger number, like 0100, to allow for backfilling old RFCs
A2: When adding old RFCs, add a prefix indicating that it is a legacy change, for example L0001, L0002, etc.

README.md
areusch: @tqchen I think it would be great to avoid linking to design documentation in the forum (as a goal--this will be impossible in the first few RFCs which will need to reference those posted on past forum topics). However, I think that it's still worth allowing linking to reference e.g. community discussion to understand how each design conclusion was reached. So the line would be: if you're linking to something which describes a design, that needs to be another RFC in the repo (or a past RFC on Discuss which was not yet ported). If you're just linking to additional context around a decision, or to something directly evidenced in a forum post, that's ok. However, we preserve the original goal of fully describing a design in the GH RFCs.

README.md
areusch: ok, this seems reasonable to me. I agree with prefixing old RFCs with a letter to make it clear.

hogepodge:
@areusch @tqchen @jroesch ready for final review

README.md
```diff
@@ -1,6 +1,110 @@
...
+- **Implementation**: Work will begin on the RFC, with
+  pull requests linking back to the tracking issue. Upon completion of the RFC,
+  the tracking issue will be closed and the RFC will be moved to the
+  docs/rfc/completed directory.
+- **Changes**: It is not uncommon for design changes to be required during or
+  after the initial implementation. If this is the case, the RFC should be
+  updated to reflect the change. In the instance where the change is a
+  significant addition rather than a simple modification, a new RFC should be
+  posted.
+- **Postponement**: An RFC may be postponed either
```
tqchen, line 81: This seems to be a departure from the rust RFC model https://github.com/rust-lang/rfcs. I wonder if we want to make the process more consistent with the rust RFC process. In particular, a RFC pull request itself can be postponed, but the implementation of RFC should follow a separate process. See https://github.com/rust-lang/rfcs#implementing-an-rfc

Maintaining the state of the RFC implemention inside the rfc repo can also be a maintaince overhead. The status implementation should be tracked in a tracking issue.

tqchen (changes requested):

jroesch:
I think we just need to resolve @tqchen's comment since I think the request was brought up by @areusch and we can merge.

hogepodge:
@tqchen @jroesch I updated the process to more closely match the Rust process. A few comments on it:

* I still kept it more lightweight than the Rust process. As a smaller community, I want to strike a balance between formalism and ease of application.
* One major difference is in postponement. The Rust community appears to postpone RFCs before they are merged, adding a tag. This is in contract to TVM postponement, which is meant to capture features that may have been accepted but aren't actively being developed. If this is a point of confusion, I would suggest we change the name to something else. I'm not convinced that the Rust model of postponement is necessary for a project of this size and scope.

tqchen:
I think my main concern is the need to move RFC text around depending on its implementation status. In practice it would add additional maintainace overheads and also loses the history of the RFC moditication. Per Rust RFC process, merging an RFC means that the community has get consensus on the design and desired way of improvements.  Ideally the concensus of the design should be independent from the implementation status. Ideally the path to the text should not change after the RFC has been merged.

The actual state of implementation can be tracked in somewhere else(e.g. a tracking issue) independent from the RFC repo(which reflects the consensus on the design). Alternatively, an RFC itself could contain a short status section that shows (proposed, implemented). I would be cautious marking an accepted RFC as postponed unless the design decision is superseded and no longer holds, in which case we might mark the rfc as abandoned.

An implementation of RFC can happen soon, after a few month depending on the progress, they can also happen asynchrously given that apache-style development is quite asynchrous.

hogepodge:
Ok, that clears things up for me. Thanks. @tqchen I've updated the PR to move more of the mechanics of handling the RFC into the tracking issue, and deleted the requirements for moving the RFC around.

tqchen:
Thanks @hogepodge . It looks good to me now. Now that we have close to finalization of the text, would be good to invite more community members to take a look by creating a new discuss post

hogepodge:
Doesn't look like there's much conversation around this. I'm ready to move forward on it when you think it's appropriate.

https://discuss.tvm.apache.org/t/rfc-preparing-to-launch-new-rfc-process/9838

tqchen:
cc @apache/tvm-committers final call for feedbacks for 36 hours before we merge and start the formal voting process

tqchen (approved):

PR #6, opened 2021-06-09
[RFC] [Relay] Automatic Mixed Precision Pass

AndrewZhaoLuo:
Relevant Links: 

https://discuss.tvm.apache.org/t/rfc-relay-fp32-fp16-model-support/9994
  - Old discussion before the new RFC process was rolled out

https://github.com/apache/tvm/pull/8069
  - Initial PR
  
cc @hogepodge @mbrookhart @anijain2305 @masahi 

Link to tracking issue: https://github.com/apache/tvm/issues/8296


comaniac:
Thanks for the RFC. I have two questions:
1. How to mark/set the color (i.e., attribute) of every operator?
2. It seems to me that if we register a casting checker instead of just a label (color), then we can simplify the algorithm a lot. Taking the case `A(green) - B(gray) - C(green)` as an example, if we could register a casting rule of B as follows, then we just need one traverse to know if we need cast around B:

    ```
    def amp_B(expr, args):
        a = args[0]
        if (a.dtype is float16):
          return fp16
        return fp32
    ```

    After all, we only need the previous nodes to determine 1) whether to use FP16 implementation, and 2) whether to insert casts. It seems to me that this pass is similar to the layout conversion pass, which uses one traverse to finish everything, so it might be possible for AMP too.

AndrewZhaoLuo:
> Thanks for the RFC. I have two questions:
> 
> 1. How to mark/set the color (i.e., attribute) of every operator?
> 2. It seems to me that if we register a casting checker instead of just a label (color), then we can simplify the algorithm a lot. Taking the case `A(green) - B(gray) - C(green)` as an example, if we could register a casting rule of B as follows, then we just need one traverse to know if we need cast around B:
>    ```
>    def amp_B(expr, args):
>        a = args[0]
>        if (a.dtype is float16):
>          return fp16
>        return fp32
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    After all, we only need the previous nodes to determine 1) whether to use FP16 implementation, and 2) whether to insert casts. It seems to me that this pass is similar to the layout conversion pass, which uses one traverse to finish everything, so it might be possible for AMP too.

Yep that is correct. This RFC has an initial PR here: https://github.com/apache/tvm/pull/8069.

To answer your questions:
1.  src/relay/transforms/fp32_to_fp16.h -- DefaultFP16Colorer is the default way. But the only thing we need is a callable with type CallNode*(Color). So you could write your own colorer that does arbitrary stuff when only looking at a single node at a time.

2. This is functionally what is done in the PR I link.

AndrewZhaoLuo:
> Thanks for the answers. I'll review the PR to get more implementation details.
> One more question regarding the extensibility: can this be extended easily to support bfloat16?

It should be trivial (hope I don't eat my words). I'm not 100% sure of the support for bfloat16 in current relay ops however.

AndrewZhaoLuo:
I don't know Chris Sullivan's github handle so if someone could cc him too that would be great.

tmoreau89:
CCing @csullivan

comaniac:
> > Thanks for the answers. I'll review the PR to get more implementation details.
> > One more question regarding the extensibility: can this be extended easily to support bfloat16?
> 
> It should be trivial (hope I don't eat my words). I'm not 100% sure of the support for bfloat16 in current relay ops however.

TVM has limited bfloat16 support now but it's on the way, so it would be better for this RFC to also consider this case, even the initial version may not cover it.

AndrewZhaoLuo:
So the associated PR is getting closer to a mergeable state. Is this RFC ready to merge?

tqchen:
cc @comaniac would be great if you can help shepherd this RFC

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+
+- Why is this design the best in the space of possible designs?
+
+Other alternatives require a lot more work and changes and could probably considered future goals of TVM.
+This include automatic mixed precision training.
+
+- What other designs have been considered and what is the rationale for not choosing them?
+
+We can support automatic mixed precision retraining though that is a much, much larger future goal. It's
+good to have this in the meantime.
```
comaniac, line 103: The answer to this question should come with a discussion of existing mechanisms used by other frameworks, such as XLA and PyTorch.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+
+- What is the impact of not doing this?
+
+TVM is not the best tool for making models go fast as we leave a lot of free speedup on the table.
+
+# Prior art
+[prior-art]: #prior-art
```
comaniac, line 110: It's better to also discuss the tensor cache mechanism used in PyTorch.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+
+Many machine learning models can move significant portions of their computational graphs into the FP16 space 
+without significant loss of accuracy. For many pieces of hardware this also comes with a boost in speed. In 
+the past utilizing FP16 in mixed precision training saw significant [increases in convergence speed](https://pytorch.org/blog/accelerating-training-on-nvidia-gpus-with-pytorch-automatic-mixed-precision/). 
+
+We should expect similar increases for inference. This speed increase without accuracy loss is highly desirable
+for many users.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
```
comaniac, line 33: It would be better to provide an example at the end of this section. i.e., how this pass is used and what's the result IR looks like.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+a function, which maps operation types like `conv2d` to an accumulation datatype as well as an output 
+datatype. The output datatype is the type other operations down the line will likely ingest from the previous
+calculation while the accumulation datatype describes the size of buffer where the results are initially
+stored. For NVidia's tensor cores for example many operations accumulate in FP32 but have an output datatype
+of FP16. The default implementation will follow this guideline closely and will by default have all 
+operations output FP16 and accumulate in FP32 only if TVM supports mixed datatypes for that particular
+operation.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
```
comaniac, line 66: This section should also discuss the implementation. Specifically, 1) the interface of annotating an op with color, 2) the coloring algorithm in the pass, 3) some corner cases (i.e., ops) that need more care.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+
+Many of the ideas are taken from Tensorflow's [automatic mixed precision training framework](https://on-demand.gputechconf.com/gtcdc/2019/pdf/dc91247-automatic-mixed-precision-in-tensorflow.pdf)
+and the initial "Green", "Gray", and "Red" lists are based [similarly](github.com/tensorflow/tensorflow/blob/v2.5.0/tensorflow/core/grappler/optimizers/auto_mixed_precision_lists.h). 
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+
+We still need to make sure that the current design and knobs exposed provide extensibility to every hardware platform out there.
```
comaniac, line 120: This seems not the answer to this question. IIUC, the initial implementation of the pass has been merged, so we should mention that here with the PR link.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+
+Probably a lot of edge cases of operations within TVM.
```
comaniac, line 124: This is too vague. It's better to have a quantitative metric toward to a stable release. For example, you can set up a benchmark with a set of models, and the goal is to make all of them work well with AMP (in terms of the performance and accuracy) on both CPU and GPU.

In addition, it would be better to also investigate how AutoScheduler works with AMP models. Since tuning is an important feature in TVM, the impact of AMP would be moderated if a tuned FP32 model can still run faster than an AMP model.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
+- Feature Name: Automatic Mixed Precision Pass
+- Start Date: 2021-06-08 
+- RFC PR: TODO
```
comaniac, line 3: Update the RFC PR.

rfcs/0001-AMP_pass.md
```diff
@@ -0,0 +1,137 @@
...
+- GitHub Issue: TODO
```
comaniac, line 4: As this RFC is guaranteed to be merged and the feature must be landed, It should be fine to open a tracking issue now and update the link here.

comaniac (changes requested):
The concept and algorithm look good to me, but it would be better to provide more implementation/design details.

rfcs/0001-AMP_pass.md
comaniac: I found that the issue has been opened, so please just update the link here.

AndrewZhaoLuo:
Thanks for driving this review @comaniac. I'll get to this later in the week.

rfcs/0001-AMP_pass.md
u99127: The M6G for instance also has FP16 instructions from the AArch64 ISA including Advanced SIMD fixed length 128 bit vectors- perhaps also investigate the suitability of the representation to support more than 1 backend with this ? I also expect that with the SVE (Scalable Vector extensions) you'd see FP16 instructions there as well but that's a different kettle of fish.

I also expect that in the uTVM case it would be interesting in a later date for the MVE instruction set with FP16 being present there as well. I'd like to look at whether there is an multiply and accumulate instruction there that matches with an FP32 accumulator.  Off hand I'm not sure about the answer to that question.

Just something to consider.

My 2 cents 

Ramana

comaniac: Thanks for the suggestion. I agree that it would be interesting to see how AMP works on m6g instances. Maybe we can document it in future possibilities section so that everyone interested in this feature can potentially take over after the AMP pass is stable on x86 platforms.

AndrewZhaoLuo:
Going to get to this tomorrow 😬. Promise 🤞

comaniac:
btw, according to #17, please update the RFC number on the file name to align with this PR number.

comaniac:
Took a quick pass to the updated RFC. I think it's almost ready to merge as long as the last 3 comments are resolved.

rfcs/0001-AMP_pass.md
AndrewZhaoLuo: Done. Please let me know if this is sufficient. Don't have the best background on some of this stuff.

rfcs/0001-AMP_pass.md
AndrewZhaoLuo: I am not familiar with this and cannot find a good article, do you have a link?

rfcs/0001-AMP_pass.md
AndrewZhaoLuo: Hey all, in the case where does not have an instruction for accumulating fp16 results in fp32, the user has the interface to turn off accumulating in fp32. 

Right now the default is designed with NVidia GPU in mind, but example pass settings can be easily made.

As for concerns about the representation being suitable for your device, do you have any specific concerns? As long as the codegen path to the device is good and clear I do not think the representation will be a problem.

AndrewZhaoLuo: @jwfromm I believe we also had good results running mixed precision on M6g (graviton) instances?

AndrewZhaoLuo: @comaniac x86 is not really the target for this pass as it has poor hardware FP16 support. I'll add a list of supported targets we will aim for.

AndrewZhaoLuo:
PTAL @comaniac

rfcs/0001-AMP_pass.md
comaniac: Yeah that would be sufficient for now. We will also need the BF16 support as well.

rfcs/0001-AMP_pass.md
comaniac: It's by tracing the PyTorch code so I don't have an article talking about this. The basic idea is since PyTorch uses interpreter to execute the model, it dynamically casts a FP32 tensor to FP16 when the op it's going to execute requires FP16 inputs. Accordingly, to avoid casting a tensor from FP32 to FP16 multiple times, it has a tensor cache to cache the casted tensors, so that they can be directly used in the future.

With the above content it should be clear that why TVM should adopt the XLA-like approach instead of PyTorch, because TVM is more compile-oriented and we are able to embrace more graph-level optimizations.

AndrewZhaoLuo: Done, added discussion on this topic

AndrewZhaoLuo:
@comaniac, I'll be talking about this at the TVM community meeting tomorrow so put off merging until after.

comaniac (approved):
LGTM. Will merge after the community meeting if there's no objection.

AndrewZhaoLuo:
If there is not other objections, this will be merged on monday.

comaniac:
Thanks @AndrewZhaoLuo

MeJerry215:
@AndrewZhaoLuo will it remove cast weight to float16 from graph? and make weight as float16 when build lib. 
in my opinion, it will reduce the bandwidth.

masahi:
@MeJerry215 Yes, casting of weight to fp16 is done at compile time by `FoldConstant` pass, so weights will be in fp16 at deploy time.

gayatripk1:
@AndrewZhaoLuo, When we enable mixed precision arithmetic, the output types of a model depends on the last operation in the relay IR, ie, if it convolution, it will be in float16 and for the case argmax it will be int32 and in the case of softmax it will be float32. So, is this expected? Or should the nodes that are marked as output nodes should be treated differently so that they return the expected specific out_type irrespective of the optimization(mixed precision pass) happening? Please let me know your thoughts on this.

PR #7, opened 2021-06-09
[UnitTests] Parametrized Unit Tests

Lunderberg:
Migrating earlier conversation from https://discuss.tvm.apache.org/t/rfc-parametrized-unit-tests/9946.  RFC should be up to date with all changes/conversations that occurred in that thread.

@areusch Let's continue any discussion here, to avoid having multiple places to check for the same RFC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Emacs temporary files
+*~
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-# Emacs temporary files
-*~
--- a/rfcs/0007-parametrized-unit-tests.md
+++ b/rfcs/0007-parametrized-unit-tests.md
@@ -522,17 +522,21 @@ Please also take into consideration that TVM intentionally diverges from other c
 
   If only serializable values are allowed to be cached, this may aid
   in debugging, since the values of all test parameters and cached
-  fixtures could be saved and reproduced.
-
-  Currently, nearly all cases (e.g. datasets, array sizes, targets)
-  are serializable.  The only non-serializable case after
-  brainstorming would be RPC server connections.  There is some
-  concern that caching RPC server connections could cause difficulties
-  in reproducing test failures.
+  fixtures could be saved and reproduced.  Currently, nearly all cases
+  (e.g. datasets, array sizes, targets) are serializable.  The only
+  non-serializable case after brainstorming would be RPC server
+  connections.  There is some concern that caching RPC server
+  connections could cause difficulties in reproducing test failures.
 
   Current proposed answer is to only cache serializable values, and
   that the discussion can be resumed when we have other possible use
   cases for caching non-serializable values.
+
+  For the time-being, both to prevent non-serializable values from
+  being cached, and to maintain separation between unit tests, all
+  cached values will be copied using
+  [`copy.deepcopy`](https://docs.python.org/3/library/copy.html#copy.deepcopy)
+  prior to returning the generated value.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/rfcs/0007-parametrized-unit-tests.md
+++ b/rfcs/0007-parametrized-unit-tests.md
@@ -1,6 +1,6 @@
 - Feature Name: Parametrized Unit Tests
 - Start Date: 2021-05-10(fill me in with today's date, YYYY-MM-DD)
-- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- RFC PR: [apache/tvm-rfcs#0007](https://github.com/apache/tvm-rfcs/pull/0007)
 - GitHub PR: [apache/tvm#8010](https://github.com/apache/tvm/issues/8010)
 
 # Summary
--- a/rfcs/0007-parametrized-unit-tests.md
+++ b/rfcs/0007-parametrized-unit-tests.md
@@ -14,18 +14,20 @@ parameters, or have setup that depends on input parameters.
 
 Some unit tests should be tested along a variety of parameters for
 better coverage.  For example, a unit test that does not depend on
-target-specific features should be tested on all targets that the test
+target-specific features could be tested on all targets that the test
 platform supports.  Alternatively, a unit test may need to pass
 different array sizes to a function, in order to exercise different
 code paths within that function.
 
 The simplest implementation would be to write a test function that
-loops over all parameters, throwing an exception if any parameter
-fails the test.  However, this does not give full information to a
-developer, as a failure from any parameter results in the entire test
-to be marked as failing.  A unit-test that fails for all targets
-requires different debugging than a unit-test that fails on a single
-specific target, and so this information should be exposed.
+internally loops over all parameters and throws an exception when the
+test fails.  However, this does not give full information to a
+developer, because `pytest` does not necessarily include the parameter
+in the test report.  Even when it does, the value will be printed in a
+different location depending on how the internal loop is written.  A
+unit-test that fails for all targets requires different debugging than
+a unit-test that fails on a single specific target, and so this
+information should be exposed.
 
 This RFC adds functionality for implementing parameterized unit tests,
 such that each set of parameters appears as a separate test result in
@@ -36,24 +38,27 @@ the final output.
 
 ## Parameters
 
-To make a new parameter for unit tests to use, define it with the
-`tvm.testing.parameter` function.  For example, the following will
-define a parameter named `array_size` that has three possible values.
-This can appear either at global scope inside a test module to be
-usable by all test functions in that module, or in a directory's
-`conftest.py` to be usable by all tests in that directory.
+Before you can use a parameter in a test case, you need to register it
+with `pytest`.  Do this with the `tvm.testing.parameter` function.
+For example, the following will define a parameter named `array_size`
+that has three possible values.  This can appear either at global
+scope inside a test module to be usable by all test functions in that
+module, or in a directory's `conftest.py` to be usable by all tests in
+that directory.
 
 ```python
 array_size = tvm.testing.parameter(8, 256, 1024)
 ```
 
 To use a parameter, define a test function that accepts the parameter
-as an input.  This test will be run once for each value of the
-parameter.  For example, the `test_function` below would be run three
-times, each time with a different value of `array_size` according to
-the earlier definition.  These would show up in the output report as
-`test_function[8]`, `test_function[256]`, and `test_function[1024]`,
-with the name of the parameter as part of the function.
+as an input, using the same argument name as was used above in the
+parameter registration.  This test will be run once for each value of
+the parameter.  For example, the `test_function` below would be run
+three times, each time with a different value of `array_size`
+according to the earlier definition.  These would show up in the
+output report as `test_function[8]`, `test_function[256]`, and
+`test_function[1024]`, with the name of the parameter as part of the
+function.
 
 ```python
 def test_function(array_size):
@@ -162,7 +167,7 @@ variable `TVM_TEST_DISABLE_CACHE` to a non-zero integer.  This can be
 useful to re-run tests that failed, to check whether the failure is
 due to modification/re-use of a cached value.
 
-A fixture can depend on parameters, or on other fixtures.  This is
+A fixture can also depend on parameters or on other fixtures.  This is
 defined by accepting additional parameters.  For example, consider the
 following test function.  In this example, the calculation of
 `correct_output` depends on the test data, and the `schedule` depends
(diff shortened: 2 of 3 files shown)


areusch:
@Lunderberg I think there were two open items from the discuss post:
1. shadowing local vars with module-level vars containing their fixtures
2. caching non-serializable data

on 1, it sounds like you had found a solution, so I think we're resolved there.

on 2, you'd suggested to table the discussion of how to cache non-serializable objects and just cache serializable ones. that sounds great to me.

so then, the real reason I was having this debate with you: i'd like to propose that in all cases (cache hot or cold), we always provide test functions with a brand-new deserialized copy of the data. i think that solves all of my concerns around tests poisoning downstream tests with cached data, and hopefully has not-too-significant performance impact. what are your thoughts?

Lunderberg:
@areusch

I think that makes sense for now, and is a good way to maintain separation between unit tests.  Running through `copy.deepcopy` would be the easiest way to do the serialize/deserialize steps on arbitrary data.  It throws a `TypeError` if a non-serializable value (e.g. `tvm.nd.array`) is passed, so that gives a good way to have the serializable-only restriction be checked by the CI.

I ran a few quick benchmarks for numpy arrays, since I think those would be the most frequently cached result.  `copy.deepcopy` is a little bit slower than calling dedicated copying functions (e.g. `my_numpy_array.copy()`), but wouldn't require any type-checking on our part.

image

areusch:
@Lunderberg I was thinking you could just require tests provide `bytes` to be cached.

Lunderberg:
How would we deserialize the objects in that case?  I thought that `__bytes__` was a one-way conversion from an object to a sequence of bytes, and didn't require objects to be deserializable.

Lunderberg:
@areusch I've added verbiage to the RFC to indicate use of `copy.deepcopy` for the results of cached fixtures, and have updated the [PR](https://github.com/apache/tvm/pull/8010) with the copying.  The error message if a non-serializable type is cached gives a link to this RFC, to avoid any confusion in the future.

.gitignore
```diff
@@ -0,0 +1,2 @@
+# Emacs temporary files
```
areusch, line 1: i think these should go in your global .gitignore not the project-specific one

Lunderberg: Fair enough, and I can remove it if that is preferable.  I try to avoid using the global `.gitignore` in most cases, so that `git add .` is safe to use in general, and not just on my environment.  The tvm repo already has this line in its `.gitignore`, so I figured it was good to add here.  (Though in retrospect it probably should have been a separate PR.)

Lunderberg:
Copying some notes from a 1:1, along with commentary at the bottom.

Caching options listed below, arranged from most flexibility at the top to least flexibility at the bottom.
1. All caching allowed.  A fixture tagged with `cache_return_value=True` can return any python object.  All tests that use the fixture receive the same python object as input.
    - Pros
        - Most potential to speed up running of CI.
        - Most potential to speed up by cross-function/file use of shared setup (e.g. np.random.uniform)
        - Works for any type.
    - Cons
        - Most potential to have interacting unit tests.
        - Could make somebody think that interactions are allowed between unit tests.
2. Cache only copy-able types.  A fixture tagged with `cache_return_value=True` has cached return values, but each parametrized unit test receives a copy made by `copy.deepcopy` prior to returning the object.  Any types that raise an error during copying result in a failed setup for the test.
    - Pros
        - Very simple to add caching to existing functions.
    - Cons
        - Unintentionally copy global state that is pointed to by local state.  Doesn't impact correctness, but would increase runtime.
        - Less flexibility, an object that cannot be deepcopy-ed results in a type error.
3. Cache only copy-able types, but with a custom implementation of `copy.deepcopy`.  Would follow the same rules if a `__deepcopy__` or pair of `__getstate__`/ `__setstate__` exists, but would otherwise require the type to be in a list of allowed types.  List of types
    - Pros
        - Opt-in both by fixture and by data-type.
    - Cons
        - Re-implements a common utility.
4. Cache using explicitly specified serialize/deserialize functions.  A fixture tagged with `cache_return_value=True` must return three values.  The first is the fixture value itself.  The second/third are serialize/deserialize functions for that value.
    - Pros
    - Cons
5. Only caching of `bytes` type is allowed.  A fixture tagged with `cache_return_value=True` must return a `bytes` object.  This object is deserialized at the start of the test function into the desired type.
    - Pros
    - Cons
6. No caching.  All fixtures are recomputed for each parametrized value.
    - Pros
        - All unit tests are independent by design.  Zero interaction possible.
    - Cons
        - Most potential impact to CI runtime.
        
Overall, @areusch 's concern is avoiding hidden failure modes, especially where there could be tests that only appear to pass due to cross-talk.  As such, he is more comfortable with more explicit usage. and would be comfortable with any of implementations 3-5.  @Lunderberg 's main concern is usability and ease of enabling, such that a slow test can have caching enabled with minimal effort where it is needed, and would be comfortable with any of implementations 2-3.  One key point that we came up with is that we expect fixtures to be uncached by default (`@tvm.testing.fixture`), and to have caching enabled on a per-fixture basis (`@tvm.testing.fixture(cache_return_value=True)` where there are performance benefits to be gained.

Since #3 is the overlap we both like between having opt-in caching, both on a per-return-value and per-fixture basis, I've put together [a test implementation](https://gist.github.com/Lunderberg/039f2f1d14c9ad7de88e1d2224a430cc).  It calls `copy.deepcopy`, but passes in a custom `memo` argument that performs type-checking of the copied type.  Types can be explicitly listed as being safe to copy, but the use of the default `object.__reduce__` is disabled.  This implementation minimizes the amount of re-implementing that would need to be done, but does rely on `ctypes` to identify the object being checked.

@areusch Can you check this over to make sure I've given an accurate description of your comments?

areusch:
@Lunderberg i think that sounds right. i agree #3 is the overlap between our positions. i'd like for others to comment whether it may be too complex to understand in debugging; otherwise i think we can proceed here. to be explicit: we intend to copy only whitelisted types, and not allow `copy.deepcopy` to work on arbitrary objects.

Lunderberg:
@areusch Any status changes on this?

.gitignore
areusch: okay, please remove this from the change then. we can debate it on another RFC

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+
+This RFC documents how to implement unit tests that depend on input
+parameters, or have setup that depends on input parameters.
+
+# Motivation
+[motivation]: #motivation
+
+Some unit tests should be tested along a variety of parameters for
+better coverage.  For example, a unit test that does not depend on
+target-specific features should be tested on all targets that the test
```
areusch, line 17: nit: "could be tested"

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+platform supports.  Alternatively, a unit test may need to pass
+different array sizes to a function, in order to exercise different
+code paths within that function.
+
+The simplest implementation would be to write a test function that
+loops over all parameters, throwing an exception if any parameter
+fails the test.  However, this does not give full information to a
+developer, as a failure from any parameter results in the entire test
+to be marked as failing.  A unit-test that fails for all targets
```
areusch, line 26:
```suggestion
The simplest implementation would be to write a test function that
internally loops over all parameters and throws an exception when the 
test fails.  However, this does not give full information to a
developer because `pytest` does not necessarily include the parameter
value in the test report (even when it does, the value is in a different place
depending on the way the internal loop is written). A unit-test that fails for all targets
```

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+such that each set of parameters appears as a separate test result in
+the final output.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Parameters
+
+To make a new parameter for unit tests to use, define it with the
+`tvm.testing.parameter` function.  For example, the following will
```
areusch, line 40:
```suggestion
Before you can use a parameter in a test case, you need to register it with `pytest`. 
Do this using the `tvm.testing.parameter` function.  For example, the following will
```

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+This can appear either at global scope inside a test module to be
+usable by all test functions in that module, or in a directory's
+`conftest.py` to be usable by all tests in that directory.
+
+```python
+array_size = tvm.testing.parameter(8, 256, 1024)
+```
+
+To use a parameter, define a test function that accepts the parameter
+as an input.  This test will be run once for each value of the
```
areusch, line 51:
```suggestion
as an input (use the same name for the argument as you used above in the 
parameter registration). This test will be run once for each value of the
```

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+def test_function(dataset):
+    assert(True) # Test succeeds
+```
+
+The caching can be disabled entirely by setting the environment
+variable `TVM_TEST_DISABLE_CACHE` to a non-zero integer.  This can be
+useful to re-run tests that failed, to check whether the failure is
+due to modification/re-use of a cached value.
+
+A fixture can depend on parameters, or on other fixtures.  This is
```
areusch, line 165:
```suggestion
A fixture can also depend on parameters or other fixtures.  This is
```

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+    return setup_schedule(array_size)
+    
+def test_function_new(dataset, correct_output, schedule):
+    output = generate_output(dataset, schedule)
+    tvm.testing.assert_allclose(output, correct_output)
+```
+
+## Target/Device Parametrization
+
+The TVM test configuration contains definitions for `target` and
```
areusch, line 210:
```suggestion
The global TVM test configuration contains definitions for `target` and
```

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+```python
+# Before
+if __name__=='__main__':
+    test_function_1()
+    test_function_2()
+    ...
+    
+# After
+if __name__=='__main__':
+    sys.exit(pytest.main(sys.argv))
```
areusch, line 297: we should ideally split this out into a helper function and eventually add a pylint plugin to enforce, but good to document standard right now

rfcs/0007-parametrized-unit-tests.md
```diff
@@ -0,0 +1,568 @@
...
+    - Extending the parameters that apply across all tests in a
+      file/directory requires updating several locations.
+    - Similar parameters (e.g. 1000 vs 1024 for an array length) would
+      be defined at separate locations, and would then require
+      separate fixture setup.
+
+# Prior art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and the bad, in relation to this proposal.
```
areusch, line 501: can you remove the boilerplate and just add a reference to i guess `pytest.mark.parametrize` and `functools.lru_cache` or whatever that is now?

areusch (changes requested):
hi @Lunderberg sorry for the long delay here. I reread this and just have a few minor clarifications I'd like you to make on the wording. Other than that, looks great! Feel free to ping aggressively when these are addressed and we can merge right away.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Good point, and changed.  The "should" is a stronger statement, and was what I initially intended, but I'd changed my mind given the current impact to CI runtime.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated, mostly following your modified wording.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated as suggested.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated, but with a comma instead of a parenthetical.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated as suggested.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated as suggested.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Agreed, and that would help with readability.  I've added it to my todo list, though it's low on the priority.

rfcs/0007-parametrized-unit-tests.md
Lunderberg: Updated with a reference to `pytest.mark.parametrize` and `pytest.fixture`'s caching using the `scope` parameter.

areusch (approved):
thanks @Lunderberg !

PR #8, opened 2021-06-30
Add Project API RFC

areusch:
This PR adds an RFC for the Project API ([embryonic discussion](https://discuss.tvm.apache.org/t/tvm-simplifying-the-compiler-interface/8703) - [m2 roadmap item #6](https://discuss.tvm.apache.org/t/tvm-microtvm-m2-roadmap/8821)`). Project API is a plugin-style infrastructure that allows TVM to integrate with a variety of platform-specific build systems. It is particularly useful to allow TVM to build and time operator implementations for non-traditional build platforms (e.g. embedded firmware) under the microTVM effort.

Draft PR: https://github.com/apache/tvm/pull/8380

@mehrdadh @guberti @tqchen @jroesch @tkonolige @csullivan @leandron @u99127 @mousius @giuseros @gromero @stoa @hogepodge

### DIFF
--- a/rfcs/0008-microtvm-project-api.md
+++ b/rfcs/0008-microtvm-project-api.md
@@ -490,8 +490,8 @@ ASCII alphabet is chosen to represent binary data, and the binary data is transl
 alphabet. Since there aren't enough ASCII characters to represent all 2^8 == 256 binary values in
 one byte, a smaller alphabet is chosen, typically with 64 or 85 characters. This is referred to as
 `base64` or `base85`, and the binary data is encoded by modular arithmetic into the smaller
-alphabet. Python provides standard support for these via the `base64` module, so the most compact
-encoding (`base85`) was chosen from those standards to encode binary data in the Project API.
+alphabet. Python provides standard support for these via the `base64` module, so the more widely-
+used encoding (`base64`) was chosen from those standards to encode binary data in the Project API.
 
 ### Timeouts
 
--- a/rfcs/0008-microtvm-project-api.md
+++ b/rfcs/0008-microtvm-project-api.md
@@ -1,0 +1,518 @@
+- Feature Name: microtvm_project_api
+- Start Date: 2020-06-09
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC describes how TVM integrates with build systems for unconventional platforms, such as those
+for microcontrollers and for other bare-metal scenarios.
+
+# Motivation
+[motivation]: #motivation
+
+Though TVM's primary goal is generating code to implement models from a high-level description,
+there are several reasons why a user might want to interact with a platform's build system through
+TVM:
+
+1. To perform autotuning. TVM's internal operator implementations are merely templates and rely on
+   an automatic search process to arrive at a fast configuration for the template on a given
+   platform. This search process requires that TVM iteratively build and time code on the platform.
+2. To perform remote model execution. A user may wish to try several different models or schedules
+   on a different platform without rewriting the platform-specific code. Users can do this by
+   building generated model code against a generic implementation of the microTVM RPC server.
+3. To debug model execution remotely. Some aspects of model execution are easy to debug using a
+   platform-specific debugger; however, some things, such as analyzing intermediate tensor values,
+   are more easily accomplished with TVM-specific tooling. By leveraging the generic microTVM
+   RPC server used in (2), TVM can provide such tooling in a platform-agnostic way.
+
+TVM currently supports these use cases through a set of interfaces:
+1. `tvm.micro.Compiler`: used to produce binary and library artifacts.
+2. `tvm.micro.Flasher`: used to program attached hardware
+3. `tvm.micro.Transport`: used to communicate with on-device microTVM RPC server
+
+Thus far, implementations of these interfaces have been made for Zephyr, mBED OS, and for
+simulated hardware using a POSIX subprocess. The latter two interfaces have proven to be a
+relatively good fit; however, `tvm.micro.Compiler` is difficult to implement because it attempts
+to replicate a platform's build system in TVM. TVM does not want to incorporate platform-specific
+build logic into its codebase.
+
+This proposal unifies these three interfaces to form a "Project-level" interface, recognizing that
+it's typical to interact with unconventional platforms and their build systems at this level. It
+simplifies the `Compiler` interaction into a project-level Build, and adds an explicit
+`generate_project` method to the interface. These changes remove the need to build components
+and drive the link process from TVM.
+
+As a goal, this proposal aims to allow for the same use cases as are currently supported with these
+improvements:
+
+1. Integrating more naturally with build systems typical of embedded platforms.
+2. Allowing TVM to automatically generate projects  platforms to define automated scripts to build projects
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+TVM can interact with platform SDKs via its **Project API**. Such SDKs are common when working with
+non-traditional OS platforms, such as those commonly used in embedded systems (e.g. Arduino, Zephyr,
+iOS). Given a platform-specific implementation of this Project API, TVM can:
+1. Generate projects that integrate implemented TVM models with generic platform runtime componeents
+2. Build those generated projects
+3. Program attached hardware
+4. Drive remote model execution via the TVM RPC Server interface.
+
+This last capability means that TVM can drive autotuning, remotely perform model inference, and debug
+models on non-traditional OS such as Arduino, Zephyr, and mobile platforms such as iOS and Android.
+
+To provide support for a platform, a **template project** is first defined. Template projects are
+expected to exist entirely inside a directory and are identified to TVM by the path to the directory
+locally. Template projects may live either inside the TVM repository (when they can be included in the
+TVM CI) or in other version control repositories. The template project contains at minium an
+implementation of the **Project API** inside an executable program known as the
+**TVM Project API Server**.
+
+To begin working with a particular platform's Project API implementation, the user supplies TVM with
+the path to the top-level directory. TVM launches an instance the Project API Server (found at a
+standard location in that directory). TVM communicates with the Project API Server using JSON-RPC
+over standard OS pipes.
+
+TVM supplies generated code to the Project API Server using [Model Library Format](0001-model-library-format.md).
+
+Below is a survey of example workflows used with the Project API Server:
+
+## Generating a project
+
+1. The user imports a model into TVM and builds it using `tvm.relay.build`.
+2. The user supplies TVM with the path to the template project and a path to a non-existent
+   directory where the generated project should live.
+3. TVM launches a Project API server in the template project.
+4. TVM verifies that the template project is indeed a template by invoking the Project API server
+   `server_info_query` method.
+5. TVM invokes the Project API server `generate_project` method to generate the new project.
+
+## Building and Flashing
+
+1. The user follows the steps under [Generating a project](#generating-a-project).
+2. TVM expects the Project API server to copy itself to the generated project. It launches a
+   Project API server in the generated project directory.
+3. TVM verifies that the generated project is not a template by invoking the Project API server
+   `server_info_query` method. This method also returns `options` that can be used to customize
+   the build.
+4. TVM invokes the Project API server `build` method to build the project.
+5. TVM invokes the Project API server `flash` method to program the attached device. The
+   `options` can be used to specify a device serial number.
+
+## Host-driven model inference
+
+1. The user follows the steps under [Generating a project](#generating-a-project).
+2. TVM invokes the Project API server `connect_transport` method to connect to the remote on-device
+   microTVM RPC server.
+3. The microTVM RPC server is attached to a traditional TVM RPC session on the host device.
+4. TVM drives inference on-device using the traditional TVM RPC methods. The Project API server
+   methods `read_transport` and `write_transport` are used to receive and send data.
+5. When the inference session is over, TVM invokes the Project API server method `close_transport`
+   to release any underlying I/O resources, and terminates the Project API server.
+
+## AutoTVM
+
+1. The user supplies a kernel to the AutoTVM tuner for search-based optimization.
+2. AutoTVM generates a set of task configurations, instantiates each task, and then invokes
+   `tvm.build` to produce a `Module` for each instantiated task.
+3. AutoTVM produces a Model Library Format archive from the `Module` for each instantiated task.
+4. AutoTVM passes the Model Library Format archive to the AutoTVM `runner`. Project API overrides the
+   traditional AutoTVM runner by providing a [`module_loader`](#module-loader). The microTVM
+   `module_loader` connects to a _supervisor_ TVM RPC server which carries out the microTVM project build
+   as part of the TVM RPC `session_constructor`. The following steps occur in the
+   `session_constructor`:
+
+    1. The Model Library Format tar is uploaded to the supervisor.
+    2. The user supplies a path, on the supervisor, to a template project.
+    3. The supervisor `session_constructor` performs the steps under
+       [Building and Flashing](#building-and-flashing).
+    4. The supervisor `session_constructor` invokes the Project API server `connect_transport` method
+       to connect to the remote device. The session constructor registers a traditional TVM RPC
+       session on the supervisor, and this session is also used by the AutoTVM runner due to the
+       `session_constructor` mechanism.
+
+5. The AutoTVM runner measures runtime as normal.
+6. The AutoTVM runner disconnects the session, closing the Project API server on the supervisor.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Project API implementation
+
+The Project API is a Remote Procedure Call (RPC)-type mechanism implemented using
+[JSON-RPC](https://www.jsonrpc.org/specification). The client and server are implemented in
+`python/tvm/micro/project_api`. Tests are implemented in
+`tests/python/unittest/test_micro_project_api.py`.
+
+## Project API interface
+
+The functions that need to be implemented as part of a Project API server are defined on the
+`ProjectAPIHandler` class in `python/tvm/micro/project_api/server.py`:
+
+```
+class ProjectAPIHandler(metaclass=abc.ABCMeta):
+
+    @abc.abstractmethod
+    def server_info_query(self) -> ServerInfo:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def generate_project(self, model_library_format_path : pathlib.Path, standalone_crt_dir : pathlib.Path, project_dir : pathlib.Path, options : dict):
+        """Generate a project from the given artifacts, copying ourselves to that project.
+
+        Parameters
+        ----------
+        model_library_format_path : pathlib.Path
+            Path to the Model Library Format tar archive.
+        standalone_crt_dir : pathlib.Path
+            Path to the root directory of the "standalone_crt" TVM build artifact. This contains the
+            TVM C runtime.
+        project_dir : pathlib.Path
+            Path to a nonexistent directory which should be created and filled with the generated
+            project.
+        options : dict
+            Dict mapping option name to ProjectOption.
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def build(self, options : dict):
+        """Build the project, enabling the flash() call to made.
+
+        Parameters
+        ----------
+        options : Dict[str, ProjectOption]
+            ProjectOption which may influence the build, keyed by option name.
+        """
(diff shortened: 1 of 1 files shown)


rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,518 @@
...
+TVM CI) or in other version control repositories. The template project contains at minium an
+implementation of the **Project API** inside an executable program known as the
+**TVM Project API Server**.
+
+To begin working with a particular platform's Project API implementation, the user supplies TVM with
+the path to the top-level directory. TVM launches an instance the Project API Server (found at a
+standard location in that directory). TVM communicates with the Project API Server using JSON-RPC
+over standard OS pipes.
+
+TVM supplies generated code to the Project API Server using [Model Library Format](0001-model-library-format.md).
```
mehrdadh, line 79: This link is broken. Is there an RFC for model library format on the way?

mehrdadh (changes requested):
@areusch thanks for the RFC! It's very informative.
Just added a minor comment.

rfcs/0008-microtvm-project-api.md
areusch: oh thanks for catching--it's actually just docs in https://github.com/apache/tvm/pull/8270 which should land soon

guberti (commented):
Really nice writeup! It might be good to explicitly state that a platform doesn't need to support all of `ProjectAPIHandler`'s methods - if I understand correctly, which workflows are supported can depend on what makes sense for each platform.

areusch:
@mdw-octoml would be great to get your feedback here as well

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+        try:
+            yield remote, remote.load_module(os.path.split(build_result.filename)[1])
+
+        finally:
+            # clean up remote files
+            remote.remove(build_result.filename)
+            remote.remove(os.path.splitext(build_result.filename)[0] + ".so")
+            remote.remove("")
+
+    return default_module_loader_mgr
```
tkonolige, line 357: Please update this definition to match apache/tvm#8363.

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
```
tkonolige, line 363: I think you can find some drawbacks.

Maybe:
- More code to maintain.
- Extra work for people using embedded OSes

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+ - Compact given the above constraint
+ - Easy to encode and decode in languages likely to be used for the API server
+
+Another common place this occurs is in transmitting binary data to and from websites, where an
+ASCII alphabet is chosen to represent binary data, and the binary data is translated into the
+alphabet. Since there aren't enough ASCII characters to represent all 2^8 == 256 binary values in
+one byte, a smaller alphabet is chosen, typically with 64 or 85 characters. This is referred to as
+`base64` or `base85`, and the binary data is encoded by modular arithmetic into the smaller
+alphabet. Python provides standard support for these via the `base64` module, so the most compact
+encoding (`base85`) was chosen from those standards to encode binary data in the Project API.
```
tkonolige, line 438: TVM uses base64 for encoding data in the relay text format. Maybe we should be consistent.

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+2. The `Compiler` interface is not the correct abstraction of a platform's build system.
+3. The interface is large and spread across multiple classes, making it difficult to assemble a
+   list of tasks needed to support new platforms.
+4. The implementation is overly complex, making it diffcult to actually support such platforms.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+1. Is anyone particularly opposed the RPC mechanism used here?
+2. Does this seem simple for downstream platforms to implement?
```
tkonolige, line 494: Do you have an example implementation?

areusch: a simple example is the subprocess-based microtvm test: https://github.com/areusch/incubator-tvm/tree/project-generator/src/runtime/crt/host

more complex is the zephyr implementation: https://github.com/areusch/incubator-tvm/tree/project-generator/apps/microtvm/zephyr/template_project

i'd recommend reading the first one if you're not familiar with zephyr; the second one is not a minimal example and includes code to communicate with qemu and physical boards

rfcs/0008-microtvm-project-api.md
areusch: that is also technically an easier modulo to compute, i think. would be open to this.

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+## Project API interface
+
+The functions that need to be implemented as part of a Project API server are defined on the
+`ProjectAPIHandler` class in `python/tvm/micro/project_api/server.py`:
+
+```
+class ProjectAPIHandler(metaclass=abc.ABCMeta):
+
+    @abc.abstractmethod
+    def server_info_query(self) -> ServerInfo:
```
areusch, line 159: todo: include tvm version here

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+   methods `read_transport` and `write_transport` are used to receive and send data.
+5. When the inference session is over, TVM invokes the Project API server method `close_transport`
+   to release any underlying I/O resources, and terminates the Project API server.
+
+## AutoTVM
+
+1. The user supplies a kernel to the AutoTVM tuner for search-based optimization.
+2. AutoTVM generates a set of task configurations, instantiates each task, and then invokes
+   `tvm.build` to produce a `Module` for each instantiated task.
+3. AutoTVM produces a Model Library Format archive from the `Module` for each instantiated task.
```
Mousius, line 121: Just to clarify, this is one combined model library format with all `Module`s combined so it only flashes once for many task permutations?

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+```
+
+## Project Options
+
+Each Project API server can return `project_options` as part of the `server_info_query` response.
+These can be specified by the user to allow them to give platform SDK-specific options to each API
+method.
+
+```
+ProjectOption = collections.namedtuple('ProjectOption', ('name', 'help'))
```
Mousius, line 289: Should we represent the server response in JSON rather than the client data types?

rfcs/0008-microtvm-project-api.md
Mousius: I don't think this extra work for those coming directly from an embedded background as they'll likely produce model files directly into their own embedded projects. For example, I wouldn't expect a current Zephyr user to use this flow over using `tvmc` and `west` directly due to their existing familiarity. 

Though I agree, with the variety of embedded OSes available, this will likely mean TVM as a project has to support a number of OSes - whether directly in checked in project generators or indirectly by features required to support different OSes.

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+        - Binary encoding
+        - Impossible to use today without compiling TVM
+        - Implementation is designed around TVM's remote inference use cases, and will
+          likely change as new demands arise there.
+
+TVM RPC was decided against given the requirement that TVM must be compiled. gRPC was considered,
+but ultimately rejected because JSON-RPC can be implemented in a single Python file without adding
+the complexities of an IDL compiler.
+
+## Transport functions
```
Mousius, line 410: I'd suggest this tightly coupling autotuning and prototyping to specific frameworks called directly from TVM is the main alternative to using the Project API, inclusive of RPC server and support frameworks?

Tightly coupling would be less code and potentially easier user journeys without the consideration of the additional server. Given we can mark dependencies to include as extras, are they a real concern for someone doing `pip install tlcpack` ?

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,518 @@
...
+   TVM tree (the main barrier is the CI and reviews).
+2. The `Compiler` interface is not the correct abstraction of a platform's build system.
+3. The interface is large and spread across multiple classes, making it difficult to assemble a
+   list of tasks needed to support new platforms.
+4. The implementation is overly complex, making it diffcult to actually support such platforms.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+1. Is anyone particularly opposed the RPC mechanism used here?
```
gromero, line 493: @areusch Hi Andrew. Sorry, I've posted a comment regarding it on the PR related to Project API code, so I'm posting it again here, since there is an item explicitly asking about feedback about it :)

So, I'd like to understand better the motivation for having the RPC mechanism since it seems to add an unnecessary complexity (to maintain, to debug, for instance). Would that be possible to have the exactly same interfaces but using merely a, let's say, Python module / class inclusion (like importing the `microtvm_api_server.py` at first from the template dir and then also later when the project dir is already created)?

I've just noticed what looks like a speed regression when talking to the serial port using the new Project API Transporter (I've shared the details on our #microtvm Discord channel). It looks like related to the limitations pointed out by you on `read` and `write` calls, regarding the timeouts. So I'm wondering if these limitations would simply go away if the RPC mechanism is removed, for instance. If so, that would weight in against the use of a RPC mechanism as-is. I'm guessing such a regression is  not observable in `test_zephyr.py` because the models there are too simple.

I also think it's a bit convoluted the way that the client-server works now, like a client creating local pipes, executing the server and passing them to it so the client and server can talk like as a parent/child mechanism, afaics. But I also wonder if you have  a more ambitious plan in the long term, like really decoupling the client/server (client runs in one host, and the server in yet another host, connected via UDP or TCP). In that case I see the value of implementing the RPC mechanism now.

gromero (commented):
I took a close look at the RFC and the Project API draft implementation.

Currently I can't comment much about the AutoTVM part.

On `generate_project` method, and considering TVMC, my only comment is that it should allow a project creation based also on  MLF .tar, instead of only a "live" executor. As we've discussed I don't think it's necessary to have it now and I'll send a patch for it mostly in the lines as we've discussed it, since TVMC will need it.

On `build` method I just think it should be a way to force a rebuild in the same project dir (i.e. even if `build` dir exists)

On `flash` method alright, no comments.

On `transport`, please consider my comments inline, specially the bit a about the speed regression.

Also, if possible, please consider the comments I've posted to the draft code also (https://github.com/apache/tvm/pull/8380)

Otherwise, looks great. Thanks for refactoring and improving it!

Finally, I confirm that the fixes for MAX_ defines are ok (no build error) and the per board configs are kicking in correctly.

Cheers,
Gustavo

rfcs/0008-microtvm-project-api.md
areusch: The main motivation for the RPC mechanism is to allow project generators to use python dependencies which may not be present in TVM's dependency set. For example, it may be necessary to include `pyserial` to facilitate communication with µTVM RPC server to support autotuning, but this dependency has no business being in TVM's official set of dependencies (we can't set a precedent that every single micro platform's dependencies become TVM's dependencies).

so the RPC server allows the microtvm_api_server.py to live in a separate process, and that breaks the dependency chain. 

> I've just noticed what looks like a speed regression when talking to the serial port using the new Project API Transporter (I've shared the details on our #microtvm Discord channel)

This was a great catch with the PoC; it's resolved now. the main reason was that I changed the semantics of `read` and `write` but forgot to update the TVM-side code in `src/runtime/micro/micro_session.cc`.

> I also think it's a bit convoluted the way that the client-server works now, like a client creating local pipes, executing the server and passing them to it so the client and server can talk like as a parent/child mechanism, afaics

your understanding is correct...the main issue with making this a network-aware RPC is that there are local paths included in the RPC interface now. I don't really intend to make this work over the network, but having a standard way of communicating rather than an ad-hoc one makes more sense to me. then there is a well-defined spec and it's clearer how bugs in the code should be handled.

rfcs/0008-microtvm-project-api.md
areusch: done

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,509 @@
...
+standard location in that directory). TVM communicates with the Project API Server using JSON-RPC
+over standard OS pipes.
+
+TVM supplies generated code to the Project API Server using [Model Library Format](0001-model-library-format.md).
+
+Below is a survey of example workflows used with the Project API Server:
+
+## Generating a project
+
+1. The user imports a model into TVM and builds it using `tvm.relay.build`.
```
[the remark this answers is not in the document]
areusch: added suggested command-lines

rfcs/0008-microtvm-project-api.md
areusch: correct, though the number of tasks here is not specified. that still needs to be added to AutoTVM past the initial work done in https://github.com/apache/tvm/pull/7545

rfcs/0008-microtvm-project-api.md
areusch: done

rfcs/0008-microtvm-project-api.md
areusch: the concern i have is that we don't want our TVM dependencies held back by some platform's need to depend on pyyaml. if e.g. TensorFlow introduces an `install_requires` of `pyyaml > 3.0`, but some mciro platform refuses to update past pyyaml `2.x`, the platform is going to lose the battle and become unsupported. going to the API server avoids this battle and also allows for development of API servers outside the TVM tree.

areusch:
Thanks @Mousius @gromero for your comments. updated the RFC to reflect accurately on PoC now that it seems to be passing regression. Please take another look so we can merge this and proceed forward with Project API.

@gromero some follow-ups on your comments:
> On generate_project method, and considering TVMC, my only comment is that it should allow a project creation based also on MLF .tar, instead of only a "live" executor.

I agree; let's merge additional logic in `project.py` to do this as follow-on to this impl.

> On build method I just think it should be a way to force a rebuild in the same project dir (i.e. even if build dir exists)

I believe that is the functionality now. Did you see something different?

> On transport, please consider my comments inline, specially the bit a about the speed regression.

Replied to your comments; I believe the speed issue is not a problem now.

> Also, if possible, please consider the comments I've posted to the draft code also (apache/tvm#8380)

Will take a look at these now that PoC is passing

> Finally, I confirm that the fixes for MAX_ defines are ok (no build error) and the per board configs are kicking in correctly.

I reworked the way these are done since posting the PoC--now `prj.conf` is generated from `microtvm_api_server.py` so we can have a single combined `template_project` and the `prj.conf` can be situation-specific. Let me know what you think.

rfcs/0008-microtvm-project-api.md
gromero: > The main motivation for the RPC mechanism is to allow project generators to use python dependencies which may not be present in TVM's dependency set

gromero: >This was a great catch with the PoC; it's resolved now. the main reason was that I changed the semantics of read and write but forgot to update the TVM-side code in src/runtime/micro/micro_session.cc.

Right, that issue is by now resolved. Thanks for fixing it :)

gromero: > your understanding is correct...the main issue with making this a network-aware RPC is that there are local paths included in the RPC interface now. 

right, I was scratching my head about it : )

>  I don't really intend to make this work over the network, but having a standard way of communicating rather than an ad-hoc one makes more sense to me. then there is a well-defined spec and it's clearer how bugs in the code should be handled.

Got it. Yeah, plus the current main reason to use an RPC interface here is well state. ok.

gromero:
>> On generate_project method, and considering TVMC, my only comment is that it should allow a project creation based also on MLF .tar, instead of only a "live" executor.
>I agree; let's merge additional logic in project.py to do this as follow-on to this impl.
Sure! I'll submit a follow up patch I've been using to test this patchset with TVMC. My comment was also more for the records for others following the discussion since we've already discussed the details about.

>> On build method I just think it should be a way to force a rebuild in the same project dir (i.e. even if build dir exists)
> I believe that is the functionality now. Did you see something different?
Yes, a functionality detail. Currently I'm handling that at TVMC side, like if the `build/` dir exists it means `tvmc micro build` was run before so it informs the user and asks to confirm a re-build using a `--force` flag. In that case, if `--force` is used TVMC will remove the `build/` and let Project API recreate it. But I was just wondering if you would like to change that behavior. I have no preference here. What do you think?

>> Also, if possible, please consider the comments I've posted to the draft code also (apache/tvm#8380)
>Will take a look at these now that PoC is passing
OK. I think you answered all my initial comments there. I'll just post a couple more related to that new round, i.e. related to last commit pushed: https://github.com/apache/tvm/commit/de75022daeb889682db80f9e90ab58b3eee898dd I have no more comments regarding the RFC itself - I'm happy with it, so I'll just post some comments about the code. I think it's pretty close to land :)

>>Finally, I confirm that the fixes for MAX_ defines are ok (no build error) and the per board configs are kicking in correctly.
> I reworked the way these are done since posting the PoC--now prj.conf is generated from microtvm_api_server.py so we can have a single combined template_project and the prj.conf can be situation-specific. Let me know what you think.

Yeah, I'm not much inclined to use that approach of using a Python code to generate the `prj.conf` because for adding new bits to the config fil will imply changing a Python code. That seems a bit strange to me. I thought of having a `prj.conf` per `project_type` but there is also other things to being handled like the stack configuration based on running the model on QEMU or a physical board, so I'm not sure, maybe we can go ahead with that approach and see it goes overtime? 

Regarding the `MAX_` defines, even with the new approach generating the `prj.conf` the `crt/crt_config.h` is used and so the `MAX_` defines are used anyways? Or I missed something?

gromero (approved):
LTGM.

areusch:
@Mousius please take another look when you can and explicitly approve!

rfcs/0008-microtvm-project-api.md
areusch: yeah I agree that end users may likely just consume MLF. I think that as a "getting started" flow, it may help to have automatic project generation. I expect this flow to be primarily beneficial in implementing autotuning or remote execution flows on various platforms.

I added some drawbacks

areusch:
@tkonolige @guberti @mehrdadh please take another look and explicitly approve if you're good w/ this

rfcs/0008-microtvm-project-api.md
Mousius: It'd be good to have that expressed explicitly as part of the RFC, attempting to continue with a subset of RTOSes or trying to balance different dependencies is a valid alternative to building out the project API but the reasons you've articulated give a concrete example to clear that up 😸

Mousius (changes requested):

rfcs/0008-microtvm-project-api.md
areusch: done!

rfcs/0008-microtvm-project-api.md
```diff
@@ -0,0 +1,543 @@
+- Feature Name: microtvm_project_api
+- Start Date: 2020-06-09
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
```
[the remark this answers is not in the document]
areusch: done

areusch:
@Mousius @tqchen please take a look

tqchen:
@Mousius @mehrdadh @tkonolige @guberti please take another look and https://tvm.apache.org/docs/contribute/code_review.html#approve-and-request-changes-explicitly

rfcs/0008-microtvm-project-api.md
tkonolige: Can you address this?

rfcs/0008-microtvm-project-api.md
tkonolige: Lets just do base64 to be consistent

tkonolige (changes requested):

rfcs/0008-microtvm-project-api.md
mehrdadh: sorry, but this needs to change again since the PR is merged :(

mehrdadh (approved):
Please change the URL since the other PR is merged now, otherwise LGTM.
Thanks!

guberti (approved):
This fits my Arduino use case perfectly. LGTM!

Mousius (approved):
Thanks for the updates :smile_cat:

rfcs/0008-microtvm-project-api.md
areusch: oops i see. done.

rfcs/0008-microtvm-project-api.md
areusch: okay fine :)

tkonolige (approved):
LGTM, thanks @areusch!

tqchen (approved):

PR #9, opened 2021-07-06
[RFC] TVM Unified Static Memory Planning

manupa-arm:
This commits adds the RFC (.md) for USMP.

pre-RFC on discuss : https://discuss.tvm.apache.org/t/rfc-unified-static-memory-planning/10099

### DIFF
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -357,20 +357,23 @@ Inputs :
     * Each tir.allocate in the IRModule annotated with candidate pools ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md))
 
 
-```
-struct PoolInfoNode : public Object {
-  String  pool_name;
-  Integer size_bytes;
-  Integer alignment;
-  Integer pool_offset;
-  Map<Target,String> target_access; // 'rw' or 'ro'
-}
-```
+        ```
+        struct PoolInfoNode : public Object {
+        String  pool_name;
+        Integer size_bytes;
+        Integer alignment;
+        Integer pool_offset;
+        Map<Target,String> target_access; // 'rw' or 'ro'
+        }
+        ```
+        The input IRModule is expected to have "candidate_memory_pools" annotation populated with a orderered list of PoolInfo objects. The ordering will indicate to the planner the order of preference each allocate will be pinned to each Pool. The core compiler will run a pass to assign each tir.allocate with candidate_memory_pools based on the target each PrimFunc is executed, prior to invoking the USMP.
 
 
-We could use "candidate_memory_pools" ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md)) to tag buffers with suggested priority order determined by the scheduler.
+The idea is USMP will try to pool them using the preferred "candidate_memory_pools" and fallback whenever the size is exceeding the user provided max size for each pool (if any). The fallback only happens if the tir.allocate is annotated with more than one candidate memory pool. Initially, it will take the ordering provided to the TVMC interface.
 
-The idea is USMP will try to pool them using the preferred "candidate_memory_pools" and fallback whenever the size is exceeding the user provided max size for each pool (if any). The fallback only happens if the user provide more than one candidate memory pool. If the fallback is not desired by the user, the user need not to provide multiple candidate_memory_pools with size constraints or the scheduling. If the fallback is not desired by the scheduler, the scheduling passes could remove the memory pools from the candidate_memory_pools.
+If the fallback is not desired, the user need not to provide multiple candidate_memory_pools with size constraints to TVMC interface.
+
+If the fallback is not desired by the scheduler, the scheduling passes could remove the memory pools from the candidate_memory_pools.
 
 Outputs : 
 * AoT TIR PrimFunc accepting pool buffers from the user.
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -10,6 +10,7 @@ Currently, given a ML model primarily TVM will generate two main artifacts :
 * A1 : Description of the sequential execution of operators :
   1. If the "executor" is "graph", this would be a JSON
   2. if the "executor" is "aot", this would be a main function describing call graph of operators
+  3. if the "executor" is "vm", this would be a series of VM bytecode instructions
 * A2 : library of operators (in the form of runtime.Module)
 
 A1 is generally created out of lowering the "main" relay function and A2 is created lowering fused relay primitive functions → TIR PrimFuncs → C or LLVM artifacts of the operator library.
@@ -361,10 +362,10 @@ Outputs :
 Special Parametric Inputs : 
 * function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
 
-The current proposal for the interface is as follows :
+The current proposal for the interface of the memory planning algorithm is as follows :
 ```
     struct BufferInfo {
-        Integer uid;
+        String name_hint; // this is the tir.buffer name
         Integer size_bytes;
         Integer alignment;
         Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
@@ -376,6 +377,7 @@ The current proposal for the interface is as follows :
 ```
 void (*foo)(Array<ByfferInfo> buffers, Map<String, Integer> pool_sizes)
 ```
+The memory planning algorithm is expected to populate the assigned pool_name and the offset. Additionally, the second argument provides size constraints for each pool (if any).
 ### Special Considerations :
 
 * tir.constants : TIR does not have the ability to represent constants – which is limiting and often leads to having side-channels to carry constants between TIR compiler passes including this one.
@@ -465,3 +467,7 @@ After Step 1 (introducing tir.constants to hold constant data) : the TIR code sh
 NOTE 1: All the above passes will have a mirror in the python.
 
 NOTE 2: to support tir.constants generally, we'll be enhancing the bound relay.constants to be lowered down to tir.constants to codegen. Those changes will appear through out the stack accordingly.
+
+# Drawbacks
+
+* The relay "main" function that describes the call order to operator PrimFuncs has to be described in TIR to be able to integrate the USMP into the respective executor codegen. However, we dont view this as a major problem as the relay "main" function could easily be lowered to TIR.
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -375,9 +375,9 @@ The current proposal for the interface of the memory planning algorithm is as fo
     }
 ```
 ```
-void (*foo)(Array<ByfferInfo> buffers, Map<String, Integer> pool_sizes)
+Array<BufferInfo> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
 ```
-The memory planning algorithm is expected to populate the assigned pool_name and the offset. Additionally, the second argument provides size constraints for each pool (if any).
+The memory planning algorithm is expected to populate the assigned pool_name with the offset and return the updated array of BufferInfo objects. Additionally, the second argument provides size constraints for each pool (if any).
 ### Special Considerations :
 
 * tir.constants : TIR does not have the ability to represent constants – which is limiting and often leads to having side-channels to carry constants between TIR compiler passes including this one.
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -7,10 +7,10 @@
 
 Currently, given a ML model primarily TVM will generate two main artifacts :
 
-* A1 : executor configuration : the description of the sequential execution of operators
-  1. If the "executor" is "graph", this would be a JSON
-  2. if the "executor" is "aot", this would be a main function describing call graph of operators
-  3. if the "executor" is "vm", this would be a series of VM bytecode instructions
+* A1 : A compilation artifact that describes a sequence of calls to the operators
+  1. If building for graph executor, this would be a JSON
+  2. If building for AoT executor, this would be a main function describing call graph of operators
+  3. If building for VM executor, this would be a series of VM bytecode instructions
 * A2 : library of operators (in the form of runtime.Module)
 * A3 : compiled parameters of the model
 
@@ -26,7 +26,7 @@ For A2, the operators are lowered to TIR PrimFuncs. There exist a pass called St
 
 # Motivation
 
-For embedded use-cases, its widely accepted that aggressive memory optimizations are vital. Intially we are looking at enable memory planning for embedded use-cases using the AoT executor.
+For embedded use-cases, its widely accepted that aggressive memory optimizations are vital. Intially we are looking at enabling memory planning for embedded use-cases using the AoT executor.
 
 Therefore, there exist two main shortcomings of the current approach :
 
@@ -56,7 +56,7 @@ Example TIR :
 
 The above TIR snippet shows that two intra operator buffers PaddedInput, DepthwiseConv2d are not visible for optimization by the Relay-level GraphPlanMemory approach.
 
-* Assumption of local optimization : performing sharing inside the operator first and sub-subsequently sharing that workspace with inter-operator tensors, would be sub-optimal.
+* Assumption of local optimization : performing sharing inside the operator first and subsequently sharing that workspace with inter-operator tensors, would be sub-optimal.
 
 Thus, for the embedded use-cases, we'd need a unified static memory planner that performs memory planning of all tensors holistically to achieve best memory utilization.
 
@@ -72,7 +72,8 @@ There are a variety of memory planning algorithms in discussion with different t
 
 G3. Multiple pool support (including constants)
 
-Ideally, the user would expect to provide these buffers in the granularity of the memories they'd want to pin them to. E.g., if there are two RW memories : DRAM and SRAM, the buffers need to be identified and pooled by the compiler. Similiarly, for constant data, we need to have a mechanism to allow user to pin them to appropriate memories and addresses in the IR would simply be offsets into the constant buffer(s) provided by the user
+Ideally, the user would expect to provide these buffers in the granularity of the memories they'd want to pin them to. E.g., if there are two RW memories : DRAM and SRAM, the buffers need to be identified and pooled by the compiler. Similiarly for constant data, we need to have a mechanism to allow users to pin them to appropriate memories and addresses. In the IR, they would simply be offsets into the constant buffer(s) provided by the user
+
 
 # Guide-level explanation
 
@@ -144,14 +145,16 @@ tvmc compile my_model.tflite --executor=aot --output-format=mlf --target=c
     tvmc compile my_model_1.tflite
     --executor=aot 
     --output-format=mlf
-    --target=accel,c  
-    --with-workspace-buffer= "name=sram;target=c,accel"
+    --target=accel,c
+    --usmp-workspace-pools=sram
+    --usmp-workspace-pool-sram= "target=c:rw,accel:rw"
 
     tvmc compile my_model_2.tflite 
     --executor=aot
     --output-format=mlf 
     --target=accel,c
-    --with-workspace-buffer= "name=sram;target=c,accel"
+    --usmp-workspace-pools=sram
+    --usmp-workspace-pool-sram= "target=c:rw,accel:rw"
 ```
 ### Codegen'd Artifacts
 ```
@@ -264,11 +267,13 @@ tvmc compile my_model.tflite --executor=aot --output-format=mlf --target=c
 ```
     tvmc compile my_model.tflite 
     --executor=aot 
-    --target=accel,c  
-    --with-workspace-buffer= "name=dtcm;target=c;size=1000" # Here the size is more of a hint/guide provided to USMP
-    --with-workspace-buffer= "name=sram;target=c,accel"
-    --with-parameter-buffer= "name=itcm;target=c;size=5000" # Here the size is more of a hint/guide provided to USMP
-    --with-parameter-buffer= "name=flash;target=c,accel"
+    --target=accel,c
+    --usmp-workspace-pools=dtcm,sram
+    --usmp-parameter-pools=itcm,flash
+    --usmp-workspace-pool-dtcm= "target=c;size=1000" # Here the size is more of a hint/guide provided to USMP
+    --usmp-workspace-pool-sram= "target=c,accel"
+    --usmp-parameter-pool-itcm= "target=c;size=5000" # Here the size is more of a hint/guide provided to USMP
+    --usmp-parameter-pool-flash= "target=c,accel"
 ```
 ### Codegen'd Artifacts
 ```
@@ -374,6 +379,13 @@ The idea is USMP will try to pool them using the preferred "candidate_memory_poo
 If the fallback is not desired, the user need not to provide multiple candidate_memory_pools with size constraints to TVMC interface.
 
 If the fallback is not desired by the scheduler, the scheduling passes could remove the memory pools from the candidate_memory_pools.
+
+* How is the fallback pool decided ?
+
+Currently, the pool ordering the user provides to TVMC interface will be used as a priority order for determining pools for tir.allocate/tir.allocate_const nodes (i.e. USMP will try to use highest priority pool for each allocate). Each pool specifies the information on whether it could be acessed by a given set of targets. Therefore, the ordered list of pools will be further filtered by for each allocate node that belongs to different targets.
+
+It is important to note that scheduling stage could remove pools from candidate pools for performance reasons.
+
 
 Outputs : 
 * AoT TIR PrimFunc accepting pool buffers from the user.
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -1,6 +1,6 @@
     Feature Name: Unified Static Memory Planner
     Start Date: 2021 June 1
-    RFC PR: #000y
+    RFC PR: #0009
     GitHub Issue: https://github.com/apache/tvm/issues/8404
 
 # Background
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -76,6 +76,8 @@ Ideally, the user would expect to provide these buffers in the granularity of th
 
 # Guide-level explanation
 
+NOTE : the embedded runtime interface used in the example are for demonstration purposes and the actual runtime API is defined and discussed [here.](https://discuss.tvm.apache.org/t/rfc-utvm-embedded-c-runtime-interface/9951)
+
 ## U1: Most simple use case
 
 ### TVMC
@@ -348,12 +350,27 @@ tvmc compile my_model.tflite --executor=aot --output-format=mlf --target=c
 
 This should be a IRModule (TIR) → IRModule (TIR) pass.
 
-Inputs : 
-* AoT TIR PrimFunc ( the control function describing the call graph to operators)
-* All Operator Functions
-* the maximum size for each pool We could use "pinned_memory" (see below) to tag buffers with suggested priority order determined by the scheduler.
+Inputs :
+* IRModule containing 
+    * AoT TIR PrimFunc (the control function describing the call graph to operators)
+    * All Operator Functions
+    * Each tir.allocate in the IRModule annotated with candidate pools ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md))
 
-The idea is USMP will try to pool them using the preferred "pinned_memory" and fallback whenever the size is exceeding the user provided max size for each pool (if any)
+
+```
+struct PoolInfoNode : public Object {
+  String  pool_name;
+  Integer size_bytes;
+  Integer alignment;
+  Integer pool_offset;
+  Map<Target,String> target_access; // 'rw' or 'ro'
+}
+```
+
+
+We could use "candidate_memory_pools" ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md)) to tag buffers with suggested priority order determined by the scheduler.
+
+The idea is USMP will try to pool them using the preferred "candidate_memory_pools" and fallback whenever the size is exceeding the user provided max size for each pool (if any). The fallback only happens if the user provide more than one candidate memory pool. If the fallback is not desired by the user, the user need not to provide multiple candidate_memory_pools with size constraints or the scheduling. If the fallback is not desired by the scheduler, the scheduling passes could remove the memory pools from the candidate_memory_pools.
 
 Outputs : 
 * AoT TIR PrimFunc accepting pool buffers from the user.
@@ -370,21 +387,27 @@ The current proposal for the interface of the memory planning algorithm is as fo
         Integer size_bytes;
         Integer alignment;
         Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
-        Array<Integer> pool_candidates;`
-        String pool_name;`
-        Integer pool_offset;`
+        Array<PoolInfo> pool_candidates;
     }
 ```
+
 ```
-Array<BufferInfo> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
+    struct PoolAllocation {
+        PoolInfo pool;
+        Integer offset;
+    }
 ```
-The memory planning algorithm is expected to populate pool_name and pool_offset and return the updated array of BufferInfo objects. Additionally, the second argument provides size constraints for each pool (if any).
+
+
+```
+Map<BufferInfo, PoolAllocation> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
+```
+The memory planning algorithm is expected to return a Map of BufferInfo to PoolAllocation with the planned offsets into respective pool.
 ### Special Considerations :
 
 * tir.constants : TIR does not have the ability to represent constants – which is limiting and often leads to having side-channels to carry constants between TIR compiler passes including this one.
-Therefore, in this work as a pre-requisite we should aim to fix this by supporting tir.constants (similiar to relay.constants).
-  * Why do we need constants expressed in TIR ?
-    * If not, it should be represented as inputs to TIR main function (logic : anything that is not expressible in TIR will become inputs). In which case, we would need to associate that Var with a special tag to indicate its constant and its metadata (e.g., desired pools, alignment requirements, etc.)
+Therefore, in this work as a pre-requisite we should aim to fix this by supporting tir.constants (similiar to relay.constants). Please refer to the [TIR non-scalar constants RFC](https://github.com/apache/tvm-rfcs/pull/22).
+
 * Currently "with" or "let" scopes are tree structured and carry transitive property. E.g, if tensor A is live with tensor B && tensor B is live with tensor C → tensor A is live with tensor C – which may not be true always.
 Thus current "let" or "with" scopes are unable to express liveness information. Therefore, we'd need a side-channel to express this information.
 
@@ -396,38 +419,42 @@ After Step 1 (introducing tir.constants to hold constant data) : the TIR code sh
 # This snippet shows the format of pre-USMP pseudo TIR code.
 
     def main(input1: ty.handle, output1: ty.handle):
-       my_model_fused_op1 = tir.allocate(..., pinned_memory=["dtcm", "sram"])
-       my_model_fused_op2 = tir.allocate(..., pinned_memory=["sram])
+       my_model_fused_op1 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"]
+       my_model_fused_op2 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"]
        tir.call("my_model_fused_op1", input1, my_model_fused_op1, fused_op1_weights, fused_op1_biases)
        tir.call( "my_model_fused_op2" , my_model_fused_op1, my_model_fused_op2, fused_op2_weights, fused_op2_biases)
 
     def my_model_fused_op1(input : ty.handle, output : ty.handle):
        tir.func_attr({"global_symbol":"my_model_fused_op1","tir.noalias": True})
-       intermediate_tensor_1 = tir.allocate(..., pinned_memory=["dtcm", "sram"]) # By  default they will have all possible memories
-       intermediate_tensor_2 = tir.allocate(..., pinned_memory=["dtcm", "sram"]) # unless scheduler removes them
-       weights = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
-       biases = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
+       intermediate_tensor_1 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"] 
+       intermediate_tensor_2 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"] 
+       weights = tir.allocate_const(...) # attrs.candidate_memory_pools = ["itcm", "flash"]
+       biases = tir.allocate_const(...) # attrs.candidate_memory_pools = ["itcm", "flash"]
        ...
        <compute>
        ...
 
     def my_model_fused_op2(input : ty.handle, output : ty.handle):
        tir.func_attr({"global_symbol":"my_model_fused_op2", "tir.noalias": True})
-       intermediate_tensor_1 = tir.allocate(..., pinned_memory=[1, 2])
-       intermediate_tensor_2 = tir.allocate(..., pinned_memory=[1, 2])
-       weights = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
-       biases = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
+       intermediate_tensor_1 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"]
+       intermediate_tensor_2 = tir.allocate(...) # attrs.candidate_memory_pools = ["dtcm", "sram"]
+       weights = tir.allocate_const(...) # attrs.candidate_memory_pools = ["itcm", "flash"]
+       biases = tir.allocate_const(...) # attrs.candidate_memory_pools = ["itcm", "flash"]
        ...
        <compute>
        ...
 ```
-##### Step 2 : Run an analysis pass to populate a Map<tir::Var, BufferInfo> that contains buffer information as defined above (See the struct BufferInfo).
+##### Step 2 : Run an analysis pass to populate a Map<BufferInfo, tir.StmtNode> that contains buffer information as defined above (See the struct BufferInfo).
 
-##### Step 3 : Use the updated Map<tir::Var, BufferInfo> to generate Array<BufferInfo>, Map<String, Integer> pool_sizes
+Note : here tir.StmtNode is treated as a Union[tir.AllocateNode, tir.AllocateConstNode]
 
-##### Step 4 : Call the provided/default algorithm (void (*foo)(Array<ByfferInfo> buffers, Map<String, Integer> pool_sizes) to populate pool_id and pool_offset.
+This actual pass would traverse full TIR program and construct BufferInfo objects that captures liveness conflicts between allocates that are live together.
 
-##### Step 5 : Use the updated Map<tir::Var, BufferInfo> (with pool_id and pool_offset) mutate the IR that would result as following :
+##### Step 3 : Use the updated Map<BufferInfo, tir.StmtNode> to generate Array\<BufferInfo>
+
+##### Step 4 : Call the provided/default algorithm (void (*foo)(Array<BufferInfo> buffers) to generate Map<BufferInfo, PoolAllocation>
+
+##### Step 5 : Use the updated Map<BufferInfo, PoolAllocation> and Map<BufferInfo, tir.StmtNode> to mutate the IR that would result as following :
 ```
 # This snippet shows the format of post-USMP pseudo TIR code.
 
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -7,11 +7,12 @@
 
 Currently, given a ML model primarily TVM will generate two main artifacts :
 
-* A1 : Description of the sequential execution of operators :
+* A1 : executor configuration : the description of the sequential execution of operators
   1. If the "executor" is "graph", this would be a JSON
   2. if the "executor" is "aot", this would be a main function describing call graph of operators
   3. if the "executor" is "vm", this would be a series of VM bytecode instructions
 * A2 : library of operators (in the form of runtime.Module)
+* A3 : compiled parameters of the model
 
 A1 is generally created out of lowering the "main" relay function and A2 is created lowering fused relay primitive functions → TIR PrimFuncs → C or LLVM artifacts of the operator library.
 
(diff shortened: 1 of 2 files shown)


manupa-arm:
cc : @areusch @mbaret @tqchen
(I cant seem to tag the original people in the RFC --  working on it)

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+    RFC PR: #0009
+    GitHub Issue: https://github.com/apache/tvm/issues/8404
+
+# Background
+
+Currently, given a ML model primarily TVM will generate two main artifacts :
+
+* A1 : Description of the sequential execution of operators :
+  1. If the "executor" is "graph", this would be a JSON
+  2. if the "executor" is "aot", this would be a main function describing call graph of operators
```
tkonolige, line 12:
```suggestion
  2. if the "executor" is "aot", this would be a main function describing call graph of operators
  3. if the "executor" is "vm", this would be a series of VM bytecode instructions
```

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+* A2 : library of operators (in the form of runtime.Module)
+
+A1 is generally created out of lowering the "main" relay function and A2 is created lowering fused relay primitive functions → TIR PrimFuncs → C or LLVM artifacts of the operator library.
```
tkonolige, line 15: Is A1 created from lowering the main function or from lowering the entire relay program?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+
+Outputs : 
+* AoT TIR PrimFunc accepting pool buffers from the user.
+* All Operator functions accepting pool buffers.
+  * Each operator function should address using the correct offset in the correct pool buffer
+
+Special Parametric Inputs : 
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface is as follows :
```
tkonolige, line 364: What is this the interface for? The planning algorithm?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
+        String pool_name;`
+        Integer pool_offset;`
+    }
+```
+```
+void (*foo)(Array<ByfferInfo> buffers, Map<String, Integer> pool_sizes)
```
tkonolige, line 377: Why doesn't this return anything? What is the point of the integer in the second argument?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface is as follows :
+```
+    struct BufferInfo {
+        Integer uid;
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
```
tkonolige, line 371: I assume each integer in this field refers to a specific pool? Why do we also have a pool_name field for each buffer?

tkonolige (changes requested):
I'm in favor of doing memory planning, so thank you for this PR. I've got a couple questions:
1. Having memory planning would be useful across all executors (graph, vm, aot). What is preventing us from doing USMP for all the executors? It seems like the main problem with non-aot executors is that we do not have a full view of all the buffers.
2. Can you include a drawbacks section?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,467 @@
...
+    --executor=aot 
+    --output-format=mlf
+    --target=accel,c  
+    --with-workspace-buffer= "name=sram;target=c,accel"
+
+    tvmc compile my_model_2.tflite 
+    --executor=aot
+    --output-format=mlf 
+    --target=accel,c
+    --with-workspace-buffer= "name=sram;target=c,accel"
```
tkonolige, line 150: Are buffers part of the target specification with this?

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: The description of the calls to the operator PrimFuncs could be defined as lowering the "main" relay function (that has calls to the primitive relay functions -- post fuse-ops). I'd view that differently to lowering of the entire relay program that uses the compile engine lower each relay primitive function. Here A1 just refers to the former. Does that make sense?

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: these buffers should ideally be compiler flags (and same goes to flags such as --executor, -runtime). However, we are currently using the target to hold such flags. Depending on the progress of the activity, where this flags end up will change :)

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Yes, I ll add it to the text tomorrow :)

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Oops, yes we've decided to remove "Integer uid" (it became redundant after some thinking). I ll adjust the text tommorow.

The pool_name is the assigned name of the pool (as passed from the tvmc interface).

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: It does not because memory planning algorithm is supposed to populate the pool_name and pool_offset of each BufferInfo object.

The integer of the second argument is to provide user provided size constraint (if any) for each pool to be used by the algorithm.

manupa-arm:
Thanks @tkonolige  for taking a look at this.

As per the pre-RFC discussion, 
There is not anything preventing us from integrating unified static memory planner (apart from doing actual work for that of course :) ) for graph executor as long as relay "main" function is lowered to TIR before creating the JSON. Thus, the design does not block such integration.
For VM, Im not sure how useful is "static" memory planning would be.

Drawbacks -- The relay "main" function has to be lowered to TIR to use the USMP. (I ll add a section)

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Done

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: I've done the changes and added some text after that to describe the pool_name

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Done. added some text after this to describe this.

tkonolige: I think functions that modify their input are really confusing in apis. Can we change this api to take const inputs and return the correct mappings?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+* src/tir/usmp/usmp.cc -- this is main intergration of USMP that exposes the full TIR --> TIR transformation as described.
+* tests/python/unittest/test_tir_usmp_*.py -- this where unittests for each of the passes and pass pipeline for USMP as a component will live.
+
+NOTE 1: All the above passes will have a mirror in the python.
+
+NOTE 2: to support tir.constants generally, we'll be enhancing the bound relay.constants to be lowered down to tir.constants to codegen. Those changes will appear through out the stack accordingly.
+
+# Drawbacks
+
+* The relay "main" function that describes the call order to operator PrimFuncs has to be described in TIR to be able to integrate the USMP into the respective executor codegen. However, we dont view this as a major problem as the relay "main" function could easily be lowered to TIR.
```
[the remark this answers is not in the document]
manupa-arm: How is it different from tir.call_extern or tir.call_cpacked? The main function just maintains calls. TVM cant anyway manage memories that are not exposed in the IR; it stops at calls. However, if such external calls needs memory passed in from main that needs to be allocated in TVM, then we could plan for that.

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: We could return the updated BufferInfo objects.

manupa-arm: Done!

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
+        String pool_name;`
+        Integer pool_offset;`
+    }
+```
+```
+Array<BufferInfo> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
+```
+The memory planning algorithm is expected to populate the assigned pool_name with the offset and return the updated array of BufferInfo objects. Additionally, the second argument provides size constraints for each pool (if any).
```
tkonolige, line 380:
```suggestion
The memory planning algorithm is expected to populate pool_name and pool_offset and return the updated array of BufferInfo objects. Additionally, the second argument provides size constraints for each pool (if any).
```

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface of the memory planning algorithm is as follows :
+```
+    struct BufferInfo {
+        String name_hint; // this is the tir.buffer name
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
```
tkonolige, line 372: Does the memory planning algorithm have to assign each buffer to a pool in `pool_candidates`? Is this information not used after memory planning?

manupa-arm: As of now, once they are assigned with pool and offset, pool_candidates may not be used.

Correction : Array<Integer> pool_candidates --> Array<String> pool_candidates.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+* AoT TIR PrimFunc accepting pool buffers from the user.
+* All Operator functions accepting pool buffers.
+  * Each operator function should address using the correct offset in the correct pool buffer
+
+Special Parametric Inputs : 
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface of the memory planning algorithm is as follows :
+```
+    struct BufferInfo {
```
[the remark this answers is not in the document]
manupa-arm: It makes sense and aligns with our thinking as well.

This is something that will be passed from the user from the TVMC interface. However, TVM currently is lacking support to hold compiler(-wide) flags where we currently end up posting them as target_host attrs. During the runtime of the compiler, these values would be global constants, therefore we are kind of discussing (internally) best way to represent them generically in an extensible way when more such compiler-flags (e.g. --executor is another example here) are added (e.g., possibly as some sort of AttrRegistry) and hold them during a compilation run.

Other reason for not defining them here is that the TVM's schedulers may would like to have access them in future. 

If you have any plans/ideas to this effect, we'd love to hear them :) 

cc : @leandron @Mousius

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
+        String pool_name;`
+        Integer pool_offset;`
+    }
+```
+```
+Array<BufferInfo> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
```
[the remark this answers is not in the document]
manupa-arm: One of the assumption is that once the USMP, the scheduling is already decided for the operators.

However, we do understand the requirement.
To that end, this is something we refer to as memory-aware operator scheduling that should be done prior to USMP.

In most cases, this could be done using 'memory pressure' and using let-bindings in relay as relay pass to putting them to A-normal form that results in minimum 'memory pressure'. For advanced (or in other words compile-time tolerable situations), that relay pass could call into the above memory planning interface to retrieve actual allocations.

The reasoning behind that being mixing scheduling and allocations can easily be in-tractable. However, for exceptional cases schedulers are free to be implemented (compile-time in-sensitive) to determine the order by running complex allocation algorithm in-loop. Therefore, we feel this interface is general enough to support both use cases.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+
+       allocate(DepthwiseConv2d, int32, [ 401408 ]) {
+         for (i1: int32,  0 ,  58 ) {
+           for (i2: int32,  0 ,  58 ) {
+            for(i3: int32,0,128) {
+               PaddedInput[(((i1*7424) + (i2*128)) + i3)] = @tir.if_then_else(((((1<= i1) && (i1 < 57)) && (1<= i2)) && (i2 < 57)), (int16*)placeholder_7[((((i1*7168) + (i2* 128 )) + i3) - 7296)], 0i16, dtype=int16)
+             }
+```
+
+The above TIR snippet shows that two intra operator buffers PaddedInput, DepthwiseConv2d is not visible to Relay Graph Plan Memory to be shared.
```
areusch, line 56: nit: "are not visible for optimization by the Relay-level GraphPlanMemory approach."

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
areusch: we discussed this at the µTVM meetup. The current thinking is what @manupa-arm said above--scheduling operators is considered to be separate from memory planning. However, it is entirely reasonable for an upstream scheduler to invoke parts of the memory planner to determine whether reordering the schedule could be advantageous later on. @jroesch also has plans to enable an upstream pass to run an initial "return value" through a partial set of the downstream passes and examine the result of this. For instance, an upstream scheduler could produce an initial schedule and then invoke the remainder of the passes through Memory Planning. Then, it could examine the result of Memory Planning, potentially do this a few times, and choose the best solution as the true "return value" from scheduling.

@MichaelJKlaiberBosch let us know if this seems like a poor compromise to you guys. We certainly don't have to stick with this approach if there are more advantages to another.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+       ...
+```
+# Code Structure
+
+* src/tir/usmp/analysis/ -- this is where analysis pases of USMP will live
+* src/tir/usmp/transforms/ -- this is where transform pases of USMP will live
+* src/tir/usmp/usmp.cc -- this is main intergration of USMP that exposes the full TIR --> TIR transformation as described.
+* tests/python/unittest/test_tir_usmp_*.py -- this where unittests for each of the passes and pass pipeline for USMP as a component will live.
+
+NOTE 1: All the above passes will have a mirror in the python.
```
areusch, line 467: could you define "mirror" here? also s/the python/python\//

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
areusch: cc @jroesch regarding compiler input and output interfaces.

rfcs/0009_Unified_Static_Memory_Planning.md
areusch: it does seem like it would be useful to split the pre-memory-planning data structure apart from the post-memory planning data structure. perhaps:
- an attribute `"memory_constraints"` is expected on nodes as input to memory planning
- an attribute `"memory_placement"` is expected on nodes as an output

one concern with this approach is it may be slower if we need to copy all nodes involved. cc @jroesch for opinions on how to handle this sort of thing in the new TE-compiler world.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+
+* Assumption of local optimization : performing sharing inside the operator first and sub-subsequently sharing that workspace with inter-operator tensors, would be sub-optimal.
+
+Thus, for the embedded use-cases, we'd need a unified static memory planner that performs memory planning of all tensors holistically to achieve best memory utilization.
+
+# Goals
+
+G1. There would be no TVMBackendAlloc(/Free)Workspace calls generated for tir.allocates that could be evaluated at compile time.
+
+Currently, the TVM codegen and the AoT executor relies on TVMB(A/F)W calls to increment/decrement a pointer of user provided workspace buffer. By the end of this set of work, if the backend uses Unified Static Memory Planning, there should not be TVMB(A/F)W calls rather correct offset in to the user provided buffer should be codegen'd for allocates that could be evaluated at compile time. The dynamically sized allocates will remain untouched, thus will be lowered as usual.
```
areusch, line 66: could you specify how you'll determine whether a TVMBAW can be eliminated? i think it should roughly be: can the size argument be evaluated at compile-time?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+Special Parametric Inputs : 
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface of the memory planning algorithm is as follows :
+```
+    struct BufferInfo {
+        String name_hint; // this is the tir.buffer name
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
```
areusch, line 371: how does USMP determine this set without doing some memory planning? e.g. is it expected to do lifetime analysis of the buffers and then mark those alive at the same time as "conflicts"? would it be better to encode the linearized alloc/kill info from StorageRewrite here?

areusch (commented):
@manupa-arm apologies for the delay and thanks for bringing this RFC to the table! I've left some initial comments/questions.

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
jroesch: I am trying to write down some guidelines on where to store these things, my 2c is that we should add Attributes to the IRModule to also use as a per compilation place to store global data instead of keying them off target. Is there any reason this wouldn't work for you? I think we should move in this direction so all the useful state for compilation is readable directly from the compilation unit.

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
jroesch: I am working on a set of RFCs which distill all the pain I felt removing the compile engine. I think users should choose ordering before we plan, entangling passes with different concerns creates some of the problems which we are feeling today, for example the current static memory planner invokes device planning directly and there is no (currently) easy way to split them into pieces. 

If you want to adjust the scheduling you should write a pass to plan, then analyze the plan, rewrite the program and invoke planning again instead of baking these details into the planner. The one shot mentality has led to a lot of super fragile code that only works under a single phase ordering for a single executor.

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Yes, it is about liveness conflicts.

Have a look here ?
https://github.com/apache/tvm/pull/8468/files

I feel linearized alloc/kill is raw information used in the process to generate conflicts. Therefore, having the sign-off via liveness conflicts are better for developement and integration of planning algorithms that does not need to know about a specific alloc/kill tick.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+    Start Date: 2021 June 1
+    RFC PR: #0009
+    GitHub Issue: https://github.com/apache/tvm/issues/8404
+
+# Background
+
+Currently, given a ML model primarily TVM will generate two main artifacts :
+
+* A1 : Description of the sequential execution of operators :
+  1. If the "executor" is "graph", this would be a JSON
```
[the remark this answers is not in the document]
manupa-arm: Sorry, Im not sure that I follow. Do you mean 'define' or 'specify'?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+  2. if the "executor" is "aot", this would be a main function describing call graph of operators
+  3. if the "executor" is "vm", this would be a series of VM bytecode instructions
+* A2 : library of operators (in the form of runtime.Module)
```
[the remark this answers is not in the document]
manupa-arm: Done

tkonolige (approved):
Thank you for answering my questions @manupa-arm

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
MichaelJKlaiberBosch: Thanks for the detailed answers.
@areusch @manupa-arm
 In general we are okay with separation scheduling an memory planning problem. From our point of view there could be cases where the memory planner can change the order of the callnodes is of advantage. This, however, depends a lot on the topology of future neural networks. For simple feed-forward NNs this might not be required. More complicated NNs, e.g. as a result of NAS this might be more relevant.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,473 @@
...
+
+A1 is generally created out of lowering the "main" relay function and A2 is created lowering fused relay primitive functions → TIR PrimFuncs → C or LLVM artifacts of the operator library.
+
+### Is there some sort of memory planning already being performed ?
+
+Yes, there is.
+
+For A1, the inter-(fused) operator tensors are visible in the "main" relay function. Thus, there exists currently a Relay level pass known as "GraphPlanMemory" that works on the Relay IR to share the space used by tensors which are not live simultaneously and are visible between (fused) operators . Currently, the said pass will use Shared Memory Buffer Object memory planning scheme (See https://blog.tensorflow.org/2020/10/optimizing-tensorflow-lite-runtime.html) to perform the planning.
```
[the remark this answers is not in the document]
manupa-arm: Done

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Done

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Done

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: Done

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
electriclilies: I agree with @jroesch on this, I think that it would be great to try to stope the information in the IRModule attrs. This would also be consistent with some of the cleanups we are currently doing -- I am starting to try to push the info stored in the TargetMap data structure into function attributes.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+       biases = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
+       ...
+       <compute>
+       ...
+```
+##### Step 2 : Run an analysis pass to populate a Map<tir::Var, BufferInfo> that contains buffer information as defined above (See the struct BufferInfo).
+
+##### Step 3 : Use the updated Map<tir::Var, BufferInfo> to generate Array<BufferInfo>, Map<String, Integer> pool_sizes
+
+##### Step 4 : Call the provided/default algorithm (void (*foo)(Array<ByfferInfo> buffers, Map<String, Integer> pool_sizes) to populate pool_id and pool_offset.
```
electriclilies, line 428: another typo: ByfferInfo -> BufferInfo

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
electriclilies: I think it is not different from tir.call_extern or tir.call_packed, never mind

electriclilies (approved):
Overall, looks good to me, found a few typos! Thanks for the detailed write up.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+    }
+```
+# Reference-level explanation
+
+## Overview
+
+This should be a IRModule (TIR) → IRModule (TIR) pass.
+
+Inputs : 
+* AoT TIR PrimFunc ( the control function describing the call graph to operators)
```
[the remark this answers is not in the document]
manupa-arm: Hi Chris,

> In the unified lowering flow we may choose to introduce explicit storage and tensor allocations into the relay AST, and then hoist memory planning out of the executors into the unified lowering flow.

> [IRModule(relay)] 
-> transforms (incl. mem. planning) -> [IRModule(relay)]

I dont see much value in doing memory planning at this level as the intra-operator allocate nodes are not visible. Can you elaborate ?

> With customization in the lowering flow, executors such as AoT can allow per target planning, and then generate code by matching allocation nodes in the relay AST and lowering them directly to TIR when building the main function.

Can you explain what do you mean by allocation nodes in the relay AST ? 
In our view, relay is a pure functional language that is designed as an IR to represent operator-level info. It feels wrong to do memory planning at this level.

I guess Im more interested in knowing why graph and AoT, both cannot lower the main function to TIR before performing executor specific lowering ( that is generating the JSON or main function  -- that is the executor specific lowering).

A side point to this is -- as we have discussed in the discuss post -- "I think going forward graph executor might be able to load a packed function of the tvm_main instead of json – it’ll be less confusing as how the graph executor runtime is positioned as of today which is more of a (a very thin – as its supposed to be :slight_smile: ) middleware that connect the graph json and the compiled operator library".

Having said that, I can see this work enabling a path (to extend) towards that – though we only plan to create USMP component that is a TIR IRModule → TIR IRModule which we initially test and support for the AoT executor."

https://discuss.tvm.apache.org/t/rfc-unified-static-memory-planning/10099/2

> with transforms including the ability to customize the transformations applied for different hardware targets. Moreover, transformations are applied at each stage of lowering and one or more of those transforms may include memory planning.

I dont see why this cannot happen if we have the full program in TIR.

> If we push forward with this RFC without co-designing it with unified lower, my concern is that we continue to kick the problem down the road and it will make unifying the lowering and planning for all executors more difficult.

The existence of different executors are for different use-cases, therefore there is only so much unification we could do. IMO, we should only have two executors AoT and JIT (VM). Moreover, I dont feel static memory planning is applicable to the latter. I'd view graph executor as an application that support RPC for AoT that could be used in a tuning process (Refer to the earlier comment on the importance of using a main function instead of JSON).

I'd like to hear why designing a static memory planner that works where the full program is expressed in TIR creates a divergence at least between graph and AoT.

@jroesch @areusch @tqchen @mbaret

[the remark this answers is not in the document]
csullivan: >In our view, relay is a pure functional language that is designed as an IR to represent operator-level info. It feels wrong to do memory planning at this level.

I believe @jroesch's RFC on unified lowering can help clarify this point some when it lands soon. My take is that overall we would like to move to a world in which the IR is progressively lowered from initially an operator only representation to something closer to hardware, including to TIR for AoT, with target customizable transformation available at every progressively lowered step. 

>I dont see much value in doing memory planning at this level as the intra-operator allocate nodes are not visible. Can you elaborate ?

Partial memory planning can be useful at this level. Let's take the constrained resource (SRAM) scheduling as an example, where SRAM is large enough to hold some full weight tensors. Assume weight pinning in SRAM is a special case of prefetching, and consider scheduling prefetch copy nodes that move weight tensors from DDR to SRAM storage. Depending on how many weights are pinned from the beginning and prefetched at various points in the topologically ordered execution, the amount of SRAM available for intra-operator scratch will change. In unified lowering this SRAM "stack" size could be provided to the TECompiler when doing subgraph scheduling. If this subgraph contains, for example, multiple convolutions and a striping technique is employed, the available SRAM scratch can be used as a constraint when an autoscheduler is searching the schedule space of this complicated subgraph. After scheduling another layer of transformations can occur which can include full / unified memory planning of inter- and intra- op storage.


> The existence of different executors are for different use-cases, therefore there is only so much unification we could do. IMO, we should only have two executors AoT and JIT (VM), conceptually (that is not to say graph executor is not important). 

We can consider a case in which everything is unified behind a MemoryPlan->VM->AoT lowering flow in which the AoT TIR main is generated from the relax VM representation. In this case the resulting AoT TIR also supports dynamism. But even before this we could stage the effort and make early strides towards unifying the lowering flows such that different executors utilize the same common planning infrastructure,
```
MemoryPlan -> VM
          \-> AoT
          \-> Graph
```

>I dont see why this cannot happen if we have the full program in TIR.

I'm not saying that we shouldn't be able to do final memory planning in a full program TIR. Unified lowering should support this. The argument I'm making is that supporting unified static memory planning (USMP) shouldn't be available to AoT only. My feeling is that the approach to USMP should be co-designed as part of the approach to unified lowering so that all executor paths can benefit from the ability to plan inter-op and intra-op storage together.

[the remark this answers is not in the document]
manupa-arm: Hi Chris,

> I believe @jroesch's RFC on unified lowering can help clarify this point some when it lands soon. My take is that overall we would like to move to a world in which the IR is progressively lowered from initially an operator only representation to something closer to hardware, including to TIR for AoT, with target customizable transformation available at every progressively lowered step.

Conceptually this is sensible, Im worried we will mix the responsibilities of Relay and TIR, where the former is a functional language and TIR has imperative properties (Stmts) that encapsulate Exprs. I guess my question is why do we say TIR is for AoT ?, In our view, we should not bypass TIR in the lowering, even for the main callgraph lowering that we call as specialized executor lowering. In that world, everything will get represented in TIR (not the just AoT executor).
cc : @tqchen @mbrookhart 

> Partial memory planning can be useful at this level. Let's take the constrained resource (SRAM) scheduling as an example, where SRAM is large enough to hold some full weight tensors. Assume weight pinning in SRAM is a special case of prefetching, and consider scheduling prefetch copy nodes that move weight tensors from DDR to SRAM storage. Depending on how many weights are pinned from the beginning and prefetched at various points in the topologically ordered execution, the amount of SRAM available for intra-operator scratch will change. In unified lowering this SRAM "stack" size could be provided to the TECompiler when doing subgraph scheduling. If this subgraph contains, for example, multiple convolutions and a striping technique is employed, the available SRAM scratch can be used as a constraint when an autoscheduler is searching the schedule space of this complicated subgraph. After scheduling another layer of transformations can occur which can include full / unified memory planning of inter- and intra- op storage.

Isn't this similiar to what could be done using https://tvm.apache.org/docs/api/python/te.html#tvm.te.Schedule.cache_read ? I guess my query is why are we adding Relay AST nodes to do this while we can do this with less changes and holistically in TIR ?

cc : @mbaret 

By doing so we need to duplicate most logic in scheduling primivites and TIR IR nodes to Relay as well. Do we know a strong reason to perform such a duplication ? It is not immediately obvious how this aligns with the goals of the unified lowering.

> We can consider a case in which everything is unified behind a MemoryPlan->VM->AoT lowering flow in which the AoT TIR main is generated from the relax VM representation. In this case the resulting AoT TIR also supports dynamism. But even before this we could stage the effort and make early strides towards unifying the lowering flows such that different executors utilize the same common planning infrastructure,

> MemoryPlan -> VM
          \-> AoT
          \-> Graph

Does supporting dynamism require a relay-level memory plan ?

My argument is unlike lowering it from directly from relay, if we unify the lowering after the whole program is expressed in TIR, it would avoid duplication -- which I think is one of the goals in the unified lowering refactor.

```
Relay --> TECompiler --> TIR (main is also in TIR) --> MemoryPlan --> AoT
                                                                                                           \--> Graph
                                                                                                           \--> VM
```
It'd be interesting to know what relay constructs that are used in VM is not possible to be expressed in TIR. If we can fully express the dynamism in TIR, I'd say we should not bypass the TIR lowering for any executor. If we are not bypassing the TIR lowering for any executor (operators or control code), as stated before, doing a memory plan at relay level seems a bit redundant. Its entirely possible, I might be missing some info on VM, thus feel to enlighten me :).

> Im not saying that we shouldn't be able to do final memory planning in a full program TIR. Unified lowering should support this. The argument I'm making is that supporting unified static memory planning (USMP) shouldn't be available to AoT only. My feeling is that the approach to USMP should be co-designed as part of the approach to unified lowering so that all executor paths can benefit from the ability to plan inter-op and intra-op storage together.

We agree with this fully and additionally we are saying we should not bypass 'full program TIR' for any executor, if possible. All, we are proposing here is Full Program TIR --> Full Program TIR transformation that could be attached to executor codegen pipeline, if each executor have that state (currently the graph bypasses this, it goes from Relay --> JSON directly) in the pipeline (I dont see a strong reason why they should not have that, at least yet). Then, its incremental work to attach it to each executor -- In a similiar way we do TECompiler refactor (attaching the TECompiler for each executor).

cc : @Mousius @mbaret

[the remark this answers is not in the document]
csullivan: Hi Manupa- we've been having some conversations internally trying to sync up on the unified lowering flow to understand how the USMP RFC and unified lowering can work together. @areusch will follow up with some comments. Overall I think it sounds like we have the same goals and if we can arrive at design in which only incremental refactoring is needed, like with updating AOT to use the TECompiler, then we're good. I'll also follow up directly with some of the questions you raised in response in another post, just didn't want to let the conversation lapse.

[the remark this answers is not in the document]
areusch: hi Manupa,

@jroesch @csullivan and I discussed this at length over the last two days in conjunction with @jroesch's plans to unify the Graph, AOT, and VM lowering pipelines. The root of the concern was over three issues:
1. it may be of interest for some users to explicitly specify which memory pool a Relay operator shall place its results. How shall TVM interpret such requests?
2. What if memory planning impacts scheduling (e.g. you may want to choose a different schedule because memory does not fit)?
3. As a side effect of unifying the lowering flows, it will be desirable to also unify memory planning across the three. Given the VM supports additional features over AOT/Graph executors (e.g. dynamic shapes, potentially additional control flow such as recursion), it may not be possible to perform static memory planning on all tensors in such a world. We want to ensure such a unified flow is possible without being overly complex and while still supporting features.
4. There was some confusion as to how GraphExecutor/AOTExecutor may both use USMP.

Point #1 can be handled with a Relay-level annotation on either the function or variable which can then be moved into TIR to drive the memory planner.
Point #4 is handled by writing a pass to translate the USMP-planned AOT top-level function into a GraphExecutor graph.json when possible. I've clarified this with @jroesch and @csullivan.

To point #2: there are a couple of classes of this problem: 1) when a schedule must be discarded because a single buffer always doesn't fit into RAM; 2) when a particular configuration of a schedule can fit into RAM, but not all. Ultimately, this problem can always be addressed by iteratively running the compilation pipeline starting from scheduling and discarding the results after memory planning if they don't work out. However, this solution is very slow so we prefer to avoid it. There are a couple of ways to do this:
1. introduce memory constraints at schedule time so that it's obvious when a buffer is far too big for the available memory
2. We can model allocations before scheduling, like the VM does. this can allow us to discard schedules in class 1 before doing expensive memory planning.
3. we may be able to leverage AutoTIR to "tweak" a schedule (e.g. reduce the intermediate scratchpad usage) when that usage is proportional to an inner/outer loop nest (this solution re

To point #3: After some discussion, it appears that the planning done in the VM and USMP have something in common: they both create allocate nodes to model the graph-level tensors in TIR. Since USMP relies on building an abstract representation of allocate/free events, it should be possible to continue to build this after unifying the flows based on either the AOT top-level function or a similar function produced by lowering the VM-optimized Relay module to TIR. Therefore, the static tensors used in a TIR IRModule should still be able to be planned using USMP after unifying the flows. The dynamic tensors may need to be handled separately, but in general doing much ahead-of-time planning for tensors with data-dependent lifetime (e.g. when handling recursion) is difficult.

Finally, related: @jroesch also mentioned that in unifying the flows, he wants to move to the world where a conceptual equivalent of the VM buffer/storage nodes are injected pre-scheduling. There will be some work in doing this to ensure that they are fully compatible with the memory pool concept here, but at first glance they appear to be broadly compatible.  The VM carries a somewhat similar concept of a per-device dynamic allocator, which is conceptually similar to that which may be required if we were to consider at a later time supporting dynamic models in the C runtime. We think these concepts are similar and can be unified moving forwards. 

Here our intent isn't to choose any one implementation over another but purely to ensure we can reduce the conceptual complexity of the compiler over time while retaining flexibility of the compiler. My thoughts are that adopting this approach (and therefore, going ahead with USMP now) should allow others to begin building the static memory planning algorithms while we are working to simplify the compiler, and this will ensure we can still continue to support any memory planning algorithms developed on top of USMP in the medium-long term.

Please let me know if this makes sense to you! I think this should allow everyone to continue working without stepping on each others' toes excessively.

-Andrew

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
manupa-arm: Hi @jroesch @electriclilies,

Yes, that sounds like a good idea.

Only concern is these are constant inputs in the compilation flow -- that are not supposed to get mutated in passes.
But, yea it could still work and the passes will not need anything more than IRModule -- that seems attractive. Let me know if you have anymore thoughts around this.

If we all seem to agree, yes we can introduce the 'PoolInfo' structure here and add as an Attribute to the IRModule.

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
manupa-arm: Hi @areusch @csullivan,

Thanks for the clarification and it makes sense for us.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+### TVMC
+
+
+```
+tvmc compile my_model.tflite --executor=aot --output-format=mlf --target=c
+```
+
+ ### Codegen'd artifacts
+
+
```
areusch, line 90: could you add a note saying that this interface is just to demonstrate the conceptual parameters introduced by USMP, and that the final interface is specified in other RFCs/

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+## Overview
+
+This should be a IRModule (TIR) → IRModule (TIR) pass.
+
+Inputs : 
+* AoT TIR PrimFunc ( the control function describing the call graph to operators)
+* All Operator Functions
+* the maximum size for each pool We could use "pinned_memory" (see below) to tag buffers with suggested priority order determined by the scheduler.
+
+The idea is USMP will try to pool them using the preferred "pinned_memory" and fallback whenever the size is exceeding the user provided max size for each pool (if any)
```
areusch, line 356: could you elaborate on "fallback"? Presumably this would be configurable--if the user doesn't want USMP to fallback when it exceeds memory boundaries, we should be able to configure this, no?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+# Reference-level explanation
+
+## Overview
+
+This should be a IRModule (TIR) → IRModule (TIR) pass.
+
+Inputs : 
+* AoT TIR PrimFunc ( the control function describing the call graph to operators)
+* All Operator Functions
+* the maximum size for each pool We could use "pinned_memory" (see below) to tag buffers with suggested priority order determined by the scheduler.
```
areusch, line 354: presumably you mean: "A description of memory pools and their maximum sizes (e.g. a map from pool_name to maximum size)." Can you update to reflect this?

rfcs/0009_Unified_Static_Memory_Planning.md
areusch, line 356: could you elaborate on "fallback"? Presumably this would be configurable--if the user doesn't want USMP to fallback when it exceeds memory boundaries, we should be able to configure this, no?

areusch, line 356: also can you clarify "fallback" here as well?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+
+The current proposal for the interface of the memory planning algorithm is as follows :
+```
+    struct BufferInfo {
+        String name_hint; // this is the tir.buffer name
+        Integer size_bytes;
+        Integer alignment;
+        Array<BufferInfo> conflicts; //the conflicting bufferinfo objs
+        Array<Integer> pool_candidates;`
+        String pool_name;`
```
areusch, line 374: nit: delete the ` or use it as a reference in the explanation below.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+* AoT TIR PrimFunc accepting pool buffers from the user.
+* All Operator functions accepting pool buffers.
+  * Each operator function should address using the correct offset in the correct pool buffer
+
+Special Parametric Inputs : 
+* function : The algorithm to be used for planning From a component PoV, the algorithm is a special input with a defined interface.
+
+The current proposal for the interface of the memory planning algorithm is as follows :
+```
+    struct BufferInfo {
```
areusch, line 368: now that we are aligning towards placing things as attributes in IRModules, could you separate out the input data structure to the planning algorithm from the data structure which would presumably be placed onto the buffer as an attr? e.g. pool_name and pool_offset should form a new struct which is the value of some e.g. `pin_location` attribute on the buffer.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+        Integer pool_offset;`
+    }
+```
+```
+Array<BufferInfo> (*foo)(Array<BufferInfo> buffers, Map<String, Integer> pool_sizes)
+```
+The memory planning algorithm is expected to populate pool_name and pool_offset and return the updated array of BufferInfo objects. Additionally, the second argument provides size constraints for each pool (if any).
+### Special Considerations :
+
+* tir.constants : TIR does not have the ability to represent constants – which is limiting and often leads to having side-channels to carry constants between TIR compiler passes including this one.
```
areusch, line 384: perhaps link to other RFC here

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+       tir.func_attr({"global_symbol":"my_model_fused_op2", "tir.noalias": True})
+       intermediate_tensor_1 = tir.allocate(..., pinned_memory=[1, 2])
+       intermediate_tensor_2 = tir.allocate(..., pinned_memory=[1, 2])
+       weights = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
+       biases = tir.allocate_const(..., pinned_memory=["itcm", "flash"])
+       ...
+       <compute>
+       ...
+```
+##### Step 2 : Run an analysis pass to populate a Map<tir::Var, BufferInfo> that contains buffer information as defined above (See the struct BufferInfo).
```
areusch, line 424: could you sketch this out briefly?

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -0,0 +1,476 @@
...
+
+##### Step 3 : Use the updated Map<tir::Var, BufferInfo> to generate Array<BufferInfo>, Map<String, Integer> pool_sizes
```
areusch, line 426: how are the Vars assigned to memory pools?

areusch (changes requested):
hi @manupa-arm, just a few more questions. if they're answered satisfactorily by @jroesch @mbs-octoml @tqchen and they approve this feel free to merge without my approval.

areusch:
@manupa-arm could you address my last round of comments? i think we're good to merge after that

manupa-arm:
@areusch I got some cycles to spend on this. I've updated the RFC addressing your comments and reflecting the changes discussed here as well : https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md.

PTAL when you get some time. TIA.

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -348,12 +350,27 @@ tvmc compile my_model.tflite --executor=aot --output-format=mlf --target=c
...
-* All Operator Functions
-* the maximum size for each pool We could use "pinned_memory" (see below) to tag buffers with suggested priority order determined by the scheduler.
+Inputs :
+* IRModule containing 
+    * AoT TIR PrimFunc (the control function describing the call graph to operators)
+    * All Operator Functions
+    * Each tir.allocate in the IRModule annotated with candidate pools ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md))
+
+
+```
```
areusch, line 360: er, this is kind of just here without much explanation. it makes sense in light of the comments threads here..but can you add something more explicit to indicate that this is an input? even if we are not quite sure how this is going to land and just defer to that other RFC, it's helpful to state that here.

rfcs/0009_Unified_Static_Memory_Planning.md
areusch: can you address this here? some questions i don't see answered:
1. where are the "fallback" candidate_memory_pools passed in to the runtime?
2. how should the compiler know which candidate_memory_pools can be fallbacks for a given buffer (given that it does know a particular buffer needs to be accessible per-device)
3. do we need to add an additional field to `PoolInfo` to identify which devices can use it? it's fine if this is to be decided in a follow-up RFC, but can you indicate which one (if you have one in mind) or that one hasn't been chosen (if not)?

areusch (commented):
@manupa-arm i did another pass here. just a couple more clarifying questions on the overall approach. i'm okay with deferring some decisions to follow-on RFCs when they become more clear (e.g. I think progress on @jroesch TargetDevice work may influence some of the ways we encode memory pool accessibility per-device), but can we explicitly make it clear in the RFC when we are deferring something and specify which RFC it'll be decided under, if we know?

manupa-arm:
Hi @areusch ,

I have addressed candidate_memory_pool query now.

For you question around fallback : 

> where are the "fallback" candidate_memory_pools passed in to the runtime?

The fallback only happens the in the compilation time as per this RFC. Therefore, by the time USMP is done, one pool will be assigned/decided to the tir.allocate.

> how should the compiler know which candidate_memory_pools can be fallbacks for a given buffer (given that it does know a particular buffer needs to be accessible per-device)

I have added a sentence to explain this. The core compiler should assign the tir.allocate in each PrimFunc with pool it can access because each PrimFunc know which target it gets compiled to and each pools know which target could access them. Initially it will take the priority order the user provides to the TVMC interface.

> do we need to add an additional field to PoolInfo to identify which devices can use it? it's fine if this is to be decided in a follow-up RFC, but can you indicate which one (if you have one in mind) or that one hasn't been chosen (if not)?

Im pretty sure it is listed as Map<Target,String> target_access; // 'rw' or 'ro' in the PoolInfo.

rfcs/0009_Unified_Static_Memory_Planning.md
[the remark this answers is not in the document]
areusch: sorry i mean that by `"executor"` here, you mean the value of the `"executor"` key in MLF `metadata.json`. However, that's a micro-specific (for now) artifact. To make the RFC easier to follow, suggest either including a pointer so that a reader could determine this, or just saying something less specific like "If building for the graph executor."

areusch (commented):
@manupa-arm thanks. can you do a pass through the remaining items here (e.g. @electriclillies had some clarifying suggestions to the text).

> For your question around fallback :
> 
> > where are the "fallback" candidate_memory_pools passed in to the runtime?
>
>The fallback only happens the in the compilation time as per this RFC. Therefore, by the time USMP is done, one pool will be assigned/decided to the tir.allocate.

sorry but i mean; how do we determine a pool to be a fallback pool?

>> how should the compiler know which candidate_memory_pools can be fallbacks for a given buffer (given that it does know a particular buffer needs to be accessible per-device)
> 
> I have added a sentence to explain this. The core compiler should assign the tir.allocate in each PrimFunc with pool it can access because each PrimFunc know which target it gets compiled to and each pools know which target could access them. Initially it will take the priority order the user provides to the TVMC interface.

thanks! can you add the ordering bit to the RFC?

>> do we need to add an additional field to PoolInfo to identify which devices can use it? it's fine if this is to be decided in a follow-up RFC, but can you indicate which one (if you have one in mind) or that one hasn't been chosen (if not)?
> 
> Im pretty sure it is listed as Map<Target,String> target_access; // 'rw' or 'ro' in the PoolInfo.

ah--you're correct.

manupa-arm:
@electriclilies -- sorry I missed your suggestions -- corrected them now.

@areusch -- addressed the comments now.

areusch (approved):
thanks @manupa-arm !

areusch:
@manupa-arm please open a tracking issue for this one. excited to see this land!

PR #10, opened 2021-07-14
Additional Target Hooks RFC

Mousius:
This is the an initial RFC for adding additional hooks onto the `Target` to
allow splitting up some of the compile flow but also unifying the
registration of these additional functions.


Mousius:
CC: @tqchen @jroesch 

I think this overlaps with the TE Compiler work, would be good to get your feedback :smile_cat:

jroesch:
Thanks for the work on this, after thinking about it for a while I believe the proposed hooks runs counter to goals of the TECompiler and "unified lowering" refactor that we've been working on in pieces. Our design goal is to not allow arbitrary customization of "lowering"  but instead seal it behind a stable interface. The lowering process should be a straight forward mapping from TE -> TIR, and then any necessary customization should be possible in resulting passes which are allowed to view the entire program including both the Relay code and TIR code.

The goal is to remove the need for many slightly different code paths through the central parts of the compiler and instead provide standard interfaces that allow targets, and users to customize the compiler without having to touch any of the flows or modify the data structures.  The proliferation of customizations in the old compile engine for example makes it nearly impossible to refactor the compiler in its current state, and has lead to the world where we need lots of per-flow customization because the compiler is insufficiently factored into a common prefix of code. 

I plan on providing a longer form document on this hopefully this week that argues this more in depth but I believe the tenants should be customization of the compiler happens via passes, passes only can produce IRModules, and all data is stored in the IRModule. Today's compiler requires a little more work to get to this world, but I think we should push in this direction.

Now that doesn't mean the desire to have per target behavior is necessarily bad, but I think instead of overriding the lowering code with arbitrary functions we could provide hooks that allow the insertion of target specific passes at the "correct" time in the compiler flow. In this world the proposed hooks should return a `Pass` object which can be inserted after the appropriate phases. 

Happy to discuss more, and have others such as @electriclilies @csullivan @mbs-octoml

Mousius:
Hi @jroesch,

It'd be great to discuss this further, as there's some interesting points you've raised.

> The lowering process should be a straight forward mapping from TE -> TIR, and then any necessary customization should be possible in resulting passes which are allowed to view the entire program including both the Relay code and TIR code.

In the case of library adaptations such as [CMSIS-NN](https://github.com/apache/tvm-rfcs/pull/15), translating through Relay->TE->TIR isn't required as we can do a more straight forward translation directly from Relay->TIR without any of the other pass overheads. In effect, our hook here is a single `Pass` to run over any marked functions rather than aggregating other `Pass`es so conceptually they're similar.

> The goal is to remove the need for many slightly different code paths through the central parts of the compiler and instead provide standard interfaces that allow targets, and users to customize the compiler without having to touch any of the flows or modify the data structures. 

I believe we're on the same page here, the eventual goal being to register and use the hooks directly as target properties that can be accessed as part of a standard flow. The hooks proposed here are slightly larger than you envisaged but should allow a standard and straight forward interface for users to customise the flow of the compiler. The initial implementation isn't as clean as I'd like due to the issues you've raised already with the current compile engine.

> I plan on providing a longer form document on this hopefully this week that argues this more in depth but I believe the tenants should be customization of the compiler happens via passes, passes only can produce IRModules, and all data is stored in the IRModule. Today's compiler requires a little more work to get to this world, but I think we should push in this direction.

The hooks proposed could in fact be such `Pass`es, going `IRModule` -> `IRModule` rather than taking a Relay `Function`. As stated above, it would be more optimal to be given a `IRModule` without TIR, just with the Relay attached.

> Now that doesn't mean the desire to have per target behavior is necessarily bad, but I think instead of overriding the lowering code with arbitrary functions we could provide hooks that allow the insertion of target specific passes at the "correct" time in the compiler flow. In this world the proposed hooks should return a Pass object which can be inserted after the appropriate phases.

I'm interested to understand how this would appear to users in a way that allows them to compose the compiler passes efficiently, being able to write a pipeline of passes seems more intuitive than providing many hooks to insert at different times? Potentially the hooks should return an array of passes or one aggregate pass to ensure the contract is upheld?

manupa-arm:
Hi @jroesch @Mousius ,

I think having Pass'es (instead of functions) could also work.

I guess what we are after is the ability to assemble and re-use TVM's compilation passes on a target basis.

https://github.com/apache/tvm/pull/8110#discussion_r639559530
Continuing the above comment, this also removes various places otherwise devs need to insert custom passes. Therefore, centralizing them closer to the target seems much cleaner to me. Therefore, it becomes straightforward to follow what passes are being run in the lowering for a given target while also not running passes that are not relevant (that are currently run anyway due to them having identity effect on IRModule in the absense of special attr/nodes/vars in the IRModule). 

> I believe the tenants should be customization of the compiler happens via passes, passes only can produce IRModules, and all data is stored in the IRModule. Today's compiler requires a little more work to get to this world, but I think we should push in this direction.

I think we broadly agrees on this. @Mousius ?

>  The proliferation of customizations in the old compile engine for example makes it nearly impossible to refactor the compiler in its current state, and has lead to the world where we need lots of per-flow customization because the compiler is insufficiently factored into a common prefix of code.

While we agree that its better to share the many passes that are required, but does it need to mandate a monolith of TIR passes being applied irrespective of the target ? We'd like to hear you thoughts (I think this has a connection to the so-called phases in lower implementation, right ?). While we could codify the requirement, is it absolutely necessary ? Also will it (having target based pass pipeline) preclude us from having the code factored into a common prefix ?

> but I think instead of overriding the lowering code with arbitrary functions we could provide hooks that allow the insertion of target specific passes at the "correct" time in the compiler flow.

The arbitrary function mentioned here could be a pass pipeline (and we could make it a requirement by making it a Pass as you suggested). However, it is interesting know if we can define a universally 'correct' time for the flow considering targets beyond CPUs and GPUs (w/o needing to do inverse lowering to recreate high-level abstraction that was forced lowered due to mandatory passes).  We would love to hear if you have an alternative suggestion to support target-based lowering

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
+Feature Name: additional-target-hooks
+Start Date: 2021-07-14
+RFC PR: apache/tvm-rfcs#10
+GitHub Issue: apache/tvm#8589
+
+# Summary
+[summary]: #summary
+
+In order to enable flexibility in how individual targets are lowered and built within TVM, this RFC proposes supporting additional hooks on the `Target` and that the target becomes the central place for such hooks, for example:
```
areusch, line 9: s/individual/externally-codegen'd/

areusch, line 9: nit: delete "supporting"

areusch, line 9: could you clarify what you mean by "central place for such hooks?" Specifically, qualify what is needed of a hook that makes it a candidate for adding to this target hooks map. Since we are going to a world where everything is a compiler pass, it would be great to briefly discuss the difference between this hook (used as a plug-in to LowerTE scheduler) and other passes a Target may want to add to the compilation flow.

Also, i think our discussion resolved towards that we don't want to encourage the hook to be the central way people make target-specific modifications to TIR. However, this text reads that way. Can you update it?

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+
+```
+TVM_REGISTER_TARGET_KIND("cmsisnn", kDLCPU)
+    .set_attr<String>("relay_to_tir", "target.cmsisnn.lower")
+    .set_attr<String>("tir_to_runtime", "target.cmsisnn.build");
+```
+
+This defines two new hooks as attributes on the target, referencing functions registered into the central TVM registry. In similar fashion, external generators (currently accessed directly in the compile engine) would be grouped with an appropriate `Target` as well:
```
areusch, line 17: clarify "external generators." how are these different from the other hooks you mentioned?

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+
+```
+TVM_REGISTER_TARGET_KIND("ethos-n", kDLCPU)
+    .set_attr<String>("relay_to_runtime", "relay.ext.ethos-n")
+    .set_attr<String>("constant_updater", "relay.ext.ethos-n.constant_updater");
+```
+
+Collecting all targets under the `Target` functionality and making it clearer which hooks apply to each target.
```
areusch, line 25: > Collecting all targets under the `Target` functionality
what do you mean by this?

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently to introduce an external code generator, the entire compilation pipeline must be recreated; this is necessary for some targets but in the case of simply re-using existing libraries or introducing a function call to use for an operator it can become more than is necessary. It also exists outside of the main `PrimFunc`, meaning it can't be inspected as part of the entire main graph; this limits the effectiveness of techniques such as memory planning. By introducing the hook `relay_to_tir`, which is similar to the default `lower` pass in that it returns TIR, it can be inspected by the memory planner and other analysis passes that only work at the TIR level. If all that is necessary is transforming into a flat `call_extern` (such is the case for the [CMSIS NN Softmax function](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c#L86)) then this can be left represented as TIR and be collected by the host code generation.
```
areusch, line 30: > Currently to introduce an external code generator,

perhaps include BYOC here for clarification

areusch, line 30: > it can become more than is necessary

what do you mean by this--what is burdensome exactly?

areusch, line 30: > outside of the main `PrimFunc`, meaning it can't be inspected as part of the entire main graph

this is AOT-centric. Can you generalize or contextualize this argument in terms of where the compiler is going?

areusch, line 30: > which is similar to the default `lower` pass in that it returns TIR

since we are now making this part of LowerTE, maybe you can rephrase so that it's clear that LowerTE uses this as a Relay function-by-function override for TIR generation.

areusch, line 30: > this can be left represented as TIR

I think you could state: "then the hook may simply return that TIR to be collected by the host code generation"

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+
+In the more complex case, we still want to take advantage of memory planning by using `relay_to_tir` and inspecting the liveness within the TIR graph, but instead want to generate out more complex calls (such as using the [CMSIS NN Structures](https://github.com/ARM-software/CMSIS_5/blob/def6f800f95661eb3451d317f7d0dde504f6020d/CMSIS/NN/Include/arm_nn_types.h#L81-L90)); the `tir_to_runtime` hook can be used to build our intermediary TIR into a Runtime module similarly to how the existing external code generation works. This allows writing of external code generators that also get the benefits of any intermediary analysis or transformation that TVM offers. Alongside being able to use the analysis passes, code generators can extend from existing host code generators, customising only the generation which is relevant to them and gaining maximum benefit from the existing optimisations made in TVM.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As a user, you can pick from additional hooks to bypass certain behaviours of the `Target`:
```
areusch, line 37: update to say "customize the lowering flow"

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This functionality is an extension of the existing use of `attr::kCompiler` to provide a hint that we can use to lookup attached target attribute, the compile engine and code generation flows can choose to store TIR and/or generate runtime modules based on the registered hooks.
+
+## Relay to TIR Hook
+[relay-to-tir-hook]: #relay-to-tir-hook
+
+This can be added to the TE Compiler by cross referencing the existing `attr::kCompiler` with the `TargetKind` registry:
```
areusch, line 107: i guess one dumb question here @jroesch @mbs-octoml: if we are making a generic interface to lowering Relay to TIR, why shouldn't AutoTVM schedule-based lowering also fit this interface? then all the test infra would be available to external codegen folks, which means we would have a much better chance writing a set of asserts to validate lowered TIR

rfcs/0010-additional-target-hooks.md
```diff
@@ -0,0 +1,155 @@
...
+This means the hook is integrated at a higher level and included in the compile flow without executors having to be aware of how these modules exist. See [Relay to TIR Hook](#relay-to-tir-hook) for how the `TargetKind` registry would be used.
+
+## Relay to Runtime Hook
+[relay-to-runtime-hook]: #relay-to-runtime-hook
+This would replace the existing `relay.ext.<target>` lookup in `compile_engine.cc`, essentially using the same logic as [Relay to TIR Hook](#relay-to-tir-hook) to cross reference with `kCompiler`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Different hooks are currently dealt with in quite disparate parts of the codebase which are being heavily refactored
```
areusch, line 141: i would also say that introducing custom TIR has the potential to add edge cases to the compiler which may uncover new bugs.

areusch (changes requested):
thanks @Mousius ! I did a thorough review in light of our discussion early last week. summarizing that discussion here for posterity:
- Ultimately we have a desire to maintain a single lowering pipeline for AOT, Graph, VM executors
- In the unified lowering pipeline, we hope to bring everything into a single IRModule, including the AOT main function. This allows most TIR-based analysis and transformation (e.g. memory planning, TIR-based operator fusion, etc) to be represented as compiler Passes
- In order for these passes to have something solid to work against, we generally want to consolidate the number of ways people generate TIR.
- However we realize the typical scheduling flow doesn't work for some Targets and agree with the shortcomings discussed here in requiring these Targets to reinvent the entire compliation pipeline. That seems worse than allowing some variance in generated TIR.
- Therefore, while ultimately we think all TIR transforms should be Passes, this particular case is essentially an override to the `LowerTE` pass introduced in the TE-Compiler refactor PR. Therefore, we support it being a hook with a well-defined signature.

xqdan:
This is a great disscussion here. Actually, we are supporting a DSA with TVM, let me share my practice.

1, We only re-use some of tvm relay or tir passes, less than 10 passes, such storage flatten, we don't need most of tvm passes, keep them in our flow means wasting compilation time.
2, We develop passes, and enhance tvm passes for our target, such stroage rewrite.
3, We develop a hybrid ir on tir, in which we can do unified memory allocation and schedule insnstruction across operators.

So we are excited to see tvm can do these in main line.

What we want to have:
1, customzation compilation flow for both relay and tir flow.
2, unified ir to view both graph and tir, supporting inter/intra pass developing.

rfcs/0010-additional-target-hooks.md
mbs-octoml: That sounds sensible. I need to pin Junru down for an hour or so to get a tour of AutoTVM to understand how subtle the extension points would need to be.

mbs-octoml:
This LGTM but I'm thinking going back to relay_to_tir being a Function->PrimFunc packed func as you implemented in the companion PR is better since it makes it easier for all the caching, call rewriting, dynamic shape handling and other bookkeeping to be shared. Have you tried out the pass approach and convinced youself that reuse is still possible?

Mousius:
Hi @mbs-octoml, sorry I missed a few replies! 

The reason I hoisted it outside of the `LowerTE` pass is that it effectively bypasses it anyway, `LowerExternalFunctions` in `te_compiler.cc` is a massive bypass which starts there and goes around most of TVM before it ends up as a `runtime::Module` right at the very end. Effectively what's happening is the "default" `relay_to_tir` is `LowerTE` and then `Target`s register their own variants. I think where @jroesch was going with the `(IRModule, Function) -> (IRModule, GlobalVar)` is that you will want to produce many `PrimFunc`s to implement one `Function` and thus you need the entire module, by just introducing this as a `Pass` we won't have to introduce a new signature specifically for this hook.

mbs-octoml:
> (IRModule, Function) -> (IRModule, GlobalVar)
I'm still is favor of this signature since it's a cheep and cheerful way to ensure we don't end up with N ways to implement the lower-and-rewrite-calls machinery embedded in te_complier. I think my earlier question still stands:
> Have you tried out the pass approach and convinced youself that reuse is still possible?

Mousius:
@mbs-octoml  / @areusch apologies for the delay, there was a fair amount to tweak in here to reflect all the discussions around this RFC. Please could you take another look as I think this is lined up now :smile_cat:

rfcs/0010-target-registered-compiler-flow-customisation.md
```diff
@@ -0,0 +1,186 @@
...
+```c++
+using FTVMRelayToRuntime = runtime::TypedPackedFunc<runtime::Module(const Function&)>;
+using FTVMConstantUpdater = runtime::TypedPackedFunc<Map<String, runtime::NDArray>(Expr, std::string)>;
+
+TVM_REGISTER_TARGET_KIND("ethos-n", kDLCPU)
+    .set_attr<FTVMRelayToRuntime>("RelayToRuntime", EthosNCodeGen)
+    .set_attr<FTVMConstantUpdater>("UpdateConstants", EthosNConstantUpdater);
+```
+
+Collecting all targets under the `Target` functionality (as opposed to registering additional `Target`s through the function registry using the namespace `relay.ext.`) and making it clearer which hooks apply to each target.
```
mbs-octoml, line 31: nit: '...makes it clearer which...'

rfcs/0010-target-registered-compiler-flow-customisation.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently to introduce an external code generator (otherwise known as [BYOC](https://tvm.apache.org/docs/dev/relay_bring_your_own_codegen.html)), the entire compilation pipeline must be recreated; this is necessary for some targets but in the case of simply re-using existing libraries or introducing a function call to use for an operator it can become more than is necessary; to implement an external code generator requires going directly from Relay to a `runtime::Module` and re-implementing any compiler passes and code generation functionality rather than being able to extend upon the existing compiler infrastructure.
```
mbs-octoml, line 36: nit: May be worth a pithy opening sentence saying we want to make BOYC more modular: instead of going from Relay to Module in one big step you can a) break it into phases and b) make use of existing transformations between phases. 

 Or at least that's the dumbed-down way I've been thinking of it.

rfcs/0010-target-registered-compiler-flow-customisation.md
```diff
@@ -0,0 +1,186 @@
...
+    .set_attr<FTVMRelayToTIR>("RelayToTIR", CMSISNNLowering)
+    .set_attr<FTVMTIRToRuntime>("TIRToRuntime", CMSISNNCodeGen);
+
+TVM_REGISTER_TARGET_KIND("ethos-n", kDLCPU)
+    .set_attr<FTVMRelayToRuntime>("RelayToRuntime", EthosNCodeGen)
+    .set_attr<FTVMConstantUpdater>("UpdateConstants", EthosNConstantUpdater);
+```
+
+## Relay -> TIR
+With this change, this path splits, depending on whether you wanted to generate a full `Module` or introduce some specific TIR nodes into the code generation flow; the addition of the `RelayToTIR` hook allows you to write trivial external TIR generators such as calling out to a third party library:
```
mbs-octoml, line 66: nit: I'd spell out that by this point every Function has a kTarget annotation thanks to AnnotateTarget, and that IRModule will hold all defns for all targets right up until the transition to Module.

rfcs/0010-target-registered-compiler-flow-customisation.md
```diff
@@ -0,0 +1,186 @@
...
+    ir_module_->Add(new_global_var, replacement_func);
+}
+```
+This is then registered on a target:
+```c++
+TVM_REGISTER_TARGET_KIND("woofles", kDLCPU)
+    .set_attr<FTVMRelayToTIR>("RelayToTIR", relay::contrib::woofles::RelayToTIR());
+```
+The signature for this hook is as the same as any other `Pass`, which takes an `IRModule` with `Function`s and returns an `IRModule` with transformed `PrimFunc`s.
+
```
mbs-octoml, line 92: nit: something like "The registered pass is responsible for both establishing the PrimFunc definitions (with any caching) and rewriting Relay calls to those functions. At this time we feel it's not worth worrying about code sharing between different custom passes". Just to capture the consensus on that.

rfcs/0010-target-registered-compiler-flow-customisation.md
```diff
@@ -0,0 +1,186 @@
...
+## TIR -> Runtime
+Extending from the above, a second hook is introduced to do further transformations from TIR -> Runtime named `TIRToRuntime`, this bypasses the default `target.build.X` and instead uses the registered `TIRToRuntime` build:
+```c++
+runtime::Module BuildWooflesHost(IRModule mod, Target target) {
```
mbs-octoml, line 96: nit: Make clear the IRModule can still be cross-target, right? Ie all of these hooks are responsible for doing the 'just look at defns for my target' filtering?

mbs-octoml (approved):
LGTM, just some nits for clarity.

We're working on cleaning up the py/cpp build split so your TIRToRuntime hook should indeed be on the unique choke point.

rfcs/0010-target-registered-compiler-flow-customisation.md
Mousius: Yip, eventually this should just be the unified `IRModule`, ideally it'd just be calling over these functions and they pluck the relevant functions to them. Essentially you'd end up with:

```
modules = [
  target->kind->TIRToRuntime(ir_mod, target)
  for target in targets
]
```

Right at the end, though I think that's beyond the scope of this RFC to get all the way there.

rfcs/0010-target-registered-compiler-flow-customisation.md
Mousius: Borrowed :smile_cat:

rfcs/0010-target-registered-compiler-flow-customisation.md
Mousius: Added a slightly vaguer line that they'll be annotated by a previous `Pass`, not necessarily prescribing `AnnotateTarget`.

rfcs/0010-target-registered-compiler-flow-customisation.md
Mousius: I have liberally taken that phrase and put it in :smile_cat: thanks!

Mousius:
Thanks @mbs-octoml, I've incorporated this back into the document :smile_cat:

mbs-octoml:
LGTM again (CI will go faster that way, right?)

Mousius:
@areusch is there anything outstanding from you on this RFC? It seems ready to merge after the changes you've requested.

areusch (approved):

areusch:
thanks for working through this with us @Mousius ! excited to see this land in TVM!

PR #12, opened 2021-07-23
[TIR] Vectorized buffer elements and indices

Lunderberg:
I put this together following [TVM PR#8528](https://github.com/apache/tvm/pull/8528), which resolved an issue with array access in the Vulkan runtime, but led to/exposed some inconsistencies in the TIR semantics for buffer indices. Vulkan/SPIR-V require all arrays to be typed, and doesn't allow type casts that would be permissible in C code, such as casting between `float32*` and `float32x2*`.  As a result, any vectorized load/store operations must act on an array whose elements are vectorized types.

Currently, many places in IR (e.g. [`Load::Load`](https://github.com/apache/tvm/blob/07243a89/src/tir/ir/expr.cc#L621) checking the number of output lanes) and codegen (e.g. [`CodeGenC`'s `LoadNode`](https://github.com/apache/tvm/blob/07243a89/src/target/source/codegen_c.cc#L719) loop over output lanes rather than element/index lanes) implicitly assume that all array elements have `lanes == 1`.

I couldn't find any documentation on intended semantics  for these multi-lane array indices, and so I wanted to propose, discuss, and document the intended semantics.

### DIFF
--- a/rfcs/0012-vectorized-tir-buffers.md
+++ b/rfcs/0012-vectorized-tir-buffers.md
@@ -1,6 +1,6 @@
 - Feature Name: vectorized-tir-buffers
 - Start Date: 2021-07-22
-- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- RFC PR: [apache/tvm-rfcs#0012](https://github.com/apache/tvm-rfcs/pull/0012)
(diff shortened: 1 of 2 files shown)


Lunderberg:
Closing, pending pre-RFC at https://discuss.tvm.apache.org/t/pre-rfc-vectorized-tir-buffers/10615

PR #14, opened 2021-07-31
[RFC] Pipeline Executor

huajsj:
Split relay graph into Subgraph then doing Subgraph Pipeline :RFC

In this RFC, we present a new framework for subgraph pipelining that first split relay graph into multiple subgraph/group operator s then doing subgraph/group operators scheduling

Discussion: https://discuss.tvm.apache.org/t/rfc-compute-graph-pipeline-with-new-subgraph-executor/9839


rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+
+The network pipeline compute effect
+![meta-schedule-workflow](../resources/pipeline-executor-pipeline.png)
+
+
+## 5. Drawbacks
+
+
+Pipeline executor currently need manually subgraph split and configuration construct,
+Further graph split feature would doing automatically split.
```
comaniac, line 344:
```suggestion
Pipeline executor currently needs manually subgraph splitting and configuration construction.
Further graph splitting feature would do automatically split.
```

I would say this is an important feature for this RFC, so it might be better to include possible solutions even it is not going to be implemented for now. Without this feature, I don't really believe pipeline executor itself would be useful for the community.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+
+## 6. Rationale and alternative
+
+
+Compute graph can get split into subgraph and pipeline execution can implement parallism
+when these subgraph have dependency relation.
```
comaniac, line 350: This statement is not related to "Rationale and alternative". In this section, you should include other possible solutions without pipeline executor to illustrate that pipeline executor is really necessary to achieve the goal.

comaniac (changes requested):
Thanks for the RFC. My general opinions:
- The whole picture of this RFC is not clear. If it only targets to a runtime executor, it is not a complete end-to-end solution. At high-level, you should present the ultimate goal, including graph partitioning and execution. Based on that, you could further break them down to several milestones so that we can track the progress while having a clear final picture.
- Please do proof-reading in advance. There are too many spelling and grammar errors to read the proposal smoothly.

cc @tmoreau89 @areusch @jtuyls

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [apache/tvm-rfcs#0014](https://github.com/apache/tvm-rfcs/pull/0014)
+- GitHub Issue: [apache/tvm#8596](https://github.com/apache/tvm/issues/8596)
+
+## 1. Summary
+
+
+This proposal introduces Pipeline Executor: A runtime executor that by scheduling
+splitted subgraph of relay graph in pipeline to implement task level parallism to
+reduce compute latency.
```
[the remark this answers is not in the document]
huajsj: about #2, compute latency, in blocking/serialize mode, let's say we get 2 data at same time and need to do compute, for every single data the processing time is 4ms, then first data latency is 4ms, second data latency is 4ms + 4ms = 8ms, in 4 stage pipe line mode,  every stage spend 1ms, then first data latency still is 4ms but second data processing time get reduced into 1ms + 4ms = 5ms, hence we can say latency get reduced.

about #1, yes this RFC only include executor, i will add the expected input part

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+with a smaller schedule unit like subgraph (a group of operator with dependency relation)
+to be more efficient to use SOC heterogenous hardware resource and get better performance.
+
+
+### Benefits of Pipeline Executor
+
+There are three benefit for Pipeline Executor
+
+Pipeline Executor provides:
+* Compute single network on  Multiple backend in parallel to improve performance.
```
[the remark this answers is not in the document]
huajsj: for example for a resnet18 compute, first we manually split the network into 4 subgraph, then use pipeline executor to build these four subgraph with target  llvm+cpu(0), opencl-gpu(0), opencl -gpu(1), VTA(FPGA), and generate the dependency relation,  when pipeline executor run the compute, the resnet18(now already get split into 4 subgraph) would get running in multiple backend in parallel(pipeline).

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+
+* With RPC help do ML network distribute computation cross multiple remote device
+
+* User can use Pipeline Executor to integrate pre-compute processing and pos-processing with
+  network compute together and compute in same executor.
```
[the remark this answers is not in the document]
huajsj: let's take yolov3-tiny deploy as a example, before run yolov3-tiny tvm build library, user need to prepare the data that include #1 image format transfer(for example YUV to RGB) #2 , image reshape to match network size #3. data normalization as network required, all these I called it as pre-compute processing,  after yolov3-tiny tvm build library run, user can get a [1, 3, 85, 13,13], [1, 3,85,26,26]  array, user need to parse these 2 array to NMS sort to get bunding box, this I called it post-processing.

Currently pre-processing and post-processing not handled/scheduled by TVM,  Pipeline executor provide such scheduling solution to involve these 2 part with network compute together to provide  complete solution for model deploy.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+
+### 3.3. Use pipeline_executor to build pipeline module with the said subgraph and configuration.
+
+Pipeline executor provide a build function to compile and save the compile output into disk,
+following is a example
+
+```python
+    with relay.build_config(opt_level=3):
+        pipeline_mods, string_config = pipeline_executor.build_pipeline(
+            mod_config, "<path to storage the build output>"
```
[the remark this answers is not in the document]
huajsj: build output are multiple files,  to hide the complexity of export/import here add a path to save multiple output into a disk directory, then import only need the path as parameter.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
jtuyls: On point 2, with latency in ML we typically mean the time it takes to process one unit of data, which for image classification is one image if the batch size is 1. In the example given, I would argue that the batch size is still 1 instead of 2 as the data inputs are just being queued and executed separately and not being processed together. Therefore, the latency is not reduced but the throughput is improved.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
comaniac: So it basically says pipeline executor can support general python functions in addition to DNN models. Please rephrase this part to make it clearer.

p.s. The syntax #<number> is referrring another PR or issue, so please do not use it if that's not your intention.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
comaniac: I don't think it's necessary...for example, you could have a wrapper like `PipelineIRModule` to put all built modules together.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
huajsj: about "not being processed together", actually that is not true,  for image processing in pipeline case, when image 1 released stage 1 and doing compute in stage 2 that time, image 2 will do processing in stage 1, in such scenario,  image 1 and image 2  being processed together. 

I agree one unit of data latency is the typically latency in ML, but from my point of view, in deployment case multiple data  latency are more meaningful for product performance evaluation.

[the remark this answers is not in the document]
comaniac: I actually don't think it's a typical case only in ML. This concept is also widely used in HPC. The "latency" is based on the time of executing one task; while the "throughput" is the number of finished task in a time unit. Pipelining definitely cannot improve latency (of executing one task) but throughput, because you can finish more tasks in the same time period.

General speaking, I think this is just the terminology issue, but it would be good to use consistent terminology for efficient communications. The confusion of "not being processed together" falls into the similar reason. Jorn and probably most people may feel confuse about why two stages of the same task can be processed together, but you actually meant stage 2 of the _first task_ and the stage 1 of the _second task_ can be processed in parallel. This is the point you should make it clear in the RFC.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
huajsj: this is good summary, instead of to say "pipeline executor can support general python function", I think "pipeline executor provide the capability to integrate non-DNN model function" should be more accurate, because current as I know tvm not support normal python function and pipeline executor not provide such feature too.

about #<number>, yes would avoid such syntax.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
jtuyls: With `not being processed together` I meant that both data inputs are not in the same stage at the same time, which is what is meant with the batch size. You could also execute in a pipelined manner with a batch size > 1, which would mean that you process multiple data inputs at every stage at the same time. So executing with a certain batch size is orthogonal to executing in a pipelined manner and latency refers to the time it takes to process a certain batch of data (typically size 1), irrespective of how you process it (pipelined or otherwise). It's indeed a terminology issue and I think it's best to use the common meaning to avoid confusion, so I think using `increase throughput` instead of `reduce latency` would be better here.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,365 @@
...
+    anf = run_opt_pass(anf, transform.InferType())
+    ann = _recursion(anf, operator_indx, pipeline_mods, subgraph_indices, constant_expr)
+    ann = run_opt_pass(ann.body, transform.ToGraphNormalForm())
+    mod = tvm.IRModule.from_expr(ann)
+    pipeline_mods.insert(0, mod)
+    return pipeline_mods
+
+#...
+mod, params = relay.frontend.from_darknet(net, dtype=dtype, shape=dshape)
+split = [11, 22]
```
jtuyls, line 209: I think using indices makes it hard to understand and use. Could we do something similar to the `compiler_begin`, `compiler_end` annotations in the BYOC flow instead?

rfcs/0012-pipeline-executor.md
huajsj: sure, about graph split part, following are the initial solution,  this solution include 3 steps, 1. Operator Auto tune,
2. Graph dependency tree build and balance, 3. Graph Auto Tune. following are more detail.

1. Operator Auto Tune : 
    1.1 In operator Auto tune  tune section, user would using existing tuning logic to tune the every operator,
          but the tune would separately and serialized happen in all target involved by pipeline executor.
    1.2 After operator tune done ,  we would get performance data, for example , con2d_0 best perf in 
         GPU is 3ms, in VTA is 2ms etc, this perf data would get used in later graph split logic.
2.  Graph dependency tree build balance 
     2.1 Initialize a DAG, the node of the DAG is subgraph, initially for a N node DAG, first [1, N -1] node mapping to
           [1 , N-1] layer(compute density operator and others) of original compute graph, the number N node is
           mapping to [N, M] layer , M here is the original compute layer number.
     2.2, by useing the perf data generated in 1.2 , every dependency tree node can get a time consume value,  
           we would do the balance work to adjust the node subgraph scope to make every node have close perf 
           value.
3.  Graph Auto Tune.
     2. can generate more than one subgraph split solution, in this step, Graph Auto Tune would try these multiple
     solution to get best configuration.

after 1. 2. 3. , here can get an automatic graph split configuration.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
huajsj: @jtuyls, @comaniac, agree, would change into the term "throughput" to avoid confusion.

rfcs/0012-pipeline-executor.md
huajsj: currently pipeline executor rely on manually "graph split" and "configuration", this part logic just be a example to demo how to manually generate the output and not be part of this RFC solution,  in further we would go through "auto graph split" feature to hide this complexity and make such graph split logic be transparent to user,  I would update the document to avoid confuse.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,367 @@
...
+mods = pipeline_graph(mod["main"], split)
+```
+
+### 3.2. Manually contstructin pipeline subgraph configuration for dependency and target device...
+There are dependency between pipeline subgraph, for example we have 3 pipeline subgraph named
+s1, s2, and s3, s2 input is s1 output and s2 output is s3 input, we need to construct a configuation
+file to descript such dependency relation, such configuratin also need to involved "target" and
+"device" information following is a example.
```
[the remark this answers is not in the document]
huajsj: about the configuration format, current solution should be ok to handle the said scenario
lets assume A index is 1, B is 2, c is 3
for  A - B
       |-- C
the configuration
[
        "mod_indx": 1
        "output": [
            {"output_indx": 1, "dependent": [{"mod_indx": 2, "input_name": "data_0"},
                                                                    {"mod_indx": 3, "input_name": "data_0"}]
           }
        ],
    }
]

for A - B - C
      |-------|
[
        "mod_indx": 1
        "output": [
            {"output_indx": 1, "dependent": [{"mod_indx": 2, "input_name": "data_0"},
                                                                   {"mod_indx": 3, "input_name": "data_0"}]
            }
        ],
        "mod_indx": 2
        "output": [
            {"output_indx": 1, "dependent": [{"mod_indx": 3, "input_name": "data_0"}]
            }
        ]
    }
]

about the complexity of manually configuration, sure, i will update this part to mention automatic graph split would handle this part.

rfcs/0012-pipeline-executor.md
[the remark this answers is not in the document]
huajsj: sure, would remove this disk save logic

huajsj:
@comaniac , @jtuyls , all review comments addressed, please review again.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+Pipeline Executor is a runtime executor which implements pipeline execution logic for multiple
+subgraphs and relies on graph_executor for operator storage and execution.
+
+This section introduces the use case for Pipeline Executor.
+
+* 1. Using Automatic Graph Split feature to construct pipeline subgraph and configuration.
+* 2. Use pipeline_executor to build a pipeline module with the subgraphs and configuration.
+* 3. Use pipeline_executor to load the pipeline module to run network in pipeline parallelism mode.
+
+### 3.1. Using Automatic Graph Split feature to construct pipeline subgraph and configuration.
```
comaniac, line 68: I'd suggest moving an entire 3.1 to the end of this RFC as the future work. You just need to say `This feature is out of scope of this RFC. One possible implementation is presented in Section 9.` here.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+
+### 3.1.3 Graph Auto Tune.
+* a. 3.1.2 can generate more than one subgraph split solution DAG, in this step, Graph Auto Tune would try these
+multiple solution to get best configuration.
+
+after 1. 2. 3. , here can get an automatic graph split configuration.
+
+### 3.2. Use pipeline_executor to build pipeline module with the said subgraph and configuration.
+
+Pipeline executor provide a build function to compile and save the compile output into disk,
```
comaniac, line 106: Still in the disk?

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+following is a example
+
+```python
+    with autotvm.get_pipeline_model_best(mod_file) as mod_config: # this is future feature not in this RFC
```
comaniac, line 110: Since this is not in this RFC, you cannot use the API in this example. You still have to show how users can manually specify the configuration, because this is your current user interface.

Accordingly, we have to design the configuration to make it user friendly. Ideally, we probably need a set of APIs to generate the config. Here is a rough example I can think of at this moment:

```python
mod1, mod2, mod3 = my_manual_partitioner(mod)
pipe_cfg = PipelineModuleConfig()

# Define pipeline inputs. Here I assume two inputs of mod1 and one input of mod3 are the pipeline inputs.
pipe_cfg.inputs["data_0"] = (mod1, "data_0")
pipe_cfg.inputs["data_1"] = (mod1, "data_1")
pipe_cfg.inputs["data_2"] = (mod3, "data_0")

# Define pipeline outputs to be the first output of mod3.
pipe_cfg.outputs.append((mod3, 0))

# Define connections.
pipe_cfg.connect(mod1, 0, mod2, "data_0") # mod1.output(0) -> mod2.data_0
pipe_cfg.connect(mod2, 0, mod3, "data_1") # mod2.output(0) -> mod3.data_1

# Print config for debugging
print(str(pipe_cfg))
# Inputs:
#   |- data_0: mod1.data_0
#   |- data_1: mod1.data_1
#   |- data_2: mod3.data_0
# Outputs:
#   |- mod3.output(0)
# Connections:
#   |- mod1.output(0) -> mod2.data_0
#   |- mod2.output(0) -> mod3.data_1

# Use the config to build a pipeline executor
with relay.build_config(opt_level=3):
    lib = pipeline_executor.build_pipeline(pipe_cfg)
```

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+
+following is one example
+
+```python
+#...
+
+def get_output(outputs, module):
+  suc = False
+  output = pipeline_module.get_output()
+  if len(output):
```
comaniac, line 141:
```suggestion
  if output:
```

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+    suc = True
+
+  return suc
+    
+
+pipeline_outputs = []
+datas = []
+for i in range(len(mods) + 1):
+  datas.append(np.full(dshape, 3 + i).astype("float32"))
+pipeline_module = pipeline_executor.create(lib)
```
comaniac, line 153: We should not let users write `len(mod)`. Instead, this information should be self-contained, so you could have
```python
pipeline_module = pipeline_executor.create(lib)
datas = [np.full(dshape, 3 + i).astype("float32") for _ in range(pipeline_module.num_inputs())]
```

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+
+for data in datas:
+    pipeline_module.set_input("data_0", data)
+    pipeline_module.set_input("data_1", data, mod_idx=2)
```
comaniac, line 157: `mod_idx` is confusing. Suggest to use a unified pipeline config as I commented above, so that pipeline_module could know where "data_1" should feed to.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,214 @@
...
+    pipeline_module.run()
+    get_output(pipeline_outputs, pipeline_module)
```
comaniac, line 159: 1. The API is better to be:
```python
pipeline_outputs.append(pipeline_module.get_outputs())
```

2. Better to have the following scenario (I continue using the example I used above.)
```python
# Prepare inputs.
datas = []
for _ in range(5):
    # Each data includes 3 tensors (i.e., data_0, data_1, data_2 for the pipeline).
    datas.append([np.full(shape[i], 0).astype("float32") for i in range(3)])

# Feed all available inputs.
for data in datas:
    pipeline_module.set_input("data_0", data[0])
    pipeline_module.set_input("data_1", data[1])
    pipeline_module.set_input("data_2", data[2])
    pipeline_module.run()

# Get all outputs.
while pipeline_module.has_next_output():
  pipeline_outputs.append(pipeline_module.get_output())
```

comaniac (changes requested):

rfcs/0012-pipeline-executor.md
huajsj: fixed.

rfcs/0012-pipeline-executor.md
huajsj: fixed.

rfcs/0012-pipeline-executor.md
huajsj: this solution looks very readable, I will apply this logic.

rfcs/0012-pipeline-executor.md
huajsj: fixed.

rfcs/0012-pipeline-executor.md
huajsj: fixed.

rfcs/0012-pipeline-executor.md
huajsj: fixed.

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,235 @@
...
+
+* Pipeline executor provide the capability to integrate non-DNN model function.
+
+## 3. Guide-level explanation
+Pipeline Executor is a runtime executor which implements pipeline execution logic for multiple
+subgraphs and relies on graph_executor for operator storage and execution.
+
+This section introduces the use case for Pipeline Executor.
+
+* 1. Manually Split relay module a list relay modules and generate modules configuration.
```
comaniac, line 64:
```suggestion
* 1. Manually split/partition Relay module to a list of Relay modules and generate modules configuration (automatic module splitting is out of scope of this RFC and will be a future work).
```

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,235 @@
...
+# Connections:
+#   |- mod1.output(0) -> mod2.data_0
+#   |- mod2.output(0) -> mod3.data_1
+
+
+```
+
+### 3.2. Use pipeline_executor to build pipeline module with the said subgraph and configuration.
+
+following is a build example
```
comaniac, line 104:
```suggestion
The interface is mostly the same as the graph executor but accepts a pipeline configuration instead of a Relay module. Here is an example.
```

rfcs/0012-pipeline-executor.md
```diff
@@ -0,0 +1,235 @@
...
+run: run with the input at the front.
+get_output
+set_input: queue the input in the buffer.
+run: run with the input at the front.
+get_output
+get_output
+
+`get_output` can be called anytime, and it will return an empty array if no output is ready.
+
+following is one example
```
comaniac, line 131:
```suggestion
Pipeline executor works asynchronously. Unlike the blocking `run` API in graph executor,
`run` API in pipeline executor is non-blocking. As a result, we could have the following scenario:

1. set_input(): Push the input to the queue.
2. run(): Launch a task with the first input in the queue.
3. set_input(): Push the second input to the queue.
4. set_input(): Push the third input to the queue.
5. run(): Launch a task with the second input.
6. get_output(): Get the output of the first input.
7. run(): Launch a task with the third input. 
8. get_output(): Get the output of the second input.
9. get_output(): Get the output of the third input.

As can be seen, `get_output()` can be called anytime to get the first available output in the result queue,
and it will return an empty array if no output is ready.

Following is one example:
```

comaniac (approved):
Overall LGTM. It would be great if other folks can also take a look; otherwise I plan to merge this RFC after 3 days.

cc @areusch @jroesch @zhiics @junrushao1994

comaniac:
btw, please change the RFC file name to align the PR number (0014).

huajsj:
> btw, please change the RFC file name to align the PR number (0014).

@comaniac, thanks for the follow up, the file name change already done.

comaniac:
Due to no objection, this RFC is now merged. Thanks @huajsj

HungYangChang:
Hello @huajsj  I am super interested in your work!
As I have asked in [discussion form](https://discuss.tvm.apache.org/t/can-tvm-split-work-into-different-layers-and-assign-layers-into-different-cores/11161?u=popojames), I am using Hikey 970 with 4 big and 4 little CPUs to do inference now. One of my interest is to split my BERT model into the different layers (sub-graph) and then assign to different cores. 

I will definitely try to replicate your work in these two days and see what can I get. Thanks for your great work!

PR #21, opened 2021-08-15
[Refactor] Meta Schedule RFC

junrushao1994:
Renumbering meta schedule RFC to 0005 according to #17.

Also updated the text a little bit according to Andrew's suggestions


tqchen:
cc @areusch

tqchen (approved):

PR #23, opened 2021-08-17
[RFC][TIR] Adding annotation field to tir.allocate

manupa-arm:
* adding the markdown

Change-Id: I6980d8f9a228ff5e8a79d74220db9b77f88a3e1b

### DIFF
--- a/rfcs/0023-adding-annotation-field-to-tir.allocate.md
+++ b/rfcs/0023-adding-annotation-field-to-tir.allocate.md
@@ -1,14 +1,14 @@
 
-- Feature Name: TIR Pinned Memory Representation
+- Feature Name: Adding annotatation field to tir.allocate nodes
 - Start Date: 2021-06-01
 - RFC PR: https://github.com/apache/tvm-rfcs/pull/23
 - GitHub Issue: TBD
 
 # 1. Summary
 
-This RFC proposes how pinned memories could be associated with tir.allocate nodes (and allocate_const nodes) and used by passes in the lowering process.
+This RFC proposes to annotation field tir.allocate nodes. These annotations can be used as auxiliary hint to future transformations.
 
-# 2. Motivation 
+# 2. Motivational usecase : pinned memory 
 
 Currently, TVM relies on dynamic (alloc and free style) allocations in runtime to manage the intermediary memory used by operators and the network. This is sometimes not desirable, especially in microTVM.
 
@@ -53,23 +53,34 @@ The current design of [Unified Static Memory Planner (USMP)](https://github.com/
 
 Therefore, we'd need a way to represent the association of each of these memories, that the user will pin the buffers to, closer to allocate nodes in TIR.
 
-# 3. Guide-level explanation
-
-This is not particularly a user-facing feature.
-
-However, the intention here is to associate the name of the memory (along with other information) closer to the allocate IR node. Therefore at the end of the compilation, metadata module (+ header) will generate pointer structs to-be passed in the from the application layer. 
-
- In the case where user does not wish to pass in workspace or constant buffers, metadata module will generate a workspace buffer and a constant buffer, as explained in U1 of [USMP](https://github.com/apache/tvm-rfcs/pull/9).
-
- # 4. Reference-level explanation
-
- At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
+At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
 
  There are broadly two requirements here :
 
 P1 : Indicate candidate memories (a.k.a. Pools) that each allocate be associated with
 
 P2 : Indicate the final memory the allocate will be pinned on
+
+
+To serve P2, we propose to use the existing tag of the storage_scope with 'global-<memory_name>'.
+
+```
+struct StorageScope {
+  /*! \brief The rank of the storage */
+  StorageRank rank{StorageRank::kGlobal};
+  /*! \brief tag for special purpose memory. */
+  std::string tag;
+```
+
+To serve P1, this RFC introduces the addition of the annotations field
+
+# 3. Guide-level explanation
+
+This is not particularly a user-facing feature.
+
+
+ # 4. Reference-level explanation
+
 
 To serve P1, we propose to use :
 
@@ -93,27 +104,8 @@ class AllocateNode : public StmtNode {
 + Map<String, Objectref> annotations;
 ```
 
-Here the addition of the annotations field could serve offer hints/guides to future passes (i.e. Unified Static Memory Planner). We could use "candidate_memory_pools" as the key while the value being Map<String, PoolInfo>.
+Here the addition of the annotations field could serve offer hints/guides to future passes (i.e. In Unified Static Memory Planner, we could use "candidate_memory_pools" as the key while the value being Map<String, PoolInfo>.)
 
-```
-struct PoolInfoNode : public Object {
-  String  pool_name;
-  Integer size_bytes;
-  Integer alignment;
-  Integer pool_offset;
-  Map<Target,String> target_access; // 'rw' or 'ro'
-}
-```
-
-To serve P2, we propose to use the existing tag of the storage_scope with 'global-<memory_name>'.
-
-```
-struct StorageScope {
-  /*! \brief The rank of the storage */
-  StorageRank rank{StorageRank::kGlobal};
-  /*! \brief tag for special purpose memory. */
-  std::string tag;
-```
 
 # Alternatives
 
@@ -196,7 +188,7 @@ However, in the discussion with community, we felt the need to seperate changes 
 
 # 5. Drawbacks
 
-It is not direct because we are using general annotations to indicate this, however it is consistent with rest of the IR and allows future expansions.
+None. Its consistent with rest of the IR design and allows other features to use to pass hints for future transformation in the compiler.
 
 
 
--- a/rfcs/0023-adding-annotation-field-to-tir.allocate.md
+++ b/rfcs/0023-adding-annotation-field-to-tir.allocate.md
@@ -95,9 +95,7 @@ class AllocateNode : public StmtNode {
    * \brief Additional annotations about the loop.
    *
    *  These annotations can be used as auxiliary hint
-   *  to future transformations. An annotation should
-   *  not change the control flow semantics of the loop
-   *  and can be ignored in most passes.
+   *  to future transformations.
    */
 + Map<String, Objectref> annotations;
 ```
--- a/rfcs/0023-adding-annotation-field-to-tir.allocate.md
+++ b/rfcs/0023-adding-annotation-field-to-tir.allocate.md
@@ -91,9 +91,13 @@ class AllocateNode : public StmtNode {
   PrimExpr condition;
   /*! \brief The body to be executed. */
   Stmt body;
-  /*! \brief If the allocate is scoped global, this field indicates
-   *  which external memories it could be pinned to as a comma seperated
-   *  string.
+  /*!
+   * \brief Additional annotations about the loop.
+   *
+   *  These annotations can be used as auxiliary hint
+   *  to future transformations. An annotation should
+   *  not change the control flow semantics of the loop
+   *  and can be ignored in most passes.
    */
 + Map<String, Objectref> annotations;
 ```
--- a/rfcs/0023-adding-annotation-field-to-tir.allocate.md
+++ b/rfcs/0023-adding-annotation-field-to-tir.allocate.md
@@ -51,32 +51,26 @@ The current design of [Unified Static Memory Planner (USMP)](https://github.com/
     }
 ```
 
-Therefore, we'd need a way to represent the association of each of these memories, that the user will pin the buffers to, closer to allocate nodes in TIR.
+Therefore, we'd need a way to represent the association of each of these memories, that the user will pin the buffers (e.g. workspace_buffer_2 to SRAM) to. That infomation should flow closer to allocate nodes in TIR in the compilation passes.
 
-At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
+At the IR, we ll need to associate each allocate node with one (or more) memory pools that it can end up, because the scheduling might be satisfied with placing buffers in any of the memory pools in a given set of memory pools. Therefore, the scheduler might want the memory planners to decide which memory pool to use based on finding the allocation that fit.
 
  There are broadly two requirements here :
 
-P1 : Indicate candidate memories (a.k.a. Pools) that each allocate be associated with
+P1 : Indicate candidate memory pools (a.k.a. PoolInfo Objects -- for further details see [USMP RFC](https://github.com/apache/tvm-rfcs/pull/9)) that each allocate be associated with
 
-P2 : Indicate the final memory the allocate will be pinned on
+P2 : Indicate the final memory pool the allocate will be pinned on
 
 
-To serve P2, we propose to use the existing tag of the storage_scope with 'global-<memory_name>'.
-
-```
-struct StorageScope {
-  /*! \brief The rank of the storage */
-  StorageRank rank{StorageRank::kGlobal};
-  /*! \brief tag for special purpose memory. */
-  std::string tag;
-```
+To serve P2, we are proposing to use the existing tag of the storage_scope with 'global.<pool_name>' in the [USMP RFC](https://github.com/apache/tvm-rfcs/pull/9).
 
 To serve P1, this RFC introduces the addition of the annotations field
 
 # 3. Guide-level explanation
 
 This is not particularly a user-facing feature.
+
+The proposal in this RFC is to add an annotation field to tir.allocate node to be used by compilation passes.
 
 
  # 4. Reference-level explanation
--- a/rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
+++ b/rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
@@ -1,7 +1,7 @@
 
 - Feature Name: TIR Pinned Memory Representation
 - Start Date: 2021-06-01
-- RFC PR: TBD
+- RFC PR: https://github.com/apache/tvm-rfcs/pull/23
 - GitHub Issue: TBD
 
 # 1. Summary
--- a/rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
+++ b/rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
@@ -65,6 +65,58 @@ However, the intention here is to associate the name of the memory (along with o
 
  At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
 
+ There are broadly two requirements here :
+
+P1 : Indicate candidate memories (a.k.a. Pools) that each allocate be associated with
+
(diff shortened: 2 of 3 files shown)


manupa-arm:
cc : @tqchen @junrushao1994 @jroesch @areusch

tqchen:
Thank you @manupa-arm for the proposal. 

What you described is closest to the "tag" field of the storage scope, which we put as part of the pointer and associate with most of the storage. I think it would be great to think in terms of this direction.

Specifically, when we have a clear distinction of memory types, "sram", "shared_memory", "texture". We will use the special memory tag to indicate the type of such kind of memory.

The proposed pinned memory field, however, have multiple of such candidates. In my understanding tries to serve as an additional hints to follow up passes. If the intended use is multiple candidate hints, then the tag is perhaps no longer approperiate because it is used to indicate one kind of special memory that we place the data to.

In that case, I think it could be closer to an additional annotation field https://github.com/apache/tvm/blob/main/include/tvm/tir/stmt.h#L817 that stores additional attribute hints about the particular node. This would also help reduce the possible indirection introduced by AttrStmt.

tqchen:
cc @masahi @kparzysz-quic @csullivan

junrushao1994:
Thanks @manupa-arm for the nice proposal!

Currently the memory scope of a buffer not only indicates where the buffer is allocated, but also affects schedule primitives. For example, “shared” and “local” may lead to different TIRs after lowering. Therefore, it would be good to consolidate scopes that don’t affect scheduling behaviors with additional annotations fields, as @tqchen suggested. Thanks a lot 🙏

rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
```diff
@@ -0,0 +1,158 @@
...
+
+This is not particularly a user-facing feature.
+
+However, the intention here is to associate the name of the memory (along with other information) closer to the allocate IR node. Therefore at the end of the compilation, metadata module (+ header) will generate pointer structs to-be passed in the from the application layer. 
+
+ In the case where user does not wish to pass in workspace or constant buffers, metadata module will generate a workspace buffer and a constant buffer, as explained in U1 of [USMP](https://github.com/apache/tvm-rfcs/pull/9).
+
+ # 4. Reference-level explanation
+
+ At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
```
csullivan, line 66: Can you provide more context with this statement specifically:

>we'll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories.

It may be relevant to some related work we are doing.

csullivan (commented):
Thanks for the RFC, it's good to discuss this. I'm in agreement with what has been said wrt the memory scope, and using the storage_scope field (including tag) of the buffer's pointertype simplifies your S3 to using only needing this field if I understand your needs correctly. Can I ask what is meant by the below and when you might use this?

junrushao1994:
To clarify a bit, @tqchen and I are referring to a solution that we are still using "global" as the storage scope, which is served as the property of the storage, but add extra annotations (e.g. "sram") to further indicate the exact place where the buffer is stored. This way, the system is able to schedule correctly

manupa-arm:
Hi All,

Thanks for taking the time to look at it.
Initially, my thoughts were to use the same field for two purposes : 
P1) Indicate candidate memories (a.k.a. pools) that each allocate be associated with
P2) After the memory planner is run, it will pick one out of the list by reducing the candidate to exactly 1.

However, having read the comments here, I feel these could be two different fields.

>The proposed pinned memory field, however, have multiple of such candidates. In my understanding they can be served as aadditional hints to follow up passes. If the intended use is multiple candidate hints, then the tag is perhaps no longer approperiate because it is used to indicate one kind of special memory that we place the data to.

For P1, I think we could use this annotation field to initially describe this.
Something like (as proposed by @MichaelJKlaiberBosch (here)[https://github.com/apache/tvm-rfcs/pull/9/files#r671371629] )  : 

T1) :
```
Map<String, PoolInfo> candidate_memory_pools;
```
In this way, we dont need to carry this as an IRModule attribute to be de-referenced.
Or should we just add a very generic 

T2) :
```
Map<String, Objectref> annotations;
```
when the key is "candidate_memory_pools" would be the value is Map<String, PoolInfo> candidate_memory_pools.
I would personally prefer T1 for its direct approach here but T2 could also be made to work.

>What you described is closest to the "tag" field of the storage scope, which we put as part of the pointer and associate with most of the storage. I think it would be great to think in terms of this direction.

> Specifically, when we have a clear distinction of memory types, "sram", "shared_memory", "texture". We will use the special memory tag to indicate the type of such kind of memory.

Thereafter once the memory planner is run, it could read the annotation field and populate a final 'global.<memory_pool>'.
I think this should address the all of your concerns. Let me know, if agreed I'll adjust the RFC text.

tqchen:
Thanks @manupa-arm I would suggest T2 to make things consistent with the rest of IR design and allow future annotations if needed. 

Note that annotation fields only offers hints to future passes(on how to rewrite) and should not change the semantics of the IR

rfcs/0023-associating-allocate-nodes-with-pinned-memories.md
manupa-arm: Hi @csullivan,

Let's say there are multiple physical memories that have same bandwidth and latency and schedulers are just fine placing the buffer in each of them as long as all things fit. In the scenario, the scheduler would want keep more than one candidate for the memory planner to decide.

However, based on the comments, I feel we are going to use two fields for these two purposes. See my latest response below.

manupa-arm:
Thanks @tqchen , yes we could work this for now.

I take it that for P2, we could use the tag of storage_scope as well, then ?

tqchen:
Thanks @manupa-arm , P2 falls into the domain of the special memory tag, and the storage scope is exactly designed for this purpose.

manupa-arm:
Cool; I ll adjust the RFC text tommorow.

manupa-arm:
Hi all,

Sorry for delay!
I've managed to update the text to reflect the discussion. 

PTAL and if its good, shall we get this in ?

manupa-arm:
@tqchen, can you take a look?

I can modify the USMP RFC to reflect the decisions made here, if we could finalize the discussion here.

tqchen:
Thanks @manupa-arm . I think the proposal(of adding annotations) looks good. Perhaps we can change the title to

adding annotation fields to the Allocation, and discuss the need of pinned memory as a usecase. In this way it can be viewed as a IR change proposal.

manupa-arm:
Done

junrushao1994:
Thanks for the really helpful discussion! The design choice looks good to me :-)

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+             .flash = &my_model_params_1,
+             .itcm = &my_model_params_2
+         };
+         TVMSetWorkspaces(&context, &workspaces);
+         TVMSetParameters(&context, parameters);
+         TVMExecute(&my_model, &inputs, &outputs, &context);
+    }
+```
+
+Therefore, we'd need a way to represent the association of each of these memories, that the user will pin the buffers to, closer to allocate nodes in TIR.
```
areusch, line 54: the user could pin buffers or USMP could pin buffers. does this address both needs?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+
+At the IR, we ll need to associate each allocate node with one (or more) memories that it can end up, because the scheduling might be satisfied with placing buffers in any of the memories in a given set of memories. Therefore, the scheduler might want the memory planners to decide which memory to use based on finding the allocation that fit.
+
+ There are broadly two requirements here :
```
areusch, line 58: can this go in Guide-level explanation with some examples?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+  StorageRank rank{StorageRank::kGlobal};
+  /*! \brief tag for special purpose memory. */
+  std::string tag;
+```
+
+To serve P1, this RFC introduces the addition of the annotations field
+
+# 3. Guide-level explanation
+
+This is not particularly a user-facing feature.
```
areusch, line 79: i'd say if the user is pinning tir.allocate it could be.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+  Array<PrimExpr> extents;
+  /*! \brief Only allocate buffer when condition is satisfied. */
+  PrimExpr condition;
+  /*! \brief The body to be executed. */
+  Stmt body;
+  /*! \brief If the allocate is scoped global, this field indicates
+   *  which external memories it could be pinned to as a comma seperated
+   *  string.
+   */
++ Map<String, Objectref> annotations;
```
areusch, line 104: I'm not sure I love having both the attr mechanism and this arbitrary map<str, value>. Can you say why we should use ObjectRef as the value here?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+TIR:
+
+```
+allocate_node_1 = tir.allocate([157323], "int16", "global.(foo_memory,bar_memory)")
+```
+
+
+Out of the options, S1 seems the most non-invasive. However, we will need special handlers to obtain the information.
```
areusch, line 181: however it does seem like this is a common problem with Attr and we could write a common one.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+
+S2 does not change the storage scope (or the 'tags') and adds an additional field to allocates to note this information.
+
+S3 fold the information into storage_scope and utilizes the 'tag' denote the memory. The change to IR, is just to support more tags.
+
+However, in the discussion with community, we felt the need to seperate changes that serves P1 and P2, respectively. In fact, for P2 we could re-use the tag of storage_scope without an IR change. For P1, it seems a good and general change to include annotations for the allocate node.
+
+# 5. Drawbacks
+
+None. Its consistent with rest of the IR design and allows other features to use to pass hints for future transformation in the compiler.
```
areusch, line 191: Adding a field to `tir.allocate` is a bigger code lift. can we think of some other drawbacks?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,201 @@
...
+ There are broadly two requirements here :
+
+P1 : Indicate candidate memories (a.k.a. Pools) that each allocate be associated with
+
+P2 : Indicate the final memory the allocate will be pinned on
+
+
+To serve P2, we propose to use the existing tag of the storage_scope with 'global-<memory_name>'.
+
+```
```
areusch, line 67: could you add some explanation of this struct? it's just here with no explanation of its purpose. also it looks cut-off.

areusch (commented):
thanks @manupa-arm i added some feedback

areusch:
cc @mbs-octoml can you take a look?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I think both cases are 'users' of this addition to the IR in a very similiar way. 
However, I dont think users have an API directly to pin tir.allocates.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I dont think it should go to the guide-level explanation of this RFC because this is just motivating the addition to the IR.
Instead, I ll add a reference to USMP RFC to see for further usecases of this IR change.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I'll just remove it as tags of storage_scope is an established concept and not relevant to this RFC now.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I dont think we have a user API to do that.

This IR change is to be used by compilation passes to pass on hints to future optimizations.
Therefore, the core compiler could use them to pass on the PoolInfo to tir.allocates in the passes. 

I would not consider passes as users, in this scenario.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I think we are being consistent with other TIR nodes and this (and others) use ObjectRef as a general way to pass compilation hints from IR point of view while the passes could specialize that. I personally agreed with @tqchen on the basis that this avenue will create the lesser changes in future for the IR, rather than doing a specialized addition of PoolInfo here.

The question of having both ways of annotations, seems to be out of scope for this RFC as we are not introducing another way here.

(if you follow @tqchen's citation of ForNode)

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I dont disagree but I feel that is out of scope for this RFC.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: I dont think adding an annotation field (similiar to ForNode) to pass compilation hints for passes is bigger change compared to other TIR nodes we already have today.

manupa-arm (commented):
Thanks @areusch for taking a look.

I've anwsered your questions and adjusted the text accordingly.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
manupa-arm: Oh, in this sentence, it means the buffer mentioned in the sample application above.

USMP will not do that. USMP will change program to expect the application/runtime will pass in a buffer of the correct size pinned to correct memory.

areusch: do you think that users should have such an API? I kind of do (but it would be an advanced use case and e.g. done via Python script perhaps).

manupa-arm: It needs more thinking around the usecases. Shall we discuss that as a follow up RFC?

mbs-octoml (commented):
I think we're approaching the same problem from opposite ends. For device planning we wanted to avoid the Map<Expr, Device> convention for recording the association of a device to every Relay expression since it does not survive transformation.
We considered:
 - adding a DictAttrs to the base Expr node, and auditing the entire codebase for preserving attributes on copies.
 - adding a new WithAttrs Relay node to capture annotations which are implicitly scoped over the sub-expression.
 - using 'fake' "on_device" calls to record device associations for a sub-expression.
 - using attributes on the existing Function definitions, which are already copied correctly.
I've pushed on a combination of the last two to see where it takes us. 

There is a general desire to extend device planning to be memory scope aware, which is where the above and this RFC make contact. In particular, would you ever want the set of candidate scopes to have been provided by some extended device planning pass in Relay? Or are you imagining that the candidate scopes are attached directly during lowering and are separate from the current device planning machinery?

If we want these worlds to work together it's worth adding DictAttrs further up (maybe even Object???) and make sure the codebase preserves attributes by default in all transforms.

tqchen:
There is generally a tradeoff between the:
- A0: The desire to preserve information during transformation.
- A1: A growing set of attributes that is impossible to keep track of(thus preserve) during transformations.

From the A1's pov, allowing set of attributes that changes the semantics of the IR is in general not desirable. Since other pass writers may not be aware of all the possible set of attributes and thus being mindful in preserving them. 

The middle ground that we could reach here is to allow annotations, which only served as hints for future passes(that the code can be transformed in certain way). This means that the other pass writers can safely ignore the content of annotation itself. Note that this would certainly restrict the possible set of annotations that we can use(e.g. the object themselves should not refer to an active piece in the IR).

rfcs/0023-adding-annotation-field-to-tir.allocate.md
areusch: but couldn't a user in principle use such a thing through TVMscript?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
areusch: Ah okay nevermind, I traced through more things and I agree with ObjectRef as a value.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
areusch: Let's discuss this specific use in USMP RFC then. I do kind of agree that it's not in scope for an RFC that concerns itself only with adding the generic annotation/attr lookup to tir.allocate.

rfcs/0023-adding-annotation-field-to-tir.allocate.md
areusch: well, what about the concern about creating multiple ways to do the same thing?

areusch (commented):
@tqchen however doesn't keeping stuff just as an AttrStmtNode serve the same purpose? Why should we add this additional map-to-arbitrary-thing to the IR? It seems like a shortcut. And why should AttrStmtNode limit its values to PrimExpr but annotations not do so? finally, are there concerns about roundtripping any such metadata through tvmscript? cc @junrushao1994

I kind of agree we should consider whether DictAttrs and AttrStmt could work for this.

manupa-arm (commented):
Im not sure whether we want to roundtrip annotations that are only put via compilation passes. I'd rather have them restricted to be inserted by a controlled compilation pipeline, at least initially.

We have both ways to attach attributes/annotations to an IRNode in the codebase already (AttrStmt is restricted to be PrimExpr for now -- therefore users could attach simple attributes/pragmas -- I think, however it would not work for our current need). It is an interesting discussion to have IF (not sure we really want to) we want to stick to one, but I think that work should be done holistically and once for all IRNodes (e.g. ForNode).  My reasons currently are that annotations are bit private to the compiler and could only be inserted by passes. 

I feel annotations are simpler and straightforward while AttrStmtNode needs to handle in separate visits prior to visiting the actual StmtNode while it could have been simpler to attach them directly to the StmtNode of interest. I guess one point could be the AttrStmtNode survives by default that mutates the StmtNode, however that is a roundabout way where the mutations should've copied the Attrs/Annotations by default. 

I would prefer the annotations over AttrStmt due to the somewhat complex handling of it in all the passes that care about it.

tqchen:
@areusch yes about AttrStmt. I had to admit that AttrStmt has now becomes a legacy issue that I hope we can gradually move away from(so those that changes semantics can be directly reflected in the IR and properly documented). While we cannot immediately remove them, we can gradually move the features out and eventually deprecate

manupa-arm:
@areusch @tqchen , 

Could we agree to move on with using the annotations instead of AttrStmt?

manupa-arm:
@tqchen merge?

rfcs/0023-adding-annotation-field-to-tir.allocate.md
```diff
@@ -0,0 +1,195 @@
...
+  /*! \brief The type of the buffer. */
+  DataType dtype;
+  /*! \brief The extents of the buffer. */
+  Array<PrimExpr> extents;
+  /*! \brief Only allocate buffer when condition is satisfied. */
+  PrimExpr condition;
+  /*! \brief The body to be executed. */
+  Stmt body;
+  /*! \brief If the allocate is scoped global, this field indicates
+   *  which external memories it could be pinned to as a comma seperated
```
tqchen, line 95: NOTE: the comments here shall be updated that is independent from the pinned memory. as in the ForNode.annotations

tqchen:
I added one last minor nits. @mbs-octoml @csullivan @areusch  please take another look and let us merge in 24 h

manupa-arm:
@tqchen thanks for spotting that! did the change.

mbs-octoml:
No objections from me.

manupa-arm:
@tqchen shall we get this in then?

tqchen:
Thanks @manupa-arm @mbs-octoml @csullivan @areusch this RFC is now merged!

PR #24, opened 2021-08-20
[RFC][Frontend] Add a OneFlow Frontend

hhhfccz:
Add a OneFlow frontend for TVM, also I have created a RFC topic in forum where we can discuss at same time.
The related pull request is https://github.com/apache/tvm/pull/8790 .

@tqchen @comaniac  @junrushao1994

### DIFF
--- a/rfcs/0024-add-oneflow-frontend.md
+++ b/rfcs/0024-add-oneflow-frontend.md
@@ -1,7 +1,7 @@
 - Feature Name: (`add oneflow frontend`)
 - Start Date: (2021-8-20)
-- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/)
-- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/)
+- RFC PR: [apache/tvm-rfcs#0024](https://github.com/apache/tvm-rfcs/pull/24)
+- GitHub Issue: [apache/tvm#8804](https://github.com/apache/tvm/issues/8804)
 
 # Summary
 [summary]: #summary
--- a/rfcs/0024-add-oneflow-frontend.md
+++ b/rfcs/0024-add-oneflow-frontend.md
@@ -12,28 +12,19 @@ we have created a frontend for TVM that targets [oneflow](https://github.com/One
 # Motivation
 [motivation]: #motivation
 
-# TODO
+OneFlow, an open source deep learning framework with whole new frame design and the world's leading technology for distributed system. Here are advantages of OneFlow:
 
-# Guide-level explanation
-[guide-level-explanation]: #guide-level-explanation
+- Perfectly support container platforms(k8s & docker)
+- Handle large models easily
+- Almost zero runtime overhead & linear speedup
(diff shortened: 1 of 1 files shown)


rfcs/0024-add-oneflow-frontend.md
```diff
@@ -0,0 +1,133 @@
...
+
+```python
+import tvm
+import tvm.relay as relay
+
+
+# load eager model(assuming that resnet50 has been built)
+res50_module = resnet50()
+pretrain_models = flow.load(model_path)
+res50_module.load_state_dict(pretrain_models)
```
leandron, line 78: Does OneFlow support any input model from a file? If so, I'd like also to suggest that you also cover `tvmc` on your plans/implementation at some point.

Example implementation - https://github.com/apache/tvm/blob/e691c7f83892d7242d0992c78cec1e2f8953a9e3/python/tvm/driver/tvmc/frontends.py#L174-L198

hhhfccz: Thanks for your suggestions! We have added this task to Future possibilities.

BBuf: Thank you for your suggestions! We can't support this feature yet, but it is in our future development plan. After OneFlow supports this feature, we will add it to `tvmc` .

rfcs/0024-add-oneflow-frontend.md
```diff
@@ -0,0 +1,134 @@
...
+# Motivation
+[motivation]: #motivation
+
+OneFlow, an open source deep learning framework with whole new frame design and the world's leading technology for distributed system. Here are advantages of OneFlow:
+
+- Perfectly support container platforms(k8s & docker)
+- Handle large models easily
+- Almost zero runtime overhead & linear speedup
+- Support automatic mixed precision
+- ...
```
hogepodge, line 21: Is this just a placeholder? We should fill out the rest of the motivations or remove the ellipses.

hogepodge (commented):
Perhaps this isn't an issue that is covered in the scope of this RFC, but I'm wondering how independent the importers are from one another, and if it's worthwhile to look at a more general framework for third-party importers.

Are we aiming to support all operators? What does the mapping from OneFlow operators to TVM look like? Will there be gaps that a user will have to be concerned with? What is the plain for maintenance of the new format to import?

rfcs/0024-add-oneflow-frontend.md
BBuf: Yes，I will delete it。

BBuf: > 
> 
> Perhaps this isn't an issue that is covered in the scope of this RFC, but I'm wondering how independent the importers are from one another, and if it's worthwhile to look at a more general framework for third-party importers.
> 
> Are we aiming to support all operators? What does the mapping from OneFlow operators to TVM look like? Will there be gaps that a user will have to be concerned with? What is the plain for maintenance of the new format to import?

Yes, we will support all commonly used operators. OneFlow will release version 1.0 soon, and our operator interface is aligned with the Pytorch API.

PR #28, opened 2021-09-07
Command Line Composition from Internal Registry

Mousius:


leandron (approved):
I think this will greatly improve the usability and way we can offer documentation for `tvmc` users, and make it less scattered removing the need for multiple things encoded in the target string.

It would be to wait a bit for opinions from others, cc @areusch @gromero @comaniac

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,108 @@
...
+        metadata.Set("type", kv.second.type_key);
+        configs.Set(kv.first, metadata);
+    }
+    return configs;
+}
+```
+
+This can be replicated to provide the same information for `TargetKind` or any other registry and provide this information to generate arguments in `tvmc` using `argparse`:
+```
+parser.add_argument(f"--target-{kind}-{attr}", ...)
```
comaniac, line 71: So the TVMC argument becomes like the following?
```
tvmc --target=xx,c
--target-xx-attr
--target-cpu-mcpu
--target-cpu-device
--target-cpu-mtriple
--target-cpu-mattr
--target-cpu-mfloat-abi
```

Seems a bit tedious. How about the following:
```
tvmc --target=xx,c
--target-xx "-attr=..."
--target-cpu "-mcpu=x, -device=x, -mtriple=x, -mattr=x, -mfloat-abi=x"
```

We can still leverage the same idea as list configs to show the detail attributes of each target in this approach. You just need to embed the extracted information to the argparse description.

Mousius: There's two main reasons for using the more verbose form:
1. There's no need for a user to learn a sub-command syntax as they do now, they just read the help and add arguments as they need. Typing typically takes less time than learning a new syntax and remembering to switch between them. If we maintain them embedded in a bespoke string I don't think we gain much over the current `Target` string in terms of understandability?
2. Ideally, in combination with [Command Line Configuration Files](https://github.com/apache/tvm-rfcs/pull/30) if you have a complex set of arguments you should be saving it in a configuration file and using it to populate defaults (or in future you may be able to do this as environment variables as @hogepodge  suggested 😸 ):
```
tvmc --config=./dev_board.json --target-c-mcpu=cortex-m55
```

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, when a user uses `tvmc`, they present a target string:
+```
+tvmc --target="woofles -mcpu=woof, c -mattr=+mwoof"
+```
+
+Using a target string here means that the entire `--target` argument is used as an opaque pass-through to the internal `Target` string parser. It'd be great for a user to be able to compose these options and get meaningful help when building them up in `tvmc`.
```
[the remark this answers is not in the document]
Mousius: If we register these all with `argparse`, as an example:
```python
import argparse
parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
parser.add_argument('--target', type=str, help='comma separated target list or target string')
parser.add_argument('--target-c-mcpu', type=str, help='target c mcpu string')
parser.add_argument('--target-c-mattr', type=str, help='target c mattr string')
parser.add_argument('--target-llvm-mcpu', type=str, help='target llvm mcpu string')
parser.add_argument('--values-llvm-mattr', type=str, help='target lvm mattr string')
args = parser.parse_args()
print(args)
```

Using the default `--help` you get a lot more information for the `Target` options than with the opaque string passing:
```shell
$ python3 test.py --help
usage: test.py [-h] [--target TARGET] [--target-c-mcpu TARGET_C_MCPU] [--target-c-mattr TARGET_C_MATTR] [--target-llvm-mcpu TARGET_LLVM_MCPU]
               [--values-llvm-mattr VALUES_LLVM_MATTR]

optional arguments:
  -h, --help            show this help message and exit
  --target TARGET       comma separated target list or target string (default: None)
  --target-c-mcpu TARGET_C_MCPU
                        target c mcpu string (default: None)
  --target-c-mattr TARGET_C_MATTR
                        target c mattr string (default: None)
  --target-llvm-mcpu TARGET_LLVM_MCPU
                        target llvm mcpu string (default: None)
  --values-llvm-mattr VALUES_LLVM_MATTR
                        target lvm mattr string (default: None)
```

As a user, I can now read all the options and reason about them without having to venture into `target_kind.cc`. 

Is this the kind of example you were looking for?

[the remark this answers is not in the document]
comaniac: Yeah. It would be better to have an example like this in the RFC so that it would be easy for people to get your expectation.

rfcs/0028-command-line-registry-composition.md
comaniac: I agree that the first point is more user friendly; while my thinking was more developer friendly. I would not against this RFC if others feel it's fine to have verbose form in the CLI. One suggestion might be making this verbose form hierarchical, such as sub-commands of `--target`; otherwise it may scare people when they see tons of "helps" popping out with just `tvmc -h`.

rfcs/0028-command-line-registry-composition.md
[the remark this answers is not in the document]
Mousius: Cool, added, thanks for your feedback  :smile_cat:

rfcs/0028-command-line-registry-composition.md
Mousius: I've added groupings to the example, I think you're right if they're all just blasted onto the screen it's a bit scary but in groups it's similar to other CLIs where you just scroll to the section you're interested in. We can also iterate on help as we go if we think it's getting overwhelming.

Re: user vs developer, I've always assumed `tvmc` is a user facing tool so I started heading in that direction, @leandron is probably the right person to make the call though.

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, when a user uses `tvmc`, they present a target string:
+```
+tvmc --target="woofles -mcpu=woof, c -mattr=+mwoof"
+```
+
+Using a target string here means that the entire `--target` argument is used as an opaque pass-through to the internal `Target` string parser. Users coming to TVM should be able to compose these options and get meaningful help when doing that composition with `tvmc`. As an example, using the default `argparse` behaviour, the arguments can be registered as follows:
```
areusch, line 19: let us motivate composition of Target a bit further. what use case are we trying to support? what user burden are we trying to alleviate?

areusch: move the argparse code example to Guide-level explanation. i think the example `--help` output could stay as motivation.

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+print(args)
+```
+
+The user can get help for all available target options:
+```
+usage: test.py [-h] [--target TARGET] [--target-c-mcpu TARGET_C_MCPU] [--target-c-mattr TARGET_C_MATTR] [--target-llvm-mcpu TARGET_LLVM_MCPU] [--target-llvm-mattr TARGET_LLVM_MATTR]
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --target TARGET       comma separated target list or target string (default: None)
```
areusch, line 43: it's a little confusing to have both. how do they compose?

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+
+target c:
+  --target-c-mcpu TARGET_C_MCPU
+                        target c mcpu string (default: None)
```
areusch, line 47: it seems as though if we are going to do this, we need to supply descriptions for all of these. we should address the question of where those descriptions come from in this PR.

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+This would be translated at the `tvmc` level to:
+```bash
+tvmc --target=c \
+    --target-c-mcpu=cortex-m3
+```
+
+Which would then allow the user to compose together these options, such as:
+
+```bash
+tvmc --target=cmsisnn,c \ # Specifying multiple targets to enable in priority order
```
areusch, line 84: what happens if i accidentally include `--target-llvm-mcpu` in this target string?

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+target c:
+  --target-c-mcpu TARGET_C_MCPU
+                        target c mcpu string (default: None)
+  --target-c-mattr TARGET_C_MATTR
+                        target c mattr string (default: None)
+
+target llvm:
+  --target-llvm-mcpu TARGET_LLVM_MCPU
+                        target llvm mcpu string (default: None)
+  --target-llvm-mattr TARGET_LLVM_MATTR
```
areusch, line 54: this one should likely compose if you specify it twice (e.g. `-target-llvm-mattr +fp-armv8` on the command line and `-target-llvm-mattr +neon` in a config file should prooobably compose? If not `mattr`, i'm sure there are target flags which should. how do we differentiate these in parsing?

rfcs/0028-command-line-registry-composition.md
```diff
@@ -0,0 +1,153 @@
...
+[prior-art]: #prior-art
+
+This is not dissimilar to how `gcc` or `clang` provide options, [for example in gcc](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html):
+```
+gcc -fsanitize=address -fsanitize-address-use-after-scope
+```
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
```
areusch, line 138: i think there are some unresolved questions in my mind around handling of heterogeneous targets (e.g. with >2 target), and around some efforts i think @junrushao1994 is familiar with around representing and parsing targets from JSON.

areusch (changes requested):

areusch:
thanks @Mousius for raising this! a couple questions for you

rfcs/0028-command-line-registry-composition.md
Mousius: I'd happily get rid of the `Target` string, I left it in for backwards compatibility - it'd compose just the same as any other argument here, attaching additional parameters to the `Target`s which are built from the `--target` option. In a lot of ways keeping the `Target` string makes implementation easier as we just parse `--target` and then attach additional attributes.

rfcs/0028-command-line-registry-composition.md
Mousius: Given the time it'll take to land this in TVM, I'd rather not make more alternations to registries in this RFC, favouring small focused RFCs which deliver incremental value. I would suggest the concatenated descriptions shown should suffice for an initial implementation?

rfcs/0028-command-line-registry-composition.md
Mousius: Wrote a small blob explaining that'll error.

rfcs/0028-command-line-registry-composition.md
Mousius: This isn't covered here, this simply maps attributes to command line arguments. We can expand upon the behaviours once this lands.

rfcs/0028-command-line-registry-composition.md
Mousius: Is there an action here @areusch ?

rfcs/0028-command-line-registry-composition.md
Mousius: > let us motivate composition of Target a bit further. what use case are we trying to support? what user burden are we trying to alleviate?

I've tried to clarify this with some additional text, I've also added a series of disclaimers in the RFC that this shouldn't just apply to `Target` - a core tenet of good UX is a series of well understood patterns that the user can re-apply through-out the use of a product.

> move the argparse code example to Guide-level explanation. i think the example `--help` output could stay as motivation.

I've duplicated it a bit to ensure it makes sense in the Guide-level explanation.

Mousius:
Thanks for the feedback @areusch, I've replied to each of your comments except the last one around heterogeneous targets. Please take another look :smile_cat:

rfcs/0028-command-line-registry-composition.md
leandron: > I agree that the first point is more user friendly; while my thinking was more developer friendly. I would not against this RFC if others feel it's fine to have verbose form in the CLI. One suggestion might be making this verbose form hierarchical, such as sub-commands of `--target`; otherwise it may scare people when they see tons of "helps" popping out with just `tvmc -h`.

The thinking here is that the targets encode a lot of information which is hard to explain and expose to both users (not interested in TVM internals) and developers (interested in TVM internals). This will make it more self-explanatory for each argument that can be user from each target.

At the moment, you need a lot of internal knowledge about TVM source codes to understand which options are accepted for the targets.

In that sense, I think other tools like `clang` are a good example, in which despite the number of arguments potentially becoming large (as you raise with the output of `tvmc -h`), at least the existing arguments are structured and visible.

So IMHO this change is a very important step in making all interfaces that deal with target setting more transparent.

Mousius:
Hi @areusch, I think I've addressed the changes you requested, could you take another look?

rfcs/0028-command-line-registry-composition.md
areusch: sorry--was ccing @junrushao1994 as i couldn't remember where this was. there is currently `target._load_config_dict` as a way to parse Target from a JSON object. could we specify a way to work with this (or promote it to be the value of the `"target"` key here)?

areusch: the unresolved questions around heterogeneous targets essentially are: i suppose we will modify the `load_cnfig_dict` to work with [`TargetDevice`](https://github.com/apache/tvm/pull/8892) (e.g. we will assign a name to targets and likely the way that would make sense in `load_config_dict` is to make the target dict a sort of map from `targetdevice_name` to roughly the current dict expected by `load_config_dict`. i'm raising this here only because this is user-facing and i'd like to ensure we are surfacing any internal representations as far up as possible so it's clear what may need to change as we make these structural compiler changes.

areusch:
@Mousius okay sorry i found a code point to the area i was concerned about. good to merge/continue with this after we resolve that point. and cc @junrushao1994 @zxybazh who did some work on the target JSON format.

rfcs/0028-command-line-registry-composition.md
Mousius: Apologies, I'm not entirely sure what you're suggesting here, do you suggest `--target` takes JSON? Or some internal mechanism converts the arguments shown here into an intermediate representation for `target._load_config_dict` ?

zxybazh: Hi, just want to mention that `load_config_dict` is a function we use to parse target json strings in both python & c++ side using python json library. It is currently designed for single target though. I think it would take some effort, while feasible, to make modifications to work with heterogenes targets if we want to make the change.

PR #30, opened 2021-09-07
Command Line Configuration Files

Mousius:


hogepodge:
Can you address the common usage of environment variables for configuration, why that wasn't considered, and the potential for future work?

Mousius:
Hi @hogepodge, that's actually a great thing to include in the future possibilities (and I've done so). The reason it's not included is purely to scope the work down to what seemed a reasonable chunk, I don't disagree that environment variables would be a great addition but would likely come with their own rules and nuances to add to the functionality.

I've linked to [Terraform](https://www.terraform.io/docs/language/values/variables.html#environment-variables), as an example of where all three work really well in harmony in the future possibilities. I don't think I need to include them in the alternatives as they're additive in my mind, does that sound right to you?

u99127:
> Can you address the common usage of environment variables for configuration, why that wasn't considered, and the potential for future work?

I think the usage of environment variables for configuration of this kind in an AoT compilation flow with tvmc is quite debatable with the kinds of things it takes. .

Personally I think its cleaner to consider this directly in configuration files of this kind. 

I have had a read through and this looks pretty good LGTM

@leandron  could you please take a look ?

leandron (approved):
LGTM. I think this will be a great improvement for new TVM/`tvmc` users, without taking out the flexibility from the advanced users.

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+This would be used if the user simply specifies `--config=corstone-300`, as in the following example:
+```
+tvmc compile --config=corstone-300 my_model.tflite
+```
+
+## User Provided Configurations
+The default search path, as illustrated above, is to find a matching `<NAME>.json` to an argument `--config=<NAME>`. A user can instead specify a path in the `--config` argument such as:
+
+```bash
+--config=./my.json
```
areusch, line 68: can multiple `--config` be specified and then they are merged, or just one? how does merging work with other cmd line flags are given?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+--config=/etc/devices/my_secret_board.json
+```
+
+By default, TVM will prefer files explicitly specified as a path instead of hosted files.
+
+## Combination with existing parameters
+In the case of `tvmc`, `--config` will work alongside other arguments. Ideally anything specifiable in JSON will be specifiable in the command line to allow users to make small alterations such as:
```
areusch, line 75: usually we need to design this in. what's the "ideally" part here?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+  --config=corstone-300 \
+  --executor-aot-unpacked-api=0
+```
+
+Which allows experimentation with different parameters that can then be added to a custom JSON.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This will change the behaviour of how `tvmc` utilises `argparse`, it will first translate arguments from `argparse` into an internal dictionary of attributes and then apply those over the top of any specified configuration files. This means the default options for `argparse` are essentially nulled as they won't be aware of configuration files until after the arguments are parsed. The hierarchy is therefore:
```
areusch, line 89: can you explain how you merge complex object e.g. target?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+1. Arguments parsed by `argparse`
+2. Configuration file specified (defaults to `default`)
+3. Internal defaults for arguments in `tvmc`
+
+Because these results are additive the underlying defaults remain in `tvmc` rather than in `default.json` to ensure the user doesn't create a resulting configuration which additively makes no sense being based on an `llvm` target or other defaults.
```
areusch, line 94: couldn't you still do this when `--config` specifies one `llvm` and cmd line specifies another?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Although this presents a simpler interface for new users, the options become more complex for power users who want to mix and match arguments. This is mitigated by specifying simple rules for how these are composed.
+
+The convention of `--config` being read first then all other arguments is less intuitive than left to right parsing, but fits better with the current `argparse` infrastructure.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
```
areusch, line 106: while JSON5 seems okay, there are other config languages with comments e.g. yaml. could you speak to why you don't prefer that? looks like `json5` python package does not require any other packages--is that the one you propose to use?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+Such configuration files already exist in a number of platforms and tools to reduce the overhead on the user to define them, such as the [Zepyhr DeviceTree](https://github.com/zephyrproject-rtos/zephyr/tree/main/boards) or [gcc configurations for specific targets](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/gcc/config/arm/arm-cpus.in).
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
```
areusch, line 114: perhaps one is whether the direction of the Target specification is aligned with this (i think yes, but perhaps @junrushao1994 @mbs-octoml @jroesch could comment).

areusch (commented):
@Mousius thanks for this RFC. I would welcome `--config` option to `tvmc`. I'm wondering if there are possibilities of providing PCB-level configs this way--hence my question about merging configs.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: I added some examples, I've specified one `--config` and don't make this more complicated than is necessary. My feeling is we should limit ourselves to one `--config` to make it a simple hierarchy rather than multiple layers of merge-ability.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: Added:
```
Complex configurations which aren't easily represented well in CLI arguments may exist and can continue to be represented only in JSON.
```

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: Added examples, essentially it just takes your new `Target` parameters and overrides the previous ones but keeping previous attrs.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: I changed `additive` to merged, I meant the behaviour was `additive`, in reality the CLI would specify the targets with `--target=c,llvm` that would act on previous `Target` definitions but otherwise remove them.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: I linked to the specific `json5` package, the intention is to use JSON similar to other areas of TVM but with some more flexibility in terms of comments etc. As regards YAML, it lacks structure which seems completely out of place for a configuration format which doesn't change very often.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: I'll add it to `Prior Art` as I believe it's aligned reasonably well:
https://discuss.tvm.apache.org/t/rfc-tvm-target-specification/6844

This RFC is not for the format of the `Target`, it is `Target`ing adding configuration files - the specific fields in those files should be flexible to change after this RFC.

Mousius:
Hi @areusch, thanks for your feedback!

> I would welcome --config option to tvmc. I'm wondering if there are possibilities of providing PCB-level configs this way--hence my question about merging configs.

This is one of our core use-cases, which is why we have the folder `configs/boards` specified :smile_cat: 

I've addressed your other comments, could you take another look?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,165 @@
...
+# Motivation
+[motivation]: #motivation
+
+When a user first approaches TVM, choosing an appropriate configuration can be difficult, this is increasingly true in embedded systems where the configuration is not only a collection of devices but also how those devices are interfaced (see [Arm&reg; Corstone&trade;-300 reference package](https://developer.arm.com/ip-products/subsystem/corstone/corstone-300)). Trying to specify all of this in a target string or via command line arguments would be error prone and tedious. Predefining these in a common format allows users to get started and take the configurations for their own use cases easily.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## TVM Hosted Configurations
+Configurations will be stored as [JSON5](https://json5.org/) at `configs/<TYPE>/<NAME>.json`, this top level directory will enable other tooling to load configurations just as easily a `tvmc` and provide easy sign posting for users looking for configurations.
```
areusch, line 20: what can `<TYPE>` be?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,165 @@
...
+## Example: merging with new config
+Because these results are merged, the underlying defaults remain in `tvmc` rather than in `default.json` to ensure the user doesn't create a resulting configuration which additively makes no sense (for example, being based on an `llvm` target or other defaults). For example, the default in `tvmc` would be:
+```json
+{ "autotuning_runs": 10 }
+```
+Which would then be extended with `--config=default` (`{ "targets": [{ "kind": "llvm" }], "executor": { "kind": "graph", "system-lib": true } }`):
+```json
+{ "autotuning_runs": 10, "targets": [{ "kind": "llvm" }], "executor": { "kind": "graph", "system-lib": true } }
+```
+**Or** be extended by `--config=corstone300` (`{ "targets": [{ "kind": "c", "mcpu": "cortex-m55" }, { "kind": "ethosu" }] }`):
```
areusch, line 105: can you provide an example of what would happen if you supplied e.g. `--config=default --config=corstone300`?

rfcs/0030-tvmc-comand-line-configuration-files.md
```diff
@@ -0,0 +1,165 @@
...
+1. Arguments parsed by `argparse`
+2. Configuration file specified (defaults to `default`)
+3. Internal defaults for arguments in `tvmc`
+
+## Example: merging with new config
+Because these results are merged, the underlying defaults remain in `tvmc` rather than in `default.json` to ensure the user doesn't create a resulting configuration which additively makes no sense (for example, being based on an `llvm` target or other defaults). For example, the default in `tvmc` would be:
+```json
+{ "autotuning_runs": 10 }
+```
+Which would then be extended with `--config=default` (`{ "targets": [{ "kind": "llvm" }], "executor": { "kind": "graph", "system-lib": true } }`):
```
areusch, line 101: how does the "executor" schema map to the target string? targets seems straightfoward, but i think executor needs a refactor we haven't done yet (splitting the executor config out from Target).

areusch (commented):
thanks @Mousius for addressing my comments! i am basically good with this proposal outside of the merging aspect. i think it would be good to consider this a bit more--what i don't want to do is overcomplicate the `tvmc` tool and make it difficult to use or for us to debug. in particular i think there are a couple of aspects of config merging which aren't well-defined:
- when should specifying a target parameter delete the previous target vs append to the list of targets? it seems so far given the examples, the rules are roughly
  - on the command lineN:
    - when `-target=` is given on the command line, the set of targets is deleted
    - otherwise `-target-<name>-<attr>` is merged into the previous target.
  - in a config file:
    - if targets don't conflict: they are concatenated
    - if targets conflict: not sure I can tell, but i could be being dense here.


here are some thoughts:
 - I don't think it's likely someone is going to want to split the config at such as level as to require `--config=cortex-m4.json (`{"targets": {"kind": "llvm", "mcpu": "cortex-m4"}}`) and --config=fpu.json` (`{"targets": {"kind": "llvm", "mattr": "+fp"}}`) separately.
 - i think rather that someone may likely provide `--config=boards/abc.json` along with `--config=executor/aot.json`.
 - i also think that by choosing JSON5, we are making it possible for someone to easily write their own script or use e.g. `jq` to combine config files. i'm not sure we need to handle all the cases here in the command-line parser.

i'd suggest the following tweak:
 - attrs can be overridden on the command line but not deleted (e.g. target attrs or executor attrs)
 - otherwise, when merging configs, when combining top-level keys, you can only completely override, not merge
 - if someone wants to do something fancier, they write their own script to do it.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: It's intentionally not defined to allow for users to populate it as they see fit, I don't want to constrain the folders in that directory to requiring an RFC update. I provided `configs/boards` as an example use case but I imagine there'll be many ways of subdividing this.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: I finished a sentence above this which eludes to default `argparse` behaviour of taking `corstone300` over `default` in this case as only one configuration will be parsed.

rfcs/0030-tvmc-comand-line-configuration-files.md
Mousius: This is demonstrating a hypothetical configuration layout rather than a concrete layout, in this case I'm using the improvements from [Migrating Target Attributes to IRModule](https://github.com/apache/tvm-rfcs/pull/29) to allow for a more detailed example. In this case it doesn't make a difference as the example is for the merging of portions of the JSON rather than any specific value.

Mousius:
Hi @areusch, responding to your comments now :smile_cat: 

> thanks @Mousius for addressing my comments! i am basically good with this proposal outside of the merging aspect. i think it would be good to consider this a bit more--what i don't want to do is overcomplicate the `tvmc` tool and make it difficult to use or for us to debug. in particular i think there are a couple of aspects of config merging which aren't well-defined:
> * when should specifying a target parameter delete the previous target vs append to the list of targets? it seems so far given the examples, the rules are roughly
>   
>   * on the command lineN:
>     
>     * when `-target=` is given on the command line, the set of targets is deleted
>     * otherwise `-target-<name>-<attr>` is merged into the previous target.
>   * in a config file:
>     
>     * if targets don't conflict: they are concatenated
>     * if targets conflict: not sure I can tell, but i could be being dense here.
The hierarchy proposed within the RFC is:
1. Arguments parsed by `argparse`
2. Configuration file specified (defaults to `default`)
3. Internal defaults for arguments in `tvmc`
This means the `tvmc` arguments will win out over config and both win out over defaults. The motivation for this is similarly to not overcomplicate the user interface and provide simple rules.

> here are some thoughts:
> 
> * I don't think it's likely someone is going to want to split the config at such as level as to require `--config=cortex-m4.json (`{"targets": {"kind": "llvm", "mcpu": "cortex-m4"}}`) and --config=fpu.json` (`{"targets": {"kind": "llvm", "mattr": "+fp"}}`) separately.
> * i think rather that someone may likely provide `--config=boards/abc.json` along with `--config=executor/aot.json`.
> * i also think that by choosing JSON5, we are making it possible for someone to easily write their own script or use e.g. `jq` to combine config files. i'm not sure we need to handle all the cases here in the command-line parser.
> 
> i'd suggest the following tweak:
> 
> * attrs can be overridden on the command line but not deleted (e.g. target attrs or executor attrs)
> * otherwise, when merging configs, when combining top-level keys, you can only completely override, not merge
> * if someone wants to do something fancier, they write their own script to do it.
This RFC does not propose the merging of configuration files, instead prioritising a single configuration - as you've mentioned this is more complexity than we really need. The hierarchy from the RFC shown above should cater to how overrides are allowed to work without providing much complexity?

So I think we're in agreement that users can then construct their own configurations and only need one configuration - I'm just suggesting the CLI serves are the final override point so you can load a config and change options easily. What do you think @areusch?

rfcs/0030-tvmc-comand-line-configuration-files.md
areusch: ok can we add this note to the PR?

rfcs/0030-tvmc-comand-line-configuration-files.md
areusch: got it. can we add that as a note here?

rfcs/0030-tvmc-comand-line-configuration-files.md
areusch: ah i see it now. thanks!

areusch (commented):
ok great i see @mousius! i think there are just a few last behaviors i want to make sure we align on:
1. if the config file specifies "targets" key, all targets from the default config are deleted/overridden.
2. if the command-line supplies `--targets=`, all targets from the config file are deleted/overridden.
3. if the command-line only supplies e.g. `--target-llvm-mcpu`, then it modifies the llvm target from config/defaults.
4. if the command-line or a config file specifies a target sub-key, it always overrides it in full (e.g. there is no appending to `mattr` from the command line).

Mousius:
> ok great i see @Mousius! i think there are just a few last behaviors i want to make sure we align on:
> 
> 1. if the config file specifies "targets" key, all targets from the default config are deleted/overridden.
> 2. if the command-line supplies `--targets=`, all targets from the config file are deleted/overridden.
> 3. if the command-line only supplies e.g. `--target-llvm-mcpu`, then it modifies the llvm target from config/defaults.
> 4. if the command-line or a config file specifies a target sub-key, it always overrides it in full (e.g. there is no appending to `mattr` from the command line).

@areusch yip yip yip, I've added this to the document and referenced supporting RFCs :smile_cat:

leandron (approved):
LGTM

PR #31, opened 2021-09-07
C Device API

Mousius:


rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+uart_dev->Send(data, sizeof(data)/sizeof(data[0]));
+```
+
+In this example, you can see the diversity of RTOS implementations for drivers and why it's required to provide a flexible abstraction to pass devices for micro targets.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## User App
+`tvm_device_<device>_t`s are implemented for each RTOS or platform required, these are included by the user who chooses as appropriate for their application. Notably, to avoid dynamic allocation, the user must provide the `tvm_device_<device>_t` struct and initialise it rather than it being created and setup for them in the API. This is augmented by named functions for each device, examples in the case of the "woofles" accelerator:
```
areusch, line 44: clearer would be "TVM presumes that the RTOS, platform, or user application defines a struct type `tvm_device_<device>_t`." Also state in the last sentence here that the user must implement a set of named functions.

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+
+```c
```
areusch, line 46: can we include my life cycle picture and annotate with where each function is called? also can we document exactly why TVM calls each function and what it's expected to do (e.g. pre- and post-conditions)?

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+int32_t ret = tvmgen_mynetwork_run(
+    ...,
+    &devices
+);
+
+TVMDeviceDestroy(&accelerator);
+```
+
+## Platform Structures
+Users can take a implementations from `src/runtime/crt/device` and headers from `include/runtime/crt/device` which maps to their platform device implementation. In the case of a bare metal environment, this would default to a void pointer as there's no information available.
```
areusch, line 78:
```suggestion
Users can take a implementations from `src/runtime/crt/device` and headers from `include/runtime/crt/device` which maps to their platform device implementation. The simplest definition of `tvm_device_<device>_t` is `void*` as no information is provided to TVM.
```

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+
+```c
+#include <device.h>
+
+typedef struct {
+    struct device* dev;
+} tvm_device_woofles_t;
+```
+
+This enables the OS maximum control over the resources required and provides the opportunity to craft code in whichever way is most idiomatic for that platform, such as if an additional locking mechanism is required:
```
areusch, line 94:
```suggestion
This enables the OS maximum control over the resources required, allows the user application to consolidate the memory used by the device control structures, and provides the opportunity to craft code in whichever way is most idiomatic for that platform, such as if an additional locking mechanism is required:
```

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+    tvm_dev->device = zephyr_dev;
+}
+```
+This simple wrapper enables type checking of these functions and defining a clear translation boundary between the underlying OS implementation and TVM.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Entrypoint
+The entrypoint API defined in [Embedded C Runtime Interface](https://discuss.tvm.apache.org/t/rfc-utvm-embedded-c-runtime-interface/9951) is augmented with the `devices` structure which contains implemented `tvm_device_t` `struct`s for each device used by the network:
```
areusch, line 130:
```suggestion
The entrypoint API defined in [Embedded C Runtime Interface](https://discuss.tvm.apache.org/t/rfc-utvm-embedded-c-runtime-interface/9951) is augmented with the `tvm_mynetwork_devices` structure which contains implemented `tvm_device_t` `struct`s for each device used by the network:
```

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+These are re-cast to `void *` when entering the AOT main function to pass it without TIR understanding the struct types.
+
+```c
+int32_t tvmgen_mynetwork_run(
+    ...,
+    struct tvmgen_mynetwork_devices* devices
+) {
+    tvmgen_mynetwork_run_model(
+        ...,
+        devices->host,
```
areusch, line 146: shall we just pass the devices struct and use `tir.tvm_struct_get`?

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+    );
+}
+```
+
+## Executor Function
+Each operator is provided with a single device object which can be abstracted and passed as the `void* resource_handle`. The main function calls into the device API to setup and teardown resources before and after each operator call.
+
+```c
+int32_t tvmgen_mynetwork_run_model(..., device0, device1) {
+    TVMDeviceWooflesActivate(device0); // Could reserve or enable certain circuitry ahead of time
```
areusch, line 157: would be great to say something about the memory and perhaps include a note that memory copies into the device API-managed memory may occur before TVMDeviceWooflesOpen

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+int32_t TVMDeviceUartAccelClose(tvm_device_uart_accel_t* tvm_dev) {}
+int32_t TVMDeviceUartAccelDeactivate(tvm_device_uart_accel_t* tvm_dev) {}
+
+int32_t TVMDeviceUartAccelDestroy(tvm_device_uart_accel_t* tvm_dev) {
+    tvm_dev->dev->Uninitialize();
+}
+```
+
+## Operator Usage
+Each operator would be expected to utilise one device structure and be passed that as the `resource_handle` parameter, making the assumption that each operator or variant of an operator is only bound to one device at a time. In the following example it can be seen how a accelerators interface is implemented per platform to take this void pointer and call the platform specific driver code.
```
areusch, line 231: could you specify how TVM knows which device structure an operator uses? also, can you qualify this is just for the C runtime and c++ runtime when BYOC targets are linked in?

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+    struct device* zephyr_dev = tvm_dev->dev;
+    my_accelerator_invoke(
+        zephyr_dev,
+        ...ins,outs,params...
+    );
+}
+```
+
+## PrimFunc Resource Handle
+A `tir::Var` is added to `PrimFunc` in `include/tvm/tir/function.h` which enables a `PrimFunc` to track and use the `resource_handle` parameter. This will be used by both unpacked and packed APIs to pass the resource down without packing into `TVMValue`, instead as a `void *`.
```
areusch, line 252: it kind of seems like we could get away without doing this by tracking which Target a function is going to run on and looking up the associated device API function in AOT/graph codegen. what do you think?

rfcs/0031-devices-api.md
```diff
@@ -0,0 +1,363 @@
...
+operator(..., device1) {
+    // some work where device1 can read from memory directly
+    // then the result is copied back
+    TVMDeviceWooflesCopyFrom(device1, &buffer);
+}
+TVMDeviceWooflesClose(device1);
+
+TVMDeviceUartAccelOpen(device2);
+operator(..., device2) 
+    TVMDeviceUartAccelCopyTo(&buffer, device2);{
```
areusch, line 335: what if memory is needed across operator calls? does the device have to be Open'd for e.g. the whole runtime of the model when e.g. a parameter is pre-loaded and the device call is the last layer in the model?

areusch (commented):
thanks @Mousius just a couple questions here

rfcs/0031-devices-api.md
Mousius: I added some words to that effect, essentially stating that the memory flow is outside this RFC as it stands right now but we should be aware that means it could happen at any time.

rfcs/0031-devices-api.md
Mousius: I think this is included further down the RFC in the outline of an executor and explored further in future work? The hooks are intentionally kept simply named so they can be interpreted for a device, from my experience the `TVMDeviceWooflesActivate` could span from setting a pin high to sending a series of commands over a communications bus? The main thing I'm trying to communicate is that `Activate` is called before `Open` which is documented in the executor flow, this might not be the best way of demonstrating it though so please feel free to suggest an alternative.

rfcs/0031-devices-api.md
Mousius: I've tweaked it a bit :smile_cat:

rfcs/0031-devices-api.md
Mousius: I'd rather avoid using the BYOC branding and sticking to the technical terms given BYOC is fairly heavily overloaded, I've added a few more words here to illustrate the `Target` configuration of this.

rfcs/0031-devices-api.md
Mousius: This is an example API for memory management in future possibilities, there's a much larger question as to how devices can specify all of the above as I'm unaware of which devices you're referring to and the specific characteristics in tandem with the model you're using.

rfcs/0031-devices-api.md
Mousius: We'd still need to pass this into the function call though to the operator though?

Mousius:
@areusch I've updated this and replied to your comments, could you please take another look - I believe implementing this is now blocked on the `PrimFunc` discussion but we should be able to land this and begin work as soon as that's resolved.

rfcs/0031-devices-api.md
areusch: i think the reason i'm hoping for a life-cycle diagram is that the actions Activate/Open could take are quite varied. i agree the call scheme below is laid out in call order, but there's nothing explicit which states this so you just have to know that in advance. could you add something more explicit? i think the life-cycle diagram is the best thing as we can iterate on it as we add features here.

rfcs/0031-devices-api.md
```diff
@@ -41,7 +41,7 @@ In this example, you can see the diversity of RTOS implementations for drivers a
 [guide-level-explanation]: #guide-level-explanation
 
 ## User App
-`tvm_device_<device>_t`s are implemented for each RTOS or platform required, these are included by the user who chooses as appropriate for their application. Notably, to avoid dynamic allocation, the user must provide the `tvm_device_<device>_t` struct and initialise it rather than it being created and setup for them in the API. This is augmented by named functions for each device, examples in the case of the "woofles" accelerator:
+TVM presumes that the RTOS, platform, or user application defines a struct type `tvm_device_<device>_t`, these are included by the user who chooses an implementation as appropriate for their application. Notably, to avoid dynamic allocation, the user must provide the `tvm_device_<device>_t` struct and initialise it rather than it being created and setup for them in the API. TVM then expects an implementation of the named functions for each device, examples in the case of the "woofles" accelerator:
```
areusch, line 44: could you also explicitly highlight where `<device>` comes from from a user PoV?

rfcs/0031-devices-api.md
```diff
@@ -127,7 +127,7 @@ This simple wrapper enables type checking of these functions and defining a clea
 [reference-level-explanation]: #reference-level-explanation
 
 ## Entrypoint
-The entrypoint API defined in [Embedded C Runtime Interface](https://discuss.tvm.apache.org/t/rfc-utvm-embedded-c-runtime-interface/9951) is augmented with the `devices` structure which contains implemented `tvm_device_t` `struct`s for each device used by the network:
+The entrypoint API defined in [Embedded C Runtime Interface](https://discuss.tvm.apache.org/t/rfc-utvm-embedded-c-runtime-interface/9951) is augmented with the `tvm_mynetwork_devices` structure which contains implemented `tvm_device_t` `struct`s for each device used by the network:
```
areusch, line 130: i forget if we discussed this already, but @manupa-arm 's RFC used a different casing so let's just quickly address this here: google c++ styleguide would have us call this `TvmGenMynetworkDevices`. I kind of like the `_` separator however, since this name is generated.

rfcs/0031-devices-api.md
```diff
@@ -306,6 +293,8 @@ Another route to take is to treat RTOSes as entirely separate from TVM, requirin
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
+Is coupling `Target`s with this Device API to the C Runtime acceptable? From the authors point of view this seems an acceptable trade off given the devices won't function correctly without the flow control.
```
areusch, line 296: i think it's likely that [TargetDevice](https://github.com/apache/tvm/pull/8892) will play a role here.

rfcs/0031-devices-api.md
Mousius: I've intentionally left this as the C++ style as that's the current standard in TVM, @areusch suggested changing it [would likely require an RFC in itself and updating the style guide](https://github.com/apache/tvm/pull/8280#issuecomment-888554759).

rfcs/0031-devices-api.md
Mousius: Is this not covered by the example just below this block?

Mousius: Nope, that was not the question, amended the text to clarify the origin.

rfcs/0031-devices-api.md
Mousius: Is there a `TargetDevice` RFC I can reference? Otherwise I'd assume `TargetDevice` is speculative rather than something we should include here?

Mousius:
@areusch this one is ready for a look again :smile_cat:

rfcs/0031-devices-api.md
areusch: sorry not yet :/. it's sort of teased in the future work section of my [Artifact pre-RFC](https://discuss.tvm.apache.org/t/introduce-artifact-a-container-for-generated-code/11099). i'm more including this pointer here as an "early FYI" how i'd like to solve this (e.g. using the TargetDevice name as the mechanism to bind a sub-target to a Device API impl).

areusch (approved):

areusch:
thanks @Mousius !

PR #36, opened 2021-09-23
[RFC][TVMScript] New namespace for tvmscript

Hzfengsy:
It's a rfc for changing TVMScript namespace to enable auto-completion support and pass pylint checks.

### DIFF
--- a/rfcs/0036-tvm-script-namespace.md
+++ b/rfcs/0036-tvm-script-namespace.md
@@ -1,6 +1,6 @@
 - Feature Name: TVM Script Namespace
 - Start Date: 2021-09-23
-- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- RFC PR: [apache/tvm-rfcs#0036](https://github.com/apache/tvm-rfcs/pull/36)
 - GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
 
 # Summary
--- a/rfcs/0036-tvm-script-namespace.md
+++ b/rfcs/0036-tvm-script-namespace.md
@@ -1,0 +1,89 @@
+- Feature Name: TVM Script Namespace
+- Start Date: 2021-09-23
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+
+[summary]: #summary
+
+This is a RFC for the new namespace and user interface for TVM script.
+
+- Use tvm.script as the `root` namespace for all TVM script related stuff
+- Use `tvm.script.tir` for `TIR`, and idiomatically import it as `T`, like Keras is usually imported as K
+- To be consistent with the names of their resulting types, use
+  - `tvm.script.ir_module` for `IRModule`
+  - `T.prim_func` for `tir.PrimFunc`
+
+# Motivation
+
+[motivation]: #motivation
+
+TVMScript is a fancy tools to write TIR in Python syntax. However, it's not fully compatible with native Python tools (e.g., pylint) because it is not a runnable program in Python.
+
+We have following pain points:
+
+- No Python auto-completion support
+- Usually conflicts with pylint
+- APIs scatter in namespaces like tvm.script, tvm.tir, tvm.script.ty
+- Somewhat non-trivial to understand at first glance what the decorator generates
+
+With the new proposal, we are able to provide type stubs that provides users with TVM scripts that work well with linting and auto-completion.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Let's see an example for the same TIR program but in different syntax.
+
+## Existing Syntax
+
+```Python
+from tvm import tir
+# ^ here tir is not related to the script but we need to import it. Otherwise there are many lint errors.
+from tvm.script import ty
+# ^ ty is a sub-module for only type system, which makes no sense to import it global
+
+@tvm.script.tir
+class Module:
+  def func(a: ty.handle, b: ty.handle) -> None:
+    A = tir.match_buffer(a, [128, 128], dtype="float32")  # pylint error: tir does not have attr match_buffer
+    B = tir.match_buffer(b, [128], dtype="float32")
+    with tir.block([128, tir.reduce_axis(0, 128)], "C") as [i, k]:
+        B[i] += A[i, k]
+```
+
+## New Syntax
+
+```Python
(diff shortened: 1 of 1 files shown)


rfcs/0036-tvm-script-namespace.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## New Syntax
+
+```Python
+from tvm.script import tir as T
+# ^ there is a broadly accepted precedence in doing this in the python community:
+#   from keras import backend as K
+
+@tvm.script.ir_module                                   # it generates an IRModule
```
tqchen, line 63: should it be IRModule directly or ir_module?

rfcs/0036-tvm-script-namespace.md
```diff
@@ -0,0 +1,89 @@
...
+class Module:
+  @T.prim_func                                          # it generates a PrimFunc explicitly
```
tkonolige, line 65: Current we do not require decorators for functions within a class decorated with `tvm.script.tir`. Are you suggesting we change this so all functions need a decorator?

rfcs/0036-tvm-script-namespace.md
```diff
@@ -0,0 +1,89 @@
...
+
+[reference-level-explanation]: #reference-level-explanation
+
+Pylint and Python auto-completion will look the symbols (function definitions) in the module (e.g. `tvm.script.tir`). Just write (or generate) a concrete function in the module, then can enable auto-completion and pass Pylint checks.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+Here are some existing works based on current TVM Script syntax. It need a huge refactor to migrate it to the new one.
```
tkonolige, line 83: Is there a way we could deprecate but not remove the old syntax. It might make it easier to transition. (Though it might be impossible because `tir` is changing from a function to a module. Stackoverflow suggests some _very_ hacky solutions).

tkonolige (approved):
I really like this change. Can't wait for autocomplete!

rfcs/0036-tvm-script-namespace.md
Hzfengsy: I don't think so. Actually I have refactored all testcase on my local machine. Will pr very soon. The only problem is how to review such huge refactoring.

rfcs/0036-tvm-script-namespace.md
Hzfengsy: Personally prefer `ir_module`. Because decorators are python functions, which usually use `snake_case` style

rfcs/0036-tvm-script-namespace.md
Hzfengsy: Good question. I think it is needed in the long term (since IRModule contains not only `PrimFunc` but also functions for relay). However, only PrimFuncs are supported by TVMScript by now. It's OK not to decorate functions in this PFC.

Love to hear your opinions.

tkonolige: Yeah, I agree we need it in the future. I don't really care if we add it now or later, we should just make the decision explicit if we do add it.

rfcs/0036-tvm-script-namespace.md
```diff
@@ -0,0 +1,89 @@
...
+#   from keras import backend as K
+
+@tvm.script.ir_module                                   # it generates an IRModule
+class Module:
+  @T.prim_func                                          # it generates a PrimFunc explicitly
+  def func(a: T.handle, b: T.handle):                   # return type is not necessary for PrimFunc
+    A = T.match_buffer(a, [128, 128], dtype="float32")  # no pylint errors for match_buffer
+    B = T.match_buffer(b, [128], dtype="float32")
+    with T.block([128, T.reduce_axis(0, 128)], "C") as [i, k]:
+        B[i] += A[i, k]
```
MasterJH5574, line 70: I remember TVM script didn't support the `+=` operation. Did we support it in some PR before, or we actually should write the body like `B[i] = B[i] + A[i, k]`?

rfcs/0036-tvm-script-namespace.md
```diff
@@ -0,0 +1,89 @@
...
+from tvm.script import ty
+# ^ ty is a sub-module for only type system, which makes no sense to import it global
+
+@tvm.script.tir
+class Module:
+  def func(a: ty.handle, b: ty.handle) -> None:
+    A = tir.match_buffer(a, [128, 128], dtype="float32")  # pylint error: tir does not have attr match_buffer
+    B = tir.match_buffer(b, [128], dtype="float32")
+    with tir.block([128, tir.reduce_axis(0, 128)], "C") as [i, k]:
+        B[i] += A[i, k]
```
MasterJH5574, line 53: Ditto here.

MasterJH5574 (commented):
I noticed a minor point. @Hzfengsy Please take a look! Thanks!

rfcs/0036-tvm-script-namespace.md
MasterJH5574: Update: I had a quick check, and found that `+=` is well-supported 👍. So just resolve the comments :-)

altanh:
I support this change. One suggestion is that we keep a canonical namespace that will always be parsed correctly to the TIR namespace (e.g. `tir.block` will always work), and in particular use this identifier during printing. Mentioning this here since I noticed the PR apache/tvm#9115 is printing using the `T` alias, which seems problematic unless we print an import statement at the top.

tqchen (approved):

Hzfengsy:
Thanks, @altanh. Your suggestion makes sense to me. To be specific, here are two cases: parse from a python script and string.
1. When we parse from a python script, we detect the prefix `T` from the python env (through function `__globals__`, i.e. you can even use `XXX.block` if with `from tvm.script import tir as XXX`)
2. When we parse from a text string (e.g. printed from tir), we can specify the accept prefixes by `from_source(source_code, tir_prefix=["T", "tir"])` (default accept both `tir` and `T`). Also, we can choose the prefix during printing by `func.script(tir_prefix=“tir”)` (default `tir`)

I would update the PR with the new proposal soon. Please feel free to left your comments about the changes.

tqchen:
Great discussions. If we are encouraging a common idiom(e.g. use T for tir).

We might actually want to encourage print in T as prefix to be consistent on what users might write(and further edit with auto-completion), assuming we always encourage user write `from tvm.script import tir as T`. As long as we agree to that convention. Having a prefix option indeed makes the choice flexible.

junrushao1994 (approved):

junrushao1994:
Thanks guys for the discussion! If the comments and concerns are all addressed, let's move on and merge this PR after 24h

altanh (approved):

junrushao1994:
Thanks guys for the discussion! The RFC is merged :-)

PR #38, opened 2021-09-29
[RFC] Unified device/target/memory scope planning

mbs-octoml:
In rendered form:
https://github.com/mbs-octoml/mbs-tvm-rfcs/blob/mbs-target-and-device-planning/rfcs/0038-unified-device-target-and-memory-scope-planning.md

Some earlier discussion:
https://github.com/apache/tvm/pull/8892

Tracking issue: https://github.com/apache/tvm/issues/9327
(CORE-95 in the OctoML JIRA).


manupa-arm:
Hi @mbs-octoml ,

I may have put a related comment here : https://github.com/apache/tvm/pull/8892#issuecomment-932020564
However, partitioning for devices of same kind is a step forward from unifying the BYOC and Device annotations.

Is this the RFC intended to cover these all ?

mbs-octoml:
Thanks @manupa-arm for the reminder there were some good comments on #8892. I see a work stream:
1. get the multi-target handling under control, and stop relying on the limiting device type -> target mapping
2. allow device planning to be re-run with additional memory scope constraints to further partition, those constraints may originate in already lowered PrimFuncs.
3. allow labeling of sets of target/devices a la BYOC target labeling (but probably continue to just pick the 'first')
4. bring BYOC target labeling / partitioning and device planning together
5. replace naive partitioning with something subject to optimization

Here I want to just focus on 1 -- everything beyond that really needs face-to-face discussion.
 
From 2 onward obviously overlaps your USMP. My vague thought was we can work from opposite ends and reconcile at 5. Ie 2-4 sets us up to work in a combined Relay+TIR world, then 5 is where everything we've learned from USMP could perhaps be replayed. Anyway, that's just a vague thought so I'd love to talk more about it.

mbs-octoml:
Note to self: The With<Target> convention should probably also be removed by this work also, but I've not audited the code to see how pervasive it is.

rfcs/00xx-improved-multi-target-handling.md
```diff
@@ -0,0 +1,176 @@
...
+   every primitive operator call already found by device planning.
+6. Bind two `TargetDevice`s as attributes on every `IRModule`:
+    - The default for primitive operators not otherwise constrained by "on_device" annotations.
+    - The default for non primitive operators, such as Relay control flow and shape computation.
+7. We remove the various copies of target/target_host reconciliation, `TargetMap`
+   construction and 'default/fallback' device calculation from the codebase.
+
+This proposal tackles the original problems:
+1. There's now no ambiguity about `Targets` since we propagate them from the global registry directly.
+2. We support device ids.
```
areusch, line 75: it would be great to clarify between device IDs or target_label. Device ID is very much a runtime thing and we can't always control which e.g. GPU gets mapped to 0 or 1.

areusch (commented):
@mbs-octoml thanks for the draft RFC! added some thoughts

tqchen:
cc @ zxybazh since you authored the Target system, cc @zhiics @comaniac as it is related to BYOC

comaniac:
For the proposed BYOC flow (i.e., `MergeAndAnnotate`/`AnnotateSEScopes`/`PlanDevices`/`PartitionBySEScope`), it doesn't clear to me that whether the developer programming model will change or not. Specifically, could we still use the current approaches (i.e., op-based annotation and pattern-based annotation)? Also how the approach that has a custom Relay pass to annotate `compiler_begin`/`compiler_end` would change? Thanks.

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+   `DLDeviceType` no longer uniquely determines a `Target`.
+2. Though TVM's `Device` abstraction (an alias for `dlpack`'s `DLDevice`) is a pair of a
+   `DLDeviceType` and an arbitrary 'device id', TVM does not consistently plumb the device id
+   through annotations, passes and operators.  Thus currently we cannot use 'device id' to
+   distinguish, eg, two CPUs in the same system.
+3. Upcoming work requires us to distinguish and propagate memory scopes for data at the Relay
+   level. (See also [RFC #9](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0009_Unified_Static_Memory_Planning.md)
+   which has a similar need for memory scope propagation at the TIR level). This is an identical
+   problem to propagating devices, and it seems most natural to simply combine targets, devices and
+   memory scopes into a single 'target of device planing' rather than implementing a whole new pass.
```
areusch, line 55: i agree, but just clarifying we aren't using a single identifier to describe both the device and the memory scope?

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+4. Device planning currently has no machinery to hoist adjacent expressions which share the same device
+   into their own Relay `Function`. For all our executors except VM that's unnecessary anyway since
+   all Relay expressions left over after lowering are interpreted by the runtime. However for AOT we
+   have to compile *all* Relay code for a particular target. Note the BOYC machinery does support this,
+   but for the purposes of redirecting the compilation flow entirely. We need a middle ground.
```
areusch, line 60:
```suggestion
   have to compile *all* Relay code for a particular target. Note the BYOC machinery does support this,
   but for the purposes of more accurately modeling offloaded compute in the main compilation flow, we need a middle ground.
```

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+   Taking the 'upper bound' of the two implementations seems ideal, especially to address issues 4 (limitation
+   of device planning) and 5 (limitation of BYOC) above.
+
+Our proposal is:
+1. We introduce a new FFI-friendly class to represent a *S*torage or *E*xecution *Scope*:
+
+   ```
+   class SEScope {
+     DLDeviceType device_type;
+     int virtual_device_id;
```
areusch, line 84: i think this should be a String name which makes sense to the user. Doing this is helpful for a couple other reasons besides the compilation UI:
- In generated source code, it's possible to refer to the device by name. In particular, the embedded C API would like to have this for the conglomerate tvm_device_t struct.
- In systems with multiple e.g. CPUs, using an index here then implies some ordering (e.g. littlest CPU to biggest). It's better to make the assignment of ID to CPU capability more explicit

Finally, using a name would simplify the heterogeneous Target.

However, this is a bit of a lift. I do feel strongly we should get to this world. If it's not something that makes sense to do now, we could also revisit after or concurrent with USMP.

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+   [RFC #9](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0009_Unified_Static_Memory_Planning.md)
+   has landed.
+
+3. The `on_device` and `device_copy` call attributes use `SEScope`s instead of integers. However the Python
+   bindings for these 'operators' continue to accept a `Device` for convenience. The machinery in `LowerTEPass`
+   which resolves `DLDeviceTypes` to `Targets` is moved up in the compilation flow and becomes part of
+   `PlanDevices`. In particular, any `SEScope` encountered during device planning is 'canonicalized' to fill
+   in a `Target` by the same lookup as we do today. This means we continue to support the easy shorthand of
+   referring to devices by the `DLDeviceType` alone. However, advanced users can supply a `SEScope` to these
+   operators which contains the exact `Target` to use.
```
areusch, line 107: what would be roughly the deprecation plan here? eventually we ban all the inputs to the compiler which could refer to SEScope in terms of DLDeviceType and then tighten the typing requirements here? this would be a backwards-incompatible Relay change. cc @jroesch

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+      (Currently this is hardcoded to `kDLCPU`).
+
+5. We extend `PlanDevices` to be able to a) run *after* lowering and b) refine existing constraints.  It will
+   look inside calls to `PrimFunc`s and follow the chain:
+
+   ```
+   tir::PrimFunc.buffer_map -> tir::Buffer.data -> tir::Var.type_annotation -> PointerType.storage_scope -> String
+   ```
+
+   to discover the memory scope for each Relay argument. That scope will enter `SEScope`s and flow through the
```
areusch, line 123: what do you mean by "enter `SEScope`s"?

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+   existing unification machinery. The existing sub-pass in `PlanDevices` will insert `device_copy` calls
+   wherever sub-expressions disagree on their memory scope.
+
+   (An additional pass is planned to heuristically move `device_copy`s around, and eliminate redundant
+    copies, however that's outside the scope of this RFC.)
+
+6. We rework `PartitionGraph` to `PartitionBySEScope` to work on `SEScope` annotations instead of
+   `compiler_begin` and `compiler_end` annotations. Algorithmically it's not a big change -- maximal
+   sub-expressions which share the same `SEScope` (or a projection thereof, eg just the `target`) are hoisted
+   into global `Function`s. The function's `"result_se_scope"` attribute describes both the scope holding the
```
areusch, line 133: so then here, this sort of implements the "grouping adjacent expressions onto the same device" as a side-effect?

rfcs/0038-unified-device-target-and-memory-scope-planning.md
```diff
@@ -0,0 +1,244 @@
...
+   function's result *and* the `Target` for which the function is to be compiled.
+
+7. We allow `MergeComposite` to be used to insert `on_device` annotations, call it `MergeAndAnnotate`.
+
+8. (?) We rework `AnnotateTarget` to just look for `FTVMAnnotateTarget` operator attributes, call it
+   `AnnotateSEScopes`. When the function fires an `on_device` annotation is inserted. However since
```
areusch, line 139: clarifying my understanding:
```suggestion
   `AnnotateSEScopes`. When `FTVMAnnotateSEScopes` returns true, an `on_device` annotation is inserted. However since
```

areusch (changes requested):
@mbs-octoml some comments/clarifications, can you also remove/address the boilerplate?

mbs-octoml:
Closing as obsolete, since most of this is either already done or has been subsumed by the Collage proposal.

PR #42, opened 2021-10-28
[TIR] Adding BufferPointer node, used by BufferLoad and BufferStore

Lunderberg:
This RFC introduces a `BufferPointer` node, to be used by `BufferLoad` and `BufferStore` nodes.

[Rendered markdown link](https://github.com/Lunderberg/tvm-rfcs/blob/buffer_pointer/rfcs/0042-buffer-pointer.md)

### DIFF
--- a/rfcs/0042-buffer-pointer.md
+++ b/rfcs/0042-buffer-pointer.md
@@ -1,7 +1,7 @@
-- Feature Name: (fill me in with a unique identifier, `my_awesome_feature`)
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Feature Name: BufferPointer
+- Start Date: 2021-10-28
 - RFC PR: [apache/tvm-rfcs#0042](https://github.com/apache/tvm-rfcs/pull/0042)
-- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+- GitHub PR: [PR#9391](https://github.com/apache/tvm/pull/9391)
 - Related RFCs: [RFC#0039](https://github.com/apache/tvm-rfcs/pull/0039)
 
 # Summary
@@ -77,7 +77,9 @@ become more verbose.
 The `BufferLoad::pointer` and `BufferStore::pointer` could be generic
 `PrimExpr`, instead of being `BufferPointer` objects.  This would
 require the datatype to be a handle, with an additional parameter to
-indicate what is being stored.  However, this 
+indicate what is being stored.  However, this would require all
+visitors that interact or modify a buffer to be duplicated across both
+reads and writes.
 
 Currently, the `BufferLoad::pointer` and `BufferStore::pointer`
 objects are visited by `ExprMutator` and `StmtMutator`, but are
--- a/rfcs/0042-buffer-pointer.md
+++ b/rfcs/0042-buffer-pointer.md
@@ -1,7 +1,8 @@
 - Feature Name: (fill me in with a unique identifier, `my_awesome_feature`)
 - Start Date: (fill me in with today's date, YYYY-MM-DD)
-- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- RFC PR: [apache/tvm-rfcs#0042](https://github.com/apache/tvm-rfcs/pull/0042)
 - GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+- Related RFCs: [RFC#0039](https://github.com/apache/tvm-rfcs/pull/0039)
 
 # Summary
 [summary]: #summary
@@ -11,6 +12,8 @@ A location being access in a buffer is represented as a
 accessed, and an array of indices to specify the location.
 `BufferLoad` and `BufferStore` objects each hold a `BufferPointer`
 object to specify where they operate.
+
+
 
 # Motivation
 [motivation]: #motivation
@@ -23,6 +26,10 @@ of a matched buffers to instead be direct access of the backing
 buffer.)  By having a `BufferPointer` object to represent a pointer
 into a buffer's memory, these transformations can be done without
 repeated code.
+
+This is intended to make the layout transformations specified in
+[RFC#0039](https://github.com/apache/tvm-rfcs/pull/0039) be more
+straightforward to implement, but is not strictly required for it.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
--- a/rfcs/XXXX-buffer-pointer.md
+++ b/rfcs/XXXX-buffer-pointer.md
@@ -1,0 +1,109 @@
+- Feature Name: (fill me in with a unique identifier, `my_awesome_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+[summary]: #summary
+
+A location being access in a buffer is represented as a
+`BufferPointer` object, which holds a reference to the buffer being
+accessed, and an array of indices to specify the location.
+`BufferLoad` and `BufferStore` objects each hold a `BufferPointer`
+object to specify where they operate.
+
+# Motivation
+[motivation]: #motivation
+
+`BufferLoad` and `BufferStore` both act on a location within a buffer,
+and must know where read/write their respective values.  However, many
+transformations are unconcerned with whether an access is a read or a
+write.  (e.g. `tir.transform.LowerMatchBuffer`, which rewrites access
+of a matched buffers to instead be direct access of the backing
+buffer.)  By having a `BufferPointer` object to represent a pointer
+into a buffer's memory, these transformations can be done without
+repeated code.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `BufferPointer` object contains a reference to the buffer that it
+acts upon, and the indices at which the access is being performed.
+
+`BufferPointer` also provides a utility function
+`BufferPointerNode::value_dtype()`, which returns the expected
+datatype at the specified location.  This will typically be the same
+as the buffer's datatype, but may have a different number of lanes.
+For example, a `BufferPointer` whose buffer's datatype is `float16`,
+and whose index is `Ramp(pos, stride=1, lanes=4)` for vectorized
+access will return `float16*4` for `value_dtype()`.
+
+Previously, `BufferLoad` and `BufferStore` held references to the
+buffer and indices directly.  To migrate these codes, replace
+references to `BufferX::buffer` with `BufferX::pointer::buffer` and
+replace references to `BufferX::indices` with
+`BufferX::pointer::indices`.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, 
+and explain more fully how the detailed proposal makes those examples work.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Requires an additional indirection to access the buffer and pointer,
+so transformations that must distinguish between reads and writes
+become more verbose.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The `BufferLoad::pointer` and `BufferStore::pointer` could be generic
+`PrimExpr`, instead of being `BufferPointer` objects.  This would
+require the datatype to be a handle, with an additional parameter to
+indicate what is being stored.  However, this 
+
+Currently, the `BufferLoad::pointer` and `BufferStore::pointer`
+objects are visited by `ExprMutator` and `StmtMutator`, but are
+required to be `BufferPointer` objects.  This is implemented in
+type-checking in `ExprMutator::VisitExpr_(BufferLoad*)` and
+`StmtMutator::VisitStmt_(BufferStore*)`, but it isn't apparent at the
+callsite that the returned `PrimExpr` must be a `BufferPointer`.
+
+Prior to this RFC's implementation, all transformations that modify a
+buffer must have near-equivalent mutators for both the `BufferLoad`
+and `BufferStore` nodes.
+
+# Prior art
+[prior-art]: #prior-art
+
+This follows a C-style convention.  Given an array `int x[100]`, the
+location `int* ptr = (x+50)` represents the location of element 50 in
+array `x`.  A buffer load is then represented as `*ptr`, and buffer
+store is represented as `*ptr = val;`.
+
+This also maps onto Vulkan semantics, where a `OpAccessChain`
+instruction is used to generate a pointer into an array, which can
+then be used with either `OpLoad` or `OpStore`.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+The `BufferPointer` could represent a pointer to a specific element in
+the generated C code.  This can be used for generating pointers to
+pass into hardware-specific intrinsics, rather than using
+`BufferNode::elem_offset` or the built-in `tvm_access_ptr`.


spectrometerHBH:
If we explicitly introduce a concept similar to `pointer`, there may be some related problems.

- If we don't have corresponding language construct to independently define a pointer and use it, like
```python
p = BufferPointer(A, [i, j], value_dtype="f32")
```
then BufferPointer is merely an IR data structure that attaches to BufferLoad/BufferStore, which benefits the developer but doesn't bring additional expressiveness for the user.

-  If we introduce such definition statements of pointer variables, then a latent problem is alias, though I can't immediately mock up an example to clearly show which effects this will bring to the IR transformation analysis.

rfcs/0042-buffer-pointer.md
```diff
@@ -0,0 +1,116 @@
+- Feature Name: (fill me in with a unique identifier, `my_awesome_feature`)
```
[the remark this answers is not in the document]
Lunderberg: Thank you, updating the name along with addressing the other points.

Lunderberg:
> BufferPointer is somehow like the combination of Buffer and indices. It's good if we can have it, but my question is if it's necessary? As we are changing the core data structure, which is a breaking change, the cost is greater than the benefits in my opinion.

Thank you, and that was the main concern that I had.  This change would make the areas that I've been working on lately be much cleaner, but outside of that .

The breakage was minimized by keeping the existing `BufferLoad`/`BufferStore` constructors, while adding an overload that accepts `BufferPointer`, and the same could be added on the python side.  That would improve the source backwards compatibility, though it would still break binary backwards compatibility.

> BTW, Could you please show some more design details? such as the data structure and the AST with BufferPointer? So that more people can join the discussion.

Certainly.  I won't be able to add them here until Monday, but I have a sample implementation at [PR#9391](https://github.com/apache/tvm/pull/9391).

> If we don't have corresponding language construct to independently define a pointer and use it, ... , then BufferPointer is merely an IR data structure that attaches to BufferLoad/BufferStore, which benefits the developer but doesn't bring additional expressiveness for the user.

The current implementation does expose it, though primarily for testing purposes.  At the moment, none of the code-generators handle it explicitly, so the only current use is with BufferLoad/BufferStore.  It might be able to replace the builtin `tvm_access_ptr`, though the ease for developers was my main thought on it.

> If we introduce such definition statements of pointer variables, then a latent problem is alias, though I can't immediately mock up an example to clearly show which effects this will bring to the IR transformation analysis.

Agreed.  I kept to having the same buffer/indices as `BufferLoad`/`BufferStore` had, so that at least it wouldn't increase the risk.  I wouldn't want to have pointer arithmetic allowed, since that would make analysis to avoid aliasing or out-of-bounds access be harder.

vinx13:
I agree with @spectrometerHBH that changing the IR design may introduce problems. To reuse the code but BufferLoad/Store, we can introduce auxiliary functions or construct auxiliary data structure on the fly. The `BufferPointer` now is `PrimExpr`, we must be careful it doesn't appear in other places. We intentionally keep things restricted so that buffer do not alias and not addressable(e.g. no concrete ptr). During rewriting, we don't deal with pointers explicitly, I'd prefer keeping the concept of pointer abstract, to prevent potential issue of aliasing.

rfcs/0042-buffer-pointer.md
```diff
@@ -0,0 +1,116 @@
...
+so transformations that must distinguish between reads and writes
+become more verbose.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The `BufferLoad::pointer` and `BufferStore::pointer` could be generic
+`PrimExpr`, instead of being `BufferPointer` objects.  This would
+require the datatype to be a handle, with an additional parameter to
+indicate what is being stored.  However, this
```
kparzysz-quic, line 80: There is a truncated sentence at the end of the paragraph.

Lunderberg: Thank you for that catch, and updated.  I tend to bounce between different sentences when editing, and I thought I had gone through it to catch all of these, but it looks like I missed one.

Lunderberg:
Thank you for the feedback.  Summarizing, it seems the major issues are as follows.  I have some follow-up questions for the different points, listed in the sub-bullets.

* Calling the new data structure `BufferPointer` brings in lots of C assumptions of pointer handling, which could either confuse users if those assumptions aren't met, or would significantly complicate compile-time analysis if those assumptions are met.
  * Would renaming it to `BufferAccess` address these concerns, per @vinx13's [suggestion here](https://github.com/apache/tvm/pull/9391#pullrequestreview-794808362), by avoiding the comparison to C-style pointers?

* Having the new data structure subclass from `PrimExpr` means that it could be used outside of just `BufferLoad` or `BufferStore`, and there's no indication at the site of the `StmtExprMutator` subclass that it should only return a specific data structure.
   * What if we add another functor, similar to the existing `ExprMutator` and `StmtMutator`, which visits objects whose mutated type must be identical to the original type?  That is, the method signature would be `BufferAccess VisitObj_(BufferAccessNode* node)`.
   * This would also be applicable to other cases that aren't currently handled, such as the `Buffer` object itself.  As an example, currently a buffer's shape may be defined in terms of a `tir::Var`.  However, `tvm::tir::Substitute` doesn't touch the buffer object, and so that variable in the buffer's shape doesn't get replaced.

* Any change to TIR is a breaking change, and therefore should only be done when expanding capabilities.  Ease of development alone isn't sufficient justification for changing the TIR.
  * For my own understanding, what level of backwards compatibility does TVM aim for?  I wasn't able to find specific documentation on it.  The current implementation wouldn't satisfy any of the conditions below, but could with small modifications satisfy condition (a).  Condition (b) would be harder to meet, mostly due to the direct access of TIR node member variables, but might be possible with dummy classes and implicit type conversions.
    1. Source compatibility with external Python, but external C++ code may require updates to remain compatible.
    2. Source compatibility with external Python and C++, but external C++ code may require recompilation.
    3. Binary compatibility with external code, with new versions of `libtvm.so` usable as a drop-in replacement without recompilation.
    4. Bug-for-bug compatibility with previous versions.
  * This RFC was intended to make the proposed [RFC#0039](https://github.com/apache/tvm-rfcs/pull/39), which among other changes extends the use of `BufferLoad`/`BufferStore` and deprecates `Load`/`Store`, be simpler to implement.  Would this TIR change be more acceptable as part of a larger change, adding functionality and impacting similar TIR nodes?

Lunderberg:
From @vinx13's [suggestion on the TVM PR thread](https://github.com/apache/tvm/pull/9391#pullrequestreview-794808362), quoting here for the discussion.

> Given it is common that only buffer + indices matters in rewriting, I'd suggest refactor into `RewriteBufferAccess(pair<buffer, indices>)` without modifying IR itself wherever possible. We can use a pair or auxiliary data structure `BufferAccess` (avoid potential confusion with concrete buffer pointer) to hold the input/output.  We can discuss more in the RFC

My concern there is for newcomers coming to the codebase.  Utility functions are effective if somebody knows that the utility functions exist, but require that knowledge first.  Making the preferred method to handle a specific task be the default way to do it, without requiring the utility function, is easier for somebody to find.  That said, I agree that a utility function would be a good fallback if we don't want to change the TIR.

vinx13:
To be clear, we should discuss the tradeoff here.

T0: On one hand, for analysis/rewriting that do not need to differentiate read/write relation, having a common BufferAccess class is helpful. 
T1: On the other hand, for analysis/rewriting that do need read/write and other relation, having an explicit BufferAccess can results in methods that forget to consider r/w relations(because of the common BufferAccess class), or other relations that follow.

So from a developer experience pov, it is hard to argue which way is clearly better. T0 would call for the introduction of BufferAccess, while T1 would favor the original AST where read/write are explicit.  Given most goals of T0 can be implemented via a common util function. Considering all the factors, the clarity of read/write relation(in T1) and simplicity of AST outweights the benefits in T0.

IR change is certainly important even if it is just a developer win, as long as it is strictly better than the old design,  but any IR change would require some deliberation since there usually will be tradeoffs like T0 and T1(also the more serious problem bought by BufferPointer, which can be addressed if BufferAcccess is just an Object) as a result can take longer time.

Lunderberg:
Good point on the tradeoffs, and I agree that this turned out to be a longer discussion that should be separate from the changes in RFC0039.

For the T1 developer experience, I've been assuming that they would still be written using visitors of the `BufferLoad` and `BufferStore` objects, and so the main difference would be accessing `op->access->buffer` instead of `op->buffer`.

Lunderberg:
Following a conversation with @vinx13, below are the overall points of discussions.

- The improvement to writability/readability aren't as universal as I had previously thought.  It improves some use cases, is neutral to others, but there are some transforms that are made somewhat worse by using `BufferAccess`. Examples that we found of each are shown below.
  
  - `PrimFuncSpecializer` is made much simpler by using `BufferAccess`.  This is generally the case for transforms that replace buffers/indices.
  
    <details>
      <summary>Current implementation (click to expand)</summary>
      
      ```c++
      Stmt VisitStmt_(const BufferStoreNode* op) final {
        Stmt stmt = StmtExprMutator::VisitStmt_(op);
        op = stmt.as<BufferStoreNode>();
        ICHECK(op != nullptr);
        auto it = buffer_map_.find(op->buffer);
        if (it == buffer_map_.end()) {
          return GetRef<BufferStore>(op);
        } else {
          auto n = CopyOnWrite(op);
          n->buffer = it->second;
          return Stmt(n);
        }
      }
      
      PrimExpr VisitExpr_(const BufferLoadNode* op) final {
        PrimExpr expr = StmtExprMutator::VisitExpr_(op);
        op = expr.as<BufferLoadNode>();
        ICHECK(op != nullptr);
        auto it = buffer_map_.find(op->buffer);
        if (it == buffer_map_.end()) {
          return GetRef<BufferLoad>(op);
        } else {
          auto n = make_object<BufferLoadNode>(*op);
          n->buffer = it->second;
          return PrimExpr(n);
        }
      }
      ```
    </details>
  
    <details>
      <summary>Cleaner implementation, current TIR with helper function (click to expand)</summary>
      
      ```c++
      Stmt VisitStmt_(const BufferStoreNode* op) final {
        BufferStore store = Downcast<BufferStore>(StmtExprMutator::VisitStmt_(op));
      
        auto writer = store.CopyOnWrite();
        ModifyBuffer(writer->buffer);
      
        return std::move(store);
      }
      
      PrimExpr VisitStmt_(const BufferLoadNode* op) final {
        BufferLoad load = Downcast<BufferLoad>(StmtExprMutator::VisitExpr_(op));
      
        auto writer = load.CopyOnWrite();
        ModifyBuffer(writer->buffer);
      
        return std::move(load);
      }
      
      void ModifyBuffer(Buffer& buffer) {
        auto it = buffer_map_.find(buffer);
        if (it != buffer_map_.end()) {
          buffer = it->second;
        }
      }
      ```
    </details>
    
    <details>
      <summary>Cleanest implementation, uses `BufferAccess` (click to expand)</summary>
      
      ```c++
      BufferAccess VisitExpr_(const BufferAccessNode* op) final {
        BufferAccess access = Downcast<BufferAccess>(StmtExprMutator::VisitExpr_(op));
      
        auto it = buffer_map_.find(op->buffer);
        if (it != buffer_map_.end()) {
          auto writer = access.CopyOnWrite();
          writer->buffer = it->second;
        }
      
        return Downcast<PrimExpr>(access);
      }
      ```
    </details>
  
  - `ComputeInliner` is largely unchanged by using `BufferAccess`.  This is largely the case for transforms that read the buffers/indices, then perform some replacement to remove the buffer use.
  
    <details>
      <summary> Current implementation (click to expand)</summary>
      
      ```c++
      PrimExpr VisitExpr_(const BufferLoadNode* _load) final {
        BufferLoad load = Downcast<BufferLoad>(StmtExprMutator::VisitExpr_(_load));
        if (!load->buffer.same_as(inlined_buffer_)) {
          return std::move(load);
        }
        return ReplaceInlinedBuffer(std::move(load));
      }
      
      PrimExpr ReplaceInlinedBuffer(BufferLoad load) {
        SetIndexSubstitution(load->indices);
        return Substitute(inlined_store_->value, idx_sub_);
      }
      ```
    </details>
    
    <details>
      <summary> Implementation with `BufferAccess` (click to expand)</summary>
      
      ```c++
      PrimExpr VisitExpr_(const BufferLoadNode* _load) final {
        BufferLoad load = Downcast<BufferLoad>(StmtExprMutator::VisitExpr_(_load));
        if (!load->access->buffer.same_as(inlined_buffer_)) {
          return std::move(load);
        }
        return ReplaceInlinedBuffer(std::move(load));
      }
      
      PrimExpr ReplaceInlinedBuffer(BufferLoad load) {
        SetIndexSubstitution(load->access->indices);
        return Substitute(inlined_store_->value, idx_sub_);
      }
      ```
    </details>
    
  - `CacheReadRewriter` is made worse by use of `BufferAccess`.  This is largely the case when a transform interacts with only loads, or only stores.
    
    <details>
      <summary> Current implementation (click to expand)</summary>
      
      ```c++
      PrimExpr VisitExpr_(const BufferLoadNode* load) final {
        if (load->buffer.same_as(info_->read_buffer)) {
          ObjectPtr<BufferLoadNode> n = make_object<BufferLoadNode>(*load);
          n->buffer = info_->write_buffer;
          return PrimExpr(n);
        }
        return ExprMutator::VisitExpr_(load);
      }
      ```
    </details>
    
    <details>
      <summary> Visiting `BufferAccess` (click to expand)</summary>
      
      ```c++
      // Need to make sure that we only rewrite access that occurs within
      // BufferLoad, which was implicit in the current implementation from
      // only visiting BufferLoad.
      
      Stmt VisitStmt_(const BufferStoreNode* op) final {
        ICHECK(!op->access->buffer.same_as(info_->read_buffer))
            << "Unexpected write to cache_read buffer " << op->access->buffer->name;
        return StmtExprMutator::VisitStmt_(op);
      }
      
      PrimExpr VisitExpr_(const BufferAccessNode* op) final {
        auto access = Downcast<BufferAccess>(StmtExprMutator::VisitExpr_(op));
        if (access->buffer.same_as(info_->read_buffer)) {
          auto writer = access.CopyOnWrite();
          writer->buffer = info_->write_buffer;
        }
        return std::move(access);
      }
      ```
    </details>
    
    <details>
      <summary> Rewrite `BufferAccess` while visiting `BufferLoad` (click to expand)</summary>
      
      ```c++
      PrimExpr VisitExpr_(const BufferLoadNode* op) final {
        auto load = Downcast<BufferLoad>(StmtExprMutator::VisitExpr_(op));
        
        if (load->access->buffer.same_as(info_->read_buffer)) {
          // Need to rewrite the `BufferAccess` before we can rewrite the `BufferLoad`.
          BufferAccess access = load->access;
          auto access_writer = access.CopyOnWrite();
          access_writer->buffer = info_->write_buffer;
  
          auto load_writer = load.CopyOnWrite();
          load_writer->access = access;
        }
        return std::move(load);
      }
      ```
    </details>
  
  
- The `BufferAccess` object would have the potential to replace the `tvm_access_ptr`, which is used to generate a pointer into a buffer, which can then be passed to hardware intrinsic functions.

  - Pros
    - Layout transformations that occur on graphs including a `tvm_access_ptr` do not need special handling.
    - Could replace the built-in call with a TIR node, which would be easier to access from other transformations.
      
  - Cons
    - TIR graphs that use `tvm_access_ptr` must conform to the layout that is used by the hardware intrinsic, and probably shouldn't have layout transformations applied to it.
    - Would introduce pointers as a first-class concept in TIR.


Overall, our conclusion was that since the ergonomics of `BufferAccess` can make some types of transforms worse to read, there isn't a strong argument to add it in.  The [specific commit](https://github.com/apache/tvm/pull/9391/commits/6ac92751137bd73e4483b04ee486e79ef62a9322) that simplified transforms that could be simplified with `BufferAccess` identifies passes where a helper function could simplify the pass itself.  We'd like to wait a few weeks to see additional arguments in favor of introducing `BufferAccess` come up, but otherwise would break the tie in favor of not changing the existing TIR.

Lunderberg:
A slight modification to the example of using a helper function to simplify implementation of the visitors for `BufferStore` and `BufferLoad`.  Use of `CopyOnWrite` to pass a non-const reference to the helper function would cause unnecessary copies in the case where no change is necessary.  The copy in `CopyOnWrite` occurs when `CopyOnWrite()` is called, not when the write itself happens.  Therefore, the helper function should be rewritten in order to avoid this copy.  This doesn't change the conclusion, but would change the recommended way to write the helper function.

<details>
    <summary>BufferLoad/BufferStore, using templated helper function</summary>
    
```c++
Stmt VisitStmt_(const BufferStoreNode* op) final {
  return ModifyBuffer(Downcast<BufferStore>(StmtExprMutator::VisitStmt_(op)));
}

PrimExpr VisitStmt_(const BufferLoadNode* op) final {
  return ModifyBuffer(Downcast<BufferLoad>(StmtExprMutator::VisitExpr_(op)));
}

template<typename Node>
Node ModifyBuffer(Node node) {
  auto it = buffer_map_.find(node->buffer);
  if (it != buffer_map_.end()) {
    auto writer = node.CopyOnWrite();
    writer->buffer = it->second;
  }

  return node;
}
```
</details>

(I think the original helper function could still work if `obj.CopyOnWrite()->field` returned a proxy object that was convertible to the underlying type, and whose `operator=` performed the copy itself.  However, that's probably more trouble/magic than it's worth.)

areusch:
cc @manupa-arm

vinx13:
If there are no further comments, shall we conclude this?

Lunderberg:
Agreed, closing both this and the [implementation PR](https://github.com/apache/tvm/pull/9391).

PR #45, opened 2021-12-01
[RELAY][AST] Add virtual device as a first class field to Relay expressions

electriclilies:
This is a proposal to add virtual devices as a first class field to Relay expressions.

cc @mbs-octoml @mikepapadim @jroesch


rfcs/1111-first-class-virtual-device.md
```diff
@@ -0,0 +1,146 @@
...
+  TVM_DECLARE_BASE_OBJECT_INFO(RelayExprNode, BaseExprNode);
+};
+```
+
+Additionally, I will add virtual_device_ to the WithFields methods.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One challenge with making virtual devices first class is that passes in TVM do not propagate all fields when they visit expressions. You can see this today with spans-- most visitors do not preserve spans. When we introduce the virtual device field, we will need to ensure that it is propagated correctly throughout the Relay program. To do this, I introduced WithFields, a COW constructor that copies extra fields, including spans. I will extend WithFields to also copy virtual devices. Then, we can use WithFields to ensure that the virtual device field is correctly propagated.
```
denise-k, line 118: @electriclilies could you consider adding either a function descriptor or a link to the [code](https://github.com/apache/tvm/blob/main/src/relay/ir/expr.cc#L79-L99) for `WithFields`? The description is useful but it is easier to understand with either of those extra pieces of info.

electriclilies: Yes, will do

tqchen:
cc @comaniac @zhiics @ZihengJiang

electriclilies:
Here's my proposed implementation: https://github.com/apache/tvm/pull/9641/files

mbs-octoml:
Thanks Lily, LGTM.

Let me apologize for getting us into the current on_device + function attributes mess your proposal gets us out of. It seemed like a workable approach at the time which would not require both an AST change *and* careful treatment of AST node copies. But in retrospect the on_device + function attributes approach has caused at least four bugs due to the fragility of maintaining its annotation invariant.

I'm hoping the experience we gain from adding virtual_device_ will inform improvements to checked_type_. In particular getting us out of the need to rerun InferType after every xform (and sometimes in strict succession). But one step at a time.

rfcs/1111-first-class-virtual-device.md
```diff
@@ -0,0 +1,146 @@
...
+  TVM_DECLARE_BASE_OBJECT_INFO(RelayExprNode, BaseExprNode);
+};
+```
+
+Additionally, I will add virtual_device_ to the WithFields methods.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One challenge with making virtual devices first class is that passes in TVM do not propagate all fields when they visit expressions. You can see this today with spans-- most visitors do not preserve spans. When we introduce the virtual device field, we will need to ensure that it is propagated correctly throughout the Relay program. To do this, I introduced WithFields (code: https://github.com/apache/tvm/blob/main/src/relay/ir/expr.cc#L79-L99), a COW constructor that copies extra fields, including spans.  I will extend WithFields to also copy virtual devices. Then, we can use WithFields to ensure that the virtual device field is correctly propagated.
```
comaniac, line 118: Sorry that I didn't follow `WithFields` before but this is indeed a super useful feature exactly due to the reason you pointed out. Could you clarify that whether we need to change all Relay passes to use `WithFields` to preserve the fields, or we could have an approach to make it effective to existing passes without changing them?

rfcs/1111-first-class-virtual-device.md
```diff
@@ -0,0 +1,146 @@
...
+
+Additionally, passes that move or introduce expressions will need to cooperate with device annotation (i.e., figure out the correct device themselves and insert it in any expressions they create), or device planning will need to be run again after the pass (similar to the type inference pass). In the future, we'd like to introduce a 'lite' version of the device planning pass that "fills in the gaps" left by these passes. It will propagate virtual device information strictly upwards and will have much less overhead than the full device planning pass.
```
comaniac, line 120: IIUC, after this RFC `virtual_device_` behaves almost the same as `checked_type_`, so this requirement makes sense to me. Meanwhile, I'm not sure how frequent we should run this new pass to preserve/propagate `virtual_device_`. If we have to run it as frequent as `InferType`, one thought I'm having is whether it is reasonable to treat first class `virtual_device_` as a part of type, and directly implement the virtual device propagation logic in `InferType`; otherwise we may need to run much more passes which may further slow down the compilation.

rfcs/1111-first-class-virtual-device.md
```diff
@@ -0,0 +1,146 @@
...
+[unresolved-questions]: #unresolved-questions
+
+The name of the virtual device is up for discussion. We've considered renaming SEScope to VirtualDevice, and it seems like that is the name we will eventually go with. If we go with a different name for SEScope, the name of virtual_device_ may be different.
+
+I also need to confirm that it is feasible to propagate all spans and virtual devices through most passes. I will do this by picking a pass and making sure I can get it "span clean", meaning that it preserves all spans and propagates them correctly.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+I'll also need to change the machinery of the device planner after this field is introduced. The basic idea is that we won't have any on_device ops after device planning anymore, all the virtual device information will be stored directly in the virtual_device_ field. I won't go into much detail about this here for the sake of limiting the scope of this discussion. Also, we will probably introduce a 'lite' version of device planning that can "fill in the gaps" in device annotations that may be left by some passes.
```
comaniac, line 146: It's also worthwhile to think about how to leverage this mechanism to simplify the BYOC flow. Currently we use `compiler_begin` and `compiler_end` to annotate the graph and face two major challenges:
1. The graph size will grow to at most 3x in the case that every op is annotated as an offloadable subgraph.
2. It's non-trivial to manipulate the compiler regions.

Ideally if we could simply annotate each call node with `virtual_device_`, we don't need `compiler_begin` and `compiler_end` anymore, but at this moment it's not clear to me about how to represent a region (subgraph) of ops on the same device. Maybe a compromise approach is we leverage `virtual_device_` in AnnotateTarget, and insert `compiler_begin/end` accordingly in MergeCompilerRegion, but this topic seems out of scope of this RFC.

rfcs/1111-first-class-virtual-device.md
electriclilies: Thanks the feedback @comaniac! I outlined a few other options we considered in the rationales and alternatives section. I'll expand a bit on what I said there. 

The current approach device planning takes is to store the virtual device (SEScope) information in the Relay program in the on_device nodes. Passes will always preserve these nodes and provided that passes don't significantly change the lexical scope, they won't cause problems. 

If we store the virtual device information directly on nodes, we will need to preserve it, either manually or through the WithFields method. In general passes *should* be doing this anyways-- they shouldn't be throwing away information willy-nilly. In reality they don't, which why spans are pretty much useless right now. 

We need to balance adding overhead to passes with maintaining a global invariant. Based on our experience with the current implementation, we think that the overhead is necessary to maintain the invariant.

Fortunately, the change is only difficult because we have to do it in so many places, but it will be a big PITA, which is why @mbs-octoml used on_device to represent this information in the first place. You can take a look at this PR for an example of what the changes would look like: https://github.com/apache/tvm/pull/9533

rfcs/1111-first-class-virtual-device.md
electriclilies: Currently the domains of types and virtual devices are orthogonal-- for example, tuples can only have one virtual device but each field of a tuple has a type. There's more context on the connection between types and virtual devices in device_domains.h and in device_planner.cc if you are curious. 

@mbs-octoml could you expand on this point a bit?

rfcs/1111-first-class-virtual-device.md
comaniac: Fair enough to me about the concern of invariant. I guess we may not need to propagate virtual device after each pass but just need to do it before a pass that requires such information, as long as they are available at the node level (i.e., `WithField` is used everywhere).

I totally agree with you that `WithField` should be used in all passes. I'm just thinking how could we more or less enforce it (e.g., fail to compile TVM otherwise or something like that). Otherwise it's hard to make sure all pass implementations in the future would follow this practice.

rfcs/1111-first-class-virtual-device.md
electriclilies: @mbs-octoml Could you take this one? Thanks!

rfcs/1111-first-class-virtual-device.md
mbs-octoml: Hi Cody, thanks so much for the comments. In my travels with PlanDevices I've seen four flavors of passes.
 a) Those that just want to know the device for every sub-expression on demand (eg manifest memory). Currently those passes need the help of the DeviceAware* visitors, but with this change the device will be right at hand.
 b) Those that rewrite expressions fairly deeply and should be fully device aware ones (eg lambda lifting). Currently those accumulate and re-encode devices with the existing on_device+function attributes convention. With this change they'll do the same, but the bookkeeping should be very much simpler.
 c) Those that can be completely ignorant of devices and just do local rewriting (eg a pattern based local optimization). The before and after expressions should remain on the same device. Currently the lexical scoping rules for on_device annotations lets us avoid doing anything special for these passes -- we don't even have to worry about re-running PlanDevices. But with this change we'll not be able to get away with that anymore and we'll need to rerun PlanDevices (hopefully in an efficient form) to 'fill in the missing devices'.
 d) Passes which may move expressions around (eg dead code elimination when inlining is on, constant folding), but those expressions just need to bring their devices along with them. Currently these passes have proven to be quite fragile and the source of some nasty bugs. With this change I think they will become much more robust.

So this change is intended to help with d), but in return we need to deal with re-running PlanDevices for c).

I think passes which respectfully decline to use WithFields are just a special case of c). I think (but I'm prepared to be surprised) PlanDevices should have an 'erasure' property: ie
 PlanDevices(EraseAllVirtualDevices(PlanDevices(x)) = PlanDevices(x)
So those passes will incur the expense of reconstructing devices from scratch until they help out by switching to WithFields, but otherwise are semantically sound. Ie WithFields can be 'opt in' rather than required.

Anyway that was my rough thinking, but I remain a bit worried about the 'erasure' property so perhaps more passes than we like will need to be moved out of c) into b) or d).

rfcs/1111-first-class-virtual-device.md
mbs-octoml: Yeah, that's an excellent question Cody. Currently the 'device domain' used during PlanDevices is
  D ::=  <device> | fn(D, ..., D) -> D
So it's *slightly* type-dependent but otherwise very 'forgetful'. I think that's enough to justify keeping devices out of types, and their inference separate.

HOWEVER there are plenty of attributes which are refinements on types, and I suspect we'll want to put them in the type nodes directly. But even then I'd still not lump their analysis with type checking. That is, I'd still write the analysis as a stand alone pass, but the pass may make liberal use of induction on the already known types.

Would very  much like to keep an open mind on this one and use a concrete example to test the water. Maybe side effects. Maybe uniqueness (for in-place updates).

rfcs/1111-first-class-virtual-device.md
mbs-octoml: That was my thinking also Cody, and I started to lay it out in https://github.com/apache/tvm-rfcs/pull/38. That's somewhat moribund while we get more of the leg work in place (this change included), but would like to come back to it in Q1 with a focus on making BYOC more 'first class' (as well as address the duplication and efficiency concerns you mention.) Will be sure to ping you when that comes alive again, would very much welcome your help.

rfcs/1111-first-class-virtual-device.md
mbs-octoml: Oh, one more thing: For completeness we should mention there's a third approach (neither the current impl nor this RFC). PlanDevices inserts device_copy where needed, but otherwise doesn't bother trying to make it easy for downstream passes to recover the device for every sub-expression. Instead we have a helper visitor that just constructs that as an Map<Expr, SEScope>, and any passes that need it just call the helper. That's roughly how things used to work before I went down the current path, and maybe we'll need to go back to it should this RFC prove intractable. But I'm pretty confident we can do better and handle the c) style passes cheaply.

comaniac: Thanks for the detail analysis and explanation! I agree with your conclusions and feel conformable with the plan now. I'm now thinking as long as we could have unit tests running through the passes, we should be able to catch all missing information.

rfcs/1111-first-class-virtual-device.md
comaniac: Thanks you both for the clarification. Lily pointed out a good point that I didn't notice and Mark made it clearer. I agree that we could make it a standalone pass first and do more optimizations to shorten the compile time later if necessary. And definitely, side effects and in-place updates might bring many new issues in the future, but I believe this design should cover them in general :D

rfcs/1111-first-class-virtual-device.md
comaniac: Thanks for answering. Yeah I'd be happy to help on this at the time.

comaniac (commented):
LGTM. I don't have other comments at this moment. Gentle ping @jroesch @zhiics @ZihengJiang @junrushao1994  for review, or I'll merge this RFC tomorrow.

electriclilies:
@comaniac Thanks for reviewing! Looking forward to it going in :D

comaniac (approved):
LGTM

comaniac:
Thanks @electriclilies and all reviewers. This RFC is merged.

electriclilies:
Thanks cody!

PR #46, opened 2021-12-10
Module Based Model Runtime for AOT

areusch:
This adds the RFC for MBMR for AOT. See also discussion on the Discuss forum: https://discuss.tvm.apache.org/t/aot-module-based-model-runtime-interface-for-aot/11068

@tqchen @manupa-arm @jroesch @mehrdadh @csullivan @adstraw @Mousius @kparzysz-quic

### DIFF
--- a/rfcs/0046-module-based-model-runtime-for-aot.md
+++ b/rfcs/0046-module-based-model-runtime-for-aot.md
@@ -1,0 +1,348 @@
+# Module-based Model Runtime Interface for AOT
+
+- Feature Name: module_based_model_runtime_for_aot
+- Start Date: 2021-09-17
+- RFC PR: [apache/tvm-rfcs#0046](https://github.com/apache/tvm-rfcs/pull/0046)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# **Summary**
+
+This RFC describes a [Module-based Model Runtime
+interface](https://discuss.tvm.apache.org/t/discuss-module-based-model-runtime-interface/5025) for
+the [Ahead-of-Time Executor](https://discuss.tvm.apache.org/t/implementing-aot-in-tvm/9206), thereby
+enabling its use from the TVM C++ Runtime.
+
+# **Motivation**
+
+The microTVM project has made significant progress towards an Ahead-of-Time Executor for compiled
+Relay models. At the time of writing, it's now possible to codegen a TIR function which executes
+Relay models that have known shapes, don't have graph-level control flow, and execute only on the
+CPU device. Right now, the C runtime is the only such runtime environment which can interact with
+this generated code. However, significant interest exists in enabling the C++ runtime to use the
+Ahead-of-Time executor.
+
+# **Guide-level explanation**
+
+Users select the AOT executor at compile time through the traditional GraphExecutor compilation flow
+(e.g. `[tvm.relay.build](http://tvm.relay.build)`) by including `--executor=aot` in the Target
+[1]. The return value of `tvm.relay.build` in this case is an `AotExecutorFactory` Module
+object. Users instantiate the AOT executor via `AotExecutorFactory` as they do with `GraphExecutor`:
+
+```bash
+ir_mod = tvm.parser.fromtext("""\
+      #[version = "0.0.5"]
+      def @main(%a : Tensor[(1, 2), uint8], %b : Tensor[(1, 2), uint8]) {
+          %0 = %a + %b;
+          %0
+      }"""
+    )
+
+with PassConfig(opt_level=3):
+  factory : AotExecutorFactory = tvm.relay.build(
+       ir_mod, "llvm -executor=aot", module_name="my_mod")
+
+aot_executor : AotExecutor = factory["my_mod"](tvm.cpu(0))
+```
+
+`AotExecutor` supports the traditional Module-Based Model Runtime Interface and can be used as a
+user normally would `GraphExecutor`:
+
+```bash
+aot_executor.set_input("a", tvm.nd.array(np.ndarray([1, 2], dtype="uint8")))
+aot_executor.set_input("b", tvm.nd.array(np.ndarray([3, 5], dtype="uint8")))
+aot_exec.run()
+output = aot_exec.get_output(0)
+assert output.asnumpy() == np.ndarray([5, 7], dtype="uint8")
+```
+
+[1] NOTE: The target string is not the final place this customization should be made. However, it's
+been the place where we've been putting runtime-related stuff. A separate RFC will split the Target
+string into Target options (which affect tuning) and runtime options.
+
+# **Reference-level explanation**
+
+Already committed to TVM is the AotExecutorCodegen. This module produces a TIR top-level function
+which invokes the Relay operators (implemented in TIR) in a correct order. An example is given
+below:
+
+```bash
+PrimFunc([input1, input2, output]) attrs={"global_symbol": "tvmgen_my_mod_run_model", "runner_function": (bool)1} {
+  // attr [(nullptr)] device_id = 0
+  // attr [(nullptr)] device_type = 1
+  tir.tvm_call_packed("tvmgen_my_mod_fused_add", input1, input2, output)
+}
+```
+
+The AotExecutor then needs to accomplish the following to meet Module-based Model Runtime Interface:
+
+1. Allocate input and output tensors as defined in the `run_model` function using the correct Device
+   API.
+2. Provide a mapping from relay parameter name to positional argument.
+3. Invoke the generated TIR function and provide profiling.
+
+### Compiler ↔ Runtime Metadata
+
+In order to implement (1) and (2) above, additional metadata about the `run_model` function needs to
+be communicated from Compiler to Runtime:
+
+- The mapping between Relay parameter name and TIR argument position
+- The number of inputs and outputs
+- The type of each parameter
+- Information sufficient to choose a Device API to allocate memory for that data.
+
+At present, Metadata is passed from Compiler to Runtime in several different ways:
+
+1. Constant DLTensor can be bundled with code and supplied to `runtime::Module` via
+   `runtime::MetadataModule`
+2. Many non-DSO-exportable backends (`cuda`, `hexagon`, `metal`, `opencl`, `sdaccel`, `rocm`,
+   `vulkan`) have adopted the convention of including a
+   [1runtime::FunctionInfo`](https://github.com/apache/tvm/blob/main/src/runtime/meta_data.h#L106)
+   (NOTE: distinct from `tvm::relay::transform::FunctionInfo`) in their serialization:
+
+    ```bash
+    /*! \brief function information needed by device */
+    struct FunctionInfo {
+      std::string name;
+      std::vector<DLDataType> arg_types;
+      std::vector<std::string> launch_param_tags;
+    }
+    ```
+
+3. AotExecutorCodegen and GraphExecutorCodegen have adopted the practice of producing the
+   graph-level
+   [`runtime::MetadataNode`](https://github.com/apache/tvm/blob/main/src/runtime/meta_data.h#L55):
+
+    ```bash
+    /*!
+     * \brief Structure that can be optionally used by the executor codegen
+     */
+    class MetadataNode : public Object {
+     public:
+      /*! \brief input information for the main function */
+      Array<String> inputs;
+      /*! \brief number of outputs of the main function */
+      int num_outputs = 1;
+      /*! \brief the executor to be used to run the model */
+      String executor = kTvmExecutorGraph;
+
+      String mod_name = "";
+    }
+    ```
+
+4. The recent AOTExecutor implementation has created `tvm::relay::transform::FunctionInfo` which
+   communicates statistics about memory usage and I/O operation for each TIR operator and aggregate
+   statistics for the top-level AOT function:
+
+    ```bash
+    struct FunctionInfoNode : public Object {
+      Map<Target, Integer> workspace_sizes;
+      Map<Target, Integer> io_sizes;
+      Map<Target, Integer> constant_sizes;
+      Map<Target, tir::PrimFunc> tir_primfuncs;
+      Map<Target, Function> relay_primfuncs;
+    }
+    ```
+
+
+Some duplication of information is already present. Likely this is due in part to the existing
+middle-end compiler design, in which a separate `IRModule` is produced for each backend. Another
+factor may be: since `runtime::Module` are responsible for their own serialization, and passing
+`Node` across `PackedFunc` requires a cast, the lack of a centralized facility for
+`runtime::Modules` to obtain module-level Metadata has led backend authors to roll their own. This
+pattern means that it's very difficult to assess the full scope of metadata handed to the runtime,
+particularly across all backends.
+
+Work is currently ongoing to unify the pre-codegen `IRModule` into a single instance. After this
+work is completed, it will be much easier to produce a centralized module-level Metadata. This RFC
+argues for the expansion of `runtime::MetadataNode` in the following ways:
+
+1. Rename `runtime::MetadataModule` to `runtime::ConstLoaderModule` to disambiguate the two and make
+   its purpose in life clearer.
+2. Expand `input_args` in the existing `runtime::Metadata` to parity with `runtime::FunctionInfo`,
+   plus include `_sizes` from `tvm::relay::transform::FunctionInfoNode` and the required `shape` and
+   `dtype` information from the beginning of this section.
+3. Introduce `ModelMetadataModule` to contain this information for use with the C++ runtime.
+
+    ```bash
+    class ModelMetadataModule {
+      virtual GetFunction(const std::string& name, ObjectPtr<Object>& sptr_to_self) {
+        if (name == "get_model_metadata") {
+           return PackedFunc([](TVMArgs args, TVMRetValue* rv) {
+              *rv = ModelMetadata(metadata_);
+           });
+        } else {
+          return PackedFunc();
+        }
+      }
+
+      const struct ModelMetadata* metadata_;
+    };
+    ```
+
+4. Introduce an optional implementation for the C runtime.
+5. Export runtime::Metadata to Model Library Format.
+
+The new proposed definition of `runtime::Metadata` is as follows.  NOTE that this is a C definition
+because it will be made available both the C and C++ runtimes. A C++ wrapper will be written.
+
+```bash
+struct ParameterInfo {
+  const char* relay_name_hint;
+  const char* tir_name_hint;
+  int64_t* shape;
+  int64_t ndim;
+  DLDataType dtype;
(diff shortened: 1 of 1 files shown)


tqchen:
cc @comaniac @zhiics as it is related to MetaDataModule

manupa-arm:
Thanks! @areusch I ll try to do a pass soon.

Not sure its in scope for this work, but definitely something to consider:

w.r.t. MetaDataModule, if we are re-purposing it to be just ConstLoader, @ashutosh-arm has contributed pass that could kick constants back to the @main function. https://github.com/apache/tvm/blob/main/src/relay/backend/contrib/cmsisnn/extract_constants.cc

I think using that in the BYOC pipeline with -link-params should just work if I am not mistaken. Therefore, we would not need a Metadata module of type "metadata" that is used for constant storage.

areusch:
@manupa-arm thanks! I think ultimately the @ashutosh-arm pass could be a good way to implement ConstLoader, but for now I'm hoping we can simply rename. In the long run, I'd actually like to be able to do a similar thing with MetadataModule--define a special data type in Relay which holds this Metadata, and "link" tvm_struct_get calls against it just before codegen with an IRModule -> IRModule pass to replace such tvm_struct_get with the value from Metadata. 

I'm not sure what the best approach is here, probably need to collect some feedback from the community/users. However, at some level it all seems like metadata to me (constants could be params tho, and it may be desirable to be able to swap those params after compilation).

areusch:
@manupa-arm @comaniac @zhiics could you take a look here?

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+PrimFunc([input1, input2, output]) attrs={"global_symbol": "tvmgen_my_mod_run_model", "runner_function": (bool)1} {
+  // attr [(nullptr)] device_id = 0
+  // attr [(nullptr)] device_type = 1
+  tir.tvm_call_packed("tvmgen_my_mod_fused_add", input1, input2, output)
+}
+```
+
+The AotExecutor then needs to accomplish the following to meet Module-based Model Runtime Interface:
+
+1. Allocate input and output tensors as defined in the `run_model` function using the correct Device
```
manupa-arm, line 78: Currently AoT executor codegen does not do this. We might want to expand on this a bit.

First of all, if we are doing this -- this roughly translates into create tir.allocate nodes inside the main PrimFunc for input and output tensors.

In order to achieve this, we could do the following : 

E1) Make the main function copy the data passed into to the allocate nodes. If so, we need to perform that copy, maybe in "set_input" function ? 
E2) Expose the space allocated via tir.allocate node as a pointer to "set_input". Therefore, the user is expected either to populate the input directly there. This would avoid a copy. However, it does not align well with the current usecase of "set_input", because it is more like "get_input_space".

In this RFC, we need to make clear what is being proposed

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+    struct FunctionInfo {
+      std::string name;
+      std::vector<DLDataType> arg_types;
+      std::vector<std::string> launch_param_tags;
+    }
+    ```
+
+3. AotExecutorCodegen and GraphExecutorCodegen have adopted the practice of producing the
+   graph-level
+   [`runtime::MetadataNode`](https://github.com/apache/tvm/blob/main/src/runtime/meta_data.h#L55):
```
manupa-arm, line 113: FYI, I did move this to become ExecutorCodegenMetadata now as it did not feel like a runtime concept.

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+      Map<Target, Integer> io_sizes;
+      Map<Target, Integer> constant_sizes;
+      Map<Target, tir::PrimFunc> tir_primfuncs;
+      Map<Target, Function> relay_primfuncs;
+    }
+    ```
+
+
+Some duplication of information is already present. Likely this is due in part to the existing
+middle-end compiler design, in which a separate `IRModule` is produced for each backend. Another
```
manupa-arm, line 148: runtime.Module is produced per backend. As long as that exists, somewhere in the compiler per-backend IRModule have to exist -- that does the translation from IRModule --> runtime.Module.

However, we have hierarchical runtime.Module trees being built -- so what we can eventually do maybe create hierarchical IRModule structure. In that structure, it might make sense to keep this metadata.

cc : @tqchen @Mousius @jroesch

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+factor may be: since `runtime::Module` are responsible for their own serialization, and passing
+`Node` across `PackedFunc` requires a cast, the lack of a centralized facility for
```
manupa-arm, line 150: What is meant by Node here ?

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+`runtime::Modules` to obtain module-level Metadata has led backend authors to roll their own. This
+pattern means that it's very difficult to assess the full scope of metadata handed to the runtime,
+particularly across all backends.
+
+Work is currently ongoing to unify the pre-codegen `IRModule` into a single instance. After this
+work is completed, it will be much easier to produce a centralized module-level Metadata. This RFC
+argues for the expansion of `runtime::MetadataNode` in the following ways:
```
manupa-arm, line 157: I think the following arguments argues for re-structure (not just runtime::MetadataNode).

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. Rename `runtime::MetadataModule` to `runtime::ConstLoaderModule` to disambiguate the two and make
```
manupa-arm, line 159: Agreed for this RFC as discussed. I think we should be able to remove this in future.

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+   its purpose in life clearer.
+2. Expand `input_args` in the existing `runtime::Metadata` to parity with `runtime::FunctionInfo`,
```
manupa-arm, line 161: I believe you are referring to runtime::MetadataNode here. 

This is now : https://github.com/apache/tvm/blob/e6af87491eb250a3266b1f09b36055c6ee79146b/src/relay/backend/utils.h#L60-L84

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+3. Lookup and load any linked parameters using the `--link-params` mechanism.
+
+- `set_input`, `get_input`, `get_output` all work as they do in `GraphExecutor`.
+- `run` assembles `TVMArgs` containing inputs + outputs and invokes `tvmgen_<model_name>_run_model`.
+- `time_evaluator` is implemented in the same way as it is in `GraphExecutor`. Timing `run_model` is
+  done using the CPU timer.
+
+### Internal workings of AotExecutor (`--runtime=c --interface-api=packed`)
+
+The C runtime version works in a very similar way with C accessor functions for the `ModelMetadata`.
```
manupa-arm, line 232: Is it neccessary to change this runtime=c behaviour in this RFC ? 

Because this has implications that we have to undergo costs of allocating space for I/O tensors which is not desired always. We might still want to use packed API for dynamic shapes / symbolic codegen of PrimFuncs. I dont see a strong reason why runtime=C behaviour has to be coupled with Module-based Model Runtime's requirements just yet which comes with the mentioned overheads / inflexibility.

rfcs/0046-module-based-model-runtime-for-aot.md
```diff
@@ -0,0 +1,348 @@
...
+  return 0;
+}
+```
+
+Additional work is underway to wrap this in a firmware-friendly interface. A core design goal of
+this interface is to offload all memory management tasks to the calling code to facilitate
+integration with bare-metal embedded devices.
+
+Therefore, it would go against the goals of the C interface to introduce a generic runtime wrapper
+compatible with PackedFunc calling convention. It may be possible to do so in the future, but it
```
manupa-arm, line 254: I think its worth separating the concept of G1) PackedFunc callling convention vs  G2) Module-based Model runtime's requirement. G2 requires G1 -- however G1 does not always require G2. 

Here I would classify G2) as the need to use load a runtime.Module (dynamically) and use "set_input", "get_input", "get_output" interface.

manupa-arm (commented):
Sorry for the delay.

I did a detailed pass.

This is a great improvement. I have few suggestions for text changes and I do agree broadly with the approach with a minor comment about the need to change runtime=C , packed calling conv in this RFC/proposed work.

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: you mean that AOTExecutorCodegen does not emit information to determine the device, correct? I think right now it does in the sense that it hardcodes the device to (kDLCPU, 0) wherever it uses that, right?

here I'm not suggesting we modify the codegen to create tir.allocate nodes in the body of the emitted function. I'm only proposing that we emit metadata that contains shape, dtype, and device information for the expected inputs and outputs from that main func, and then defer allocation to either a runtime component (in the case of the C++ runtime) or a compile-time component (if used with the C runtime and microTVM).

The runtime changes proposed here align with E1. We could also implement `set_input_zero_copy` to align with E2. But again these are runtime changes which require that additional output from AOTExecutorCodegen to work properly.

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: ah updated the RFC text to point to that. thanks!

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: a couple points:
- more than one `runtime.Module` can be produced by a given backend
- depending on the load path in C++ runtime, the tree can become rearranged. see https://discuss.tvm.apache.org/t/introduce-artifact-a-container-for-generated-code/11099

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: `tvm::Node`; clarified

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: how so?

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: ah yeah. i think perhaps this is done now, then, since ExecutorCodegenMetadata exports tir::Var. However, to then fit this proposal we need to reduce it to something exportable. updated the text to clarify, see what you think.

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: i think if we are not changing the main function emitted by AotExecutorCodegen, this concern is resolved--but, please let me know if I understood it correctly. The main goal of this section is to sketch out how a possible implementation could work under these conditions, to ensure the metadata refactor is broadly applicable.

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: agreed. because this RFC proposes to implement a wrapper around the code-generated PackedFunc interface, I think these concerns are in fact separated. lmk if you disagree.

areusch (commented):
thanks @manupa-arm ! please take another look when you have a minute.

rfcs/0046-module-based-model-runtime-for-aot.md
manupa-arm: Agreed. 

The requirement to mention this here seems like stemming from the fact it is aligned with other RFC/pre-RFCs that single (tree of) IRModule --> (list of) runtime.Modules -- is a proposed change -- therefore, this motivates the change proposed here.

Thus, would it be possible to add a reference to this proposal (if any)?

rfcs/0046-module-based-model-runtime-for-aot.md
manupa-arm: I dont see how 1.) is relevant to the expansion of runtime::MetadataNode.

My comment was to adjust the text here to say something like "modify the lowering flow" rather than "the expansion of `runtime::MetadataNode`" -- because latter is a sub-part of the former and I believe former is what is proposed here and covered by the following points.

rfcs/0046-module-based-model-runtime-for-aot.md
manupa-arm: Ack.

rfcs/0046-module-based-model-runtime-for-aot.md
manupa-arm: Yea it was made clear to me in the above comments -- as mentioned above lets be explicit that this is a runtime wrapper.

manupa-arm (commented):
Hi Andrew,

Thanks for the clarifications approach LGTM.

The remaining comments are to add more clarifications, pointers for future work and arguments against alternative (if any).

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: ah i see, changed the wording. i guess i was trying to say that the name MetadataNode should cover a more complex thing, but in practice you're right that this proposes a restructuring.

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: > I suppose the proposal here is we create a runtime wrapper around it -- which also works but with cons of not exposing these allocations to the core compiler for further optimization.

I actually think we should defer the question of how the input tensors are loaded and consumed to the runtime. You're right that AOTExecutor here is intended to be a generic runtime wrapper when there are enough resources on the system to handle this at runtime. However, consider a microTVM use case where DMA is used to copy data from e.g. a camera into an SRAM dedicated to accelerator usage. In this case, TVM should really do as much as possible to stay out of the way--the copy operation is application- or at least SoC- specific. TVM should provide pointer and sizing information so this can be handled separately. That argues for exposing some type of `get_input_tensor` function as you're mentioning here as a first-class citizen of MBMR. I think we should take that up as we move on the C Device API and USMP integrations. 

> However, Im curious to know whether it would just easier to create a copy in the main body to tir.allocate node that get translated to a device copy -- which I think has the same effect.

I think I see what you're raising here--we have to explicitly inject the input and output tensors into USMP if they are not modeled by a `tir.allocate` somewhere. We could model them without specially injecting them by wrapping the TIR top-level function in a MBMR-compatible TIR function which contains `tir.allocate` nodes, IIUC. I'm not sure how we get away from needing to do this at present, though--for `DLDevice(kDLCPU, 0)`, no such copy is needed, but we would then still need the `tir.allocate` nodes to make the input tensors visible to USMP.

Additionally, I imagine in the double-buffered input use case, the user would need to somehow communicate that space for two input buffers is needed. Not sure how to do that now outside of downsizing a memory pool.

> I would propose being explicit here that this is a runtime wrapper.
Possibly, would it be possible to add a few notes that why/ why we should not do it the codegen of main ?

Added.

areusch (commented):
thanks @manupa-arm, clarified according to your comments here. PTAL!

rfcs/0046-module-based-model-runtime-for-aot.md
manupa-arm: > However, consider a microTVM use case where DMA is used to copy data from e.g. a camera into an SRAM dedicated to accelerator usage. In this case, TVM should really do as much as possible to stay out of the way--the copy operation is application- or at least SoC- specific.

> Additionally, I imagine in the double-buffered input use case, the user would need to somehow communicate that space for two input buffers is needed. Not sure how to do that now outside of downsizing a memory pool.

So it is a requirement of the accelerator that it can only access certain memories and the input tensor is not originally present in a such a memory. From an AoT perspective, ideally we should provide the compiler with this information -- because we cant generally know where the input is going to be.

Therefore, when the compiler knows whether the tensor needs copying, it could insert a copy -- that gets lowered to produce an intermediary allocate node -- to be in accelerator accessible memory. Once we have that, this should ideally translate down to a loop-nest of copy that could be scheduled via double_buffer() scheduling primitives either in TE/S-TIR. This will result in creating intermediary storage in the form of allocate nodes that could be planned properly. Additionally to that, depending on who does the copy (accelerator or the host) we need to place that -- now that copy insertion is IRModule --> IRModule transformation which I think a wrapper would not be sufficient --  I would assume copies might need ending up in the device functions.

Now, I agree that it out of scope / too much to consider for this work. In the absense of that, I think letting the set_input runtime function handling the copy make sense. It is just that the short-term solution goes way against the long term solution in my mind that sparked this conversation -- however, I agree with you that there is no easy way other than inserting the copies in the host PrimFunc of host,device PrimFunc tuple if we want to do it in a more codegen'd flow.

manupa-arm (commented):
Hi Andrew,

Thanks for the discussion, it was thought provoking :).
I have only a single comment asking for clarification about the lowering flow change that is brought by this RFC to attach model-level metadata to the tree of runtime.Modules.

Other than that this makes sense!

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: ah i see. yeah this makes sense. i'm wary of introducing too much complexity here particularly when user/platform intervention may be required to implement the double-buffer (e.g. if DMA is used to fill the buffer while the SoC is sleeping). it would be great to continue discussing this in a follow-on!

rfcs/0046-module-based-model-runtime-for-aot.md
areusch: I see--so this is more about the wording a couple paragraphs down, no?
> Work is currently ongoing to unify the pre-codegen IRModule into a single instance. After this work is completed, it will be much easier to produce a centralized module-level Metadata.

cc @jroesch i am actually not sure if there is an RFC describing this. i'm hoping to describe my ambitions to "link" metadata into TIR via `tir.load_metadata` node in a following RFC (and this is really what requires this consolidation).

areusch: per our offline discussion, clarified how the metadata is carried through the current compiler design, removed references to un-RFC'd design efforts and replaced with text to motivate them. also clarified some wording--ptal

manupa-arm (approved):
Thanks @areusch ! LGTM

manupa-arm:
Thanks @areusch . Looking forward for AoT being enabled in more places :).

TejashShah:
Thanks @manupa-arm for reviewing this RFC and providing crucial feedback in time-bound fashion.

PR #51, opened 2022-01-09
Bring `PackedFunc` into TVM Object System

cyx-6:
Bring `PackedFunc` into TVM Object System

### DIFF
--- a/rfcs/0051-PackedFunc-as-Object.md
+++ b/rfcs/0051-PackedFunc-as-Object.md
@@ -1,0 +1,82 @@
+- Feature Name: PackedFunc as Object
+- Start Date: 2022-01-01
+- RFC PR: https://github.com/apache/tvm-rfcs/pull/51/
+- GitHub Issue: TBD
+
+## 1. Summary
+
+This RFC allows developers to use `PackedFunc` as TVM objects, which completes the last missing step of TVM runtime object system.
+
+## 2. Motivation
+
+Historically, several fundamental data structures in TVM are not part of the runtime object system, namely `NDArray` (not object), `Module` (not object), `String` (not exist), `Array` (not in runtime), `Map` (not in runtime), `PackedFunc` (not yet an object).
+
+The rationale of the original design is mainly for simplicity, which is desirable for the usecases as a monolithic compiler. As time goes on, the community has come to realize the fact that the object system should be inclusive enough and by design allow more convenient integration with vendor libraries. Therefore, as part of the effort in TVM refactoring and TVM Unity, recent work strives to re-implement these core data structures to be consistent with the runtime object protocol with stable ABI guarantee, and thus could be passed across the DLL boundary.
+
+As the central piece of the TVM ecosystem, this proposal focuses on making `PackedFunc` a TVM object. By doing so, it completes the last missing piece of the object ecosystem, allows TVM containers to carry `PackedFunc`s, and enables `PackedFunc`s to be passed across the DLL boundary to bring convenience to the vendor library integration.
+
+
+## 3. Guide-level introduction
+
+This is mainly a developer-facing feature, and thus there is no sensible change to the existing functionalities to the end users, who are still supposed to use the same `PackedFunc` API.
+
+Only one major object is introduced, `PackedFuncObj`, a TVM object in the runtime system (detailed in the next section) which is an ABI stable data structure for packed functions that could be shared across language and DLL boundary.
+
+To avoid API misuse from developers, the `PackedFuncObj` cannot be created or manipulated directly, and the specialization of its creation `make_object<PackedFuncObj>` will be deleted for safety. Instead, the developer-facing class `PackedFunc` remains responsible for creating and managing the object, and for properly setting its content.
+
+In the future, it’s possible to incrementally add more information into `PackedFuncObj` to better help debugging and error reporting.
+
+Note: This RFC doesn’t change any of the existing functionality, including C ABI or `PackedFunc`’s C++ API. Any modification to the C ABI is out of scope of this RFC.
+
+## 4. Reference-level introduction
+
+As introduced below, the RFC introduces a new class:
+
+```C++
+class PackedFuncObj : public runtime::Object {
+  using FCall = void(const PackedFuncObj*, TVMArgs, TVMRetValue*);
+  FCall* f_call_;
+};
+
+```
+
+A templated subclass is introduces to do the type-erasing trick:
+
+```C++
+template <typename TCallable>
+class PackedFuncSubObj : public PackedFuncObj {
+  TCallable callable_;
+};
+```
+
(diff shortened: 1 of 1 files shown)


junrushao1994:
CC: @yzhliu @yzh119 @comaniac

junrushao1994:
CC @wweic

rfcs/0051-PackedFunc-as-Object.md
```diff
@@ -0,0 +1,82 @@
...
+
+Just like every change to the runtime, the proposed change could slightly affect runtime’s binary size. The effect, depending on the compiler, could be positive or negative.
+
+Overall, given that it brings significantly better experience as stated in the previous sections, we believe the benefits outweighs the potential drawback.
+
+## 6. Rationale and alternatives
+
+This refactoring is the last missing piece of effort that brings core data structures of the TVM runtime into the ABI-stable TVM runtime.
+
+Alternatively, one might argue that it’s not important whether `PackedFunc` should be a TVM or not; however, it significantly brings negative impact when TVM object system is used across the DLL boundary, or putting `PackedFunc` into TVM containers.
```
zxybazh, line 66:
```suggestion
Alternatively, one might argue that it’s not important whether `PackedFunc` should be a TVM object or not; however, it significantly brings negative impact when TVM object system is used across the DLL boundary, or putting `PackedFunc` into TVM containers.
```

rfcs/0051-PackedFunc-as-Object.md
```diff
@@ -0,0 +1,82 @@
...
+As the central piece of the TVM ecosystem, this proposal focuses on making `PackedFunc` a TVM object. By doing so, it completes the last missing piece of the object ecosystem, allows TVM containers to carry `PackedFunc`s, and enables `PackedFunc`s to be passed across the DLL boundary to bring convenience to the vendor library integration.
+
+
+## 3. Guide-level introduction
+
+This is mainly a developer-facing feature, and thus there is no sensible change to the existing functionalities to the end users, who are still supposed to use the same `PackedFunc` API.
+
+Only one major object is introduced, `PackedFuncObj`, a TVM object in the runtime system (detailed in the next section) which is an ABI stable data structure for packed functions that could be shared across language and DLL boundary.
+
+To avoid API misuse from developers, the `PackedFuncObj` cannot be created or manipulated directly, and the specialization of its creation `make_object<PackedFuncObj>` will be deleted for safety. Instead, the developer-facing class `PackedFunc` remains responsible for creating and managing the object, and for properly setting its content.
```
zxybazh, line 25: Can you elaborate a bit about the risk of enabling `make_object<PackedFuncObj>`?

cyx-6: Sure. Since the `PackedFuncObj` is an abstract object only, it is not allowed to be created directly, while its derived classes are allowed to do so.

areusch:
@cyx-6 thanks for raising this formal RFC. i chatted with @junrushao1994 and he reminded me that for PackedFunc registered directly with the C++ runtime, there is a fast path in calling them from C++ that avoid re-encoding the arguments into the documented PackedFunc C ABI. using this fast path in the context of a single project makes sense; however, as you point out here in "Unresolved questions," it also means that we may enshrine a de-facto ABI between TVM and other projects which is considerably more complex. 

i'm okay with creating this ABI if it provides a clear benefit over the already-defined one. if it doesn't, i would prefer that we do something different, such as extracting `packed_func.h` into a separate e.g. `dlpack` library and allow you to use the C++ PackedFunc wrappers in your code directly (and call into TVM code via the C PackedFunc ABI). 

it would be great to motivate creating this second ABI with some details about why you can't use the already-defined PackedFunc ABI. performance is a perfectly-good reason, but the RFC says nothing about that. would you mind adding this detail so it is clear why we are doing this? i don't want to create a precedent of going around the PackedFunc ABI since it is presently so universal.

cyx-6:
Thanks @areusch for your questions! Just to provide more context, it's worth mentioning that the majority of DL compilers built on top of the TVM compiler (not only the runtime) use TVM's C++ APIs directly in the project, i.e. all the C++ APIs under `include/tvm/` are considered public APIs as a common practice in C++ libraries.

> there is a fast path in calling them from C++ that avoid re-encoding the arguments into the documented `PackedFunc` C ABI

To be clear, the primary motivation of this RFC isn't about performance, but about usability - TVM-based compilers use `PackedFunc` either created by TVM or the compilers themselves. The missing piece here is that TVM's `PackedFunc` uses `std::function` internally right now, which is practically not shareable across C++ DLL boundary. Therefore, this RFC mainly focuses on stabilizing the `PackedFunc` into a layout-stable TVM object.

> i'm okay with creating this ABI if it provides a clear benefit over the already-defined one.

To clarify, This RFC isn't creating new ABIs, just refactoring existing ones. Note that existing TVM-based DL compilers all use C++ APIs instead of C APIs (the generated kernel uses C ABI though). This RFC mainly focuses on standardization of a single specific layout-stable implementation of `PackedFunc` so that its content could be shared across the DLL boundary - by "refactoring", the impact is limited to the standard TVM runtime and doesn't affect microTVM runtime and the C API is unchanged.

> such as extracting packed_func.h into a separate e.g. dlpack library and allow you to use the C++ `PackedFunc` wrappers in your code directly

I like this proposal. AFAIK there is long-term need from other libraries (e.g. DGL, MXNet) to incorporate with the TVM object/FFI system because it's really easy to use. However, I would say it's separate need from this RFC, but if anyone is interested, please feel free to do so.

> it would be great to motivate creating this second ABI with some details about why you can't use the already-defined `PackedFunc` ABI.

For context, TVM is largely used as a C++ submodule in DL compilers, where all its C++ APIs under `include/tvm` is treated as public APIs. It's more like a convention for C++ projects.

junrushao1994:
Hey @areusch would you like to follow up on the discussion? Thanks a lot!

Also CC: @comaniac @spectrometerHBH @Hzfengsy if you have further comments

areusch:
thanks @cyx-6 sorry for the delay

> To be clear, the primary motivation of this RFC isn't about performance, but about usability - TVM-based compilers use PackedFunc either created by TVM or the compilers themselves. The missing piece here is that TVM's PackedFunc uses std::function internally right now, which is practically not shareable across C++ DLL boundary. Therefore, this RFC mainly focuses on stabilizing the PackedFunc into a layout-stable TVM object.

I think this is the key part here. can you add this description (about `std::function` not being able to be passed across the DLL boundary) to the Motivation section? I am good with this RFC after that!

junrushao1994:
@cyx-6 would you mind updating the RFC according to our discussion here and on the forum? Also, please include your PR https://github.com/apache/tvm/pull/10032 in the RFC?

cyx-6:
Thanks for @areusch and @junrushao1994 's suggestions. I have updated the RFC according to discussion here and on the forum.

junrushao1994 (approved):
LGTM :-)

areusch (approved):
thanks @cyx-6 !

areusch:
@cyx-6 the RFC is merged! please open a new GitHub issue in apache/tvm (the "Feature Tracking" type), enumerate the PRs expected in that issue, and cite this RFC so others can find it for context. Thanks again for working on this!

PR #52, opened 2022-01-13
[RFC][OpenCLML] OpenCLML integration as BYOC

srkreddy1238:


rfcs/0052-OpenCLML-integratio-as-BYOC.md
```diff
@@ -0,0 +1,90 @@
+- Feature Name: OpenCL ML integration as BYOC
+- Start Date: 2022-01-13
+- RFC PR: [apache/tvm-rfcs#52](https://github.com/apache/tvm-rfcs/pull/52)
+- GitHub Issue: TBD
+
+
+# Summary
+[summary]: #summary
+
+OpenCL ML is an extension (cl_qcom_ml_ops) over OpenCL spec developed by Qualcomm to accelerate the machine learning at operation level. OpenCL SDK is publicly available at OpenCL Machine Learning Acceleration on Adreno GPU - Qualcomm Developer Network. OpenCL ML leverages deep knowledge of Adreno GPU for significant performance benefits. It offers C based DNN API with compatibility to most of the standard frameworks. Its standard OpenCL features like command queues, buffers, events and supports FP16 and FP32 data types. CLML API calls can be interleaved with other OpenCL kernels (i.e., TVM generated kernels) and dispatched to the same command queue. This extension is compatible with existing OpenCL extensions for importing memory, controlling performance and data access.
```
FrozenGene, line 10: `This extension is compatible with existing OpenCL extensions for importing memory`. Do you mean `cl_qcom_ion_host_ptr` for allocating on the ION memory or what else?

rfcs/0052-OpenCLML-integratio-as-BYOC.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The current OpenCL backend of TVM is very generic and not optimized well for Adreno performance capabilities. Adreno GPU has quite a few proprietary and standard OpenCL paths. OpenCL ML extension offers accelerated ML operations via an SDK interface.
+
+With TVM having the entire framework of frontends, graph level optimizations and OpenCL ML having kernels that perform best on Adreno GPU, in this work we aim to integrate OpenCLML SDK into TVM as a BYOC. This effort brings best of both worlds where TVM handling high level optimizations, sub graphs are scheduled on OpenCL ML based on the support and the operators not supported by OpenCL ML will take TVM’s default OpenCL path. Good thing here is we don’t need separate OpenCL workspaces or command queues for both paths, instead they can share the command queues. Also, data (DLTensor) transfer across subgraphs is seamless with OpenCL ML API’s.
```
FrozenGene, line 17: `instead they can share the command queues` -> share the same command queues?

FrozenGene, line 17: `data (DLTensor) transfer across subgraphs is seamless with OpenCL ML API’s` Do you mean `OpenCL ML` has done the extension to support DLTensor?

rfcs/0052-OpenCLML-integratio-as-BYOC.md
```diff
@@ -0,0 +1,90 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC aims to introduce OpenCLML runtime as a BYOC option into TVM. In terms of usage, it’s very similar to other BYOC integrations we have in TVM.
+
+Along with all other options we use for OpenCL target, here we introduce the below build options in config.cmake
+
+```USE_CLML``` (ON/OFF) This enables CLML codegen for compilation
+
+```USE_CLML_GRAPH_EXECUTOR``` (ON/OFF) This enables CLML runtime
```
FrozenGene, line 28: Here means we only support TVM `GraphExecutor` runtime or what you mean? If we use `VM` runtime, could it support?

rfcs/0052-OpenCLML-integratio-as-BYOC.md
```diff
@@ -0,0 +1,90 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+There exists an ongoing development for texture support on Adreno devices https://discuss.tvm.apache.org/t/rfc-texture-memory-support/9467.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+
+How do we deal with sub graphs with tiny layers? This is the case where not offloading the tiny layer performs better than accelerator.
```
FrozenGene, line 85: If we leave one `opencl_ml_op_name_list` for users passing (could disable op or specific model layer), could this help?

rfcs/0052-OpenCLML-integratio-as-BYOC.md
srkreddy1238: CLML tensor object creation happens by creating a ```cl_ml_tensor_desc_qcom``` and then a backing buffer. The backing buffer in general would be a cl_mem object created by clBuffer. Here we can also use ```cl_qcom_ion_host_ptr```.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
srkreddy1238: There is  an opencl work space (context & command queue) already created by TVM's default workspace, For CLML runtime we don't need to create a new opencl context and queue. The same will be reused by accessing it through ```tvm::runtime::Registry::Get("device_api.opencl");```. Advantage here would be the memory objects (clBuffers ...etc) created under default TVM's OpenCL device API and CLML API will be under same context.

CL memory objects under same context allows us to use CL (```clEnqueueCopyBuffer```) / CLML API (```clEnqueueCopyMLTensorDataQCOM```) for data copy across sub graph boundary.

srkreddy1238: We know the BYOC will have sub graphs based on the operator support and there exists a data (DLTensor  to/from runtime specific objects) copy while switching from one sub graph to another. This copy some time require bringing the memory to host and copy to the new runtime. In our case the OpenCL DLTensor is backed by clBuffer and CLML tensor is also backed by clBuffer. Hence, we can use OpenCL (or CLML) copy API for direct copy with in CL context. Or given the data layout is same on both sides we can make CLML Tensor use the clBuffer created by DLTensor directly.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
srkreddy1238: I will need to look into OpenCL support over ```VM```. Given TVM  default ```OpenCL``` target support over ```VM``` CLML also should work.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
srkreddy1238: Yes, such options give the user more control over offloading. This issue is generic for any BYOC we have. Example being,  with BYOC supporting a layer called ```Clip```  and in the graph the layers before and after ```Clip``` not supported by the BYOC. In this case the frontend tries to offload this layer alone to BYOC and the data copy here would be more costlier than executing it on TVM.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
masahi: I dont see a need to distinguish  graph vs vm runtime. They work equally well.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
```diff
@@ -0,0 +1,90 @@
...
+### Runtime:
+OpenCLML Runtime is again extended over JSONRuntimeBase and implements OpenCL ML initialization, implementation of CLML API invocation corresponding to CLML annotated layers.
+
+OpenCLML runtime support is verified by looking for ```cl_qcom_ml_ops``` into OpenCL extension list.
+
+OpenCLML doesn’t define a new open context instead it reused the context defined by OpenCL runtime through global API ```device_api.opencl```.
+
+OpenCLML has its own CLML tensor objects called ```cl_ml_tensor_memory_desc_qcom```. The runtime defines the copy API from OpenCL to CLML Tensors within the same OpenCL work space without bringing the data back to host.
+
+OpenCLML supports tuning too which generally produces a tuning cache file and reuses for later runs. This implementation supports looking for environment variable ```CLML_IS_TUNNING_RUN``` set to 0/1 to run for tuning and also supports ```CLML_TUNNING_CACHE``` to set the tuning cache file location.
```
masahi, line 62: When the tuning happens?

srkreddy1238: CLML tuning happens as the last step of runtime BuildEngine (after the CLML graph is built from JSON nodes). It calls CLML API ```clTuneMLOpQCOM``` followed by ```clSaveMLTuningCacheQCOM``` for saving the cache to a file. Later we have ```clLoadMLTuningCacheQCOM```  for reloading the cache for future runs without running the tune again.

masahi: Would be nice to add that to the doc too.

masahi (approved):

masahi:
I'll merge this week unless there is a further comment cc @FrozenGene

FrozenGene:
LGTM. I don't have any other comments.

rfcs/0052-OpenCLML-integratio-as-BYOC.md
srkreddy1238: Added tuning related info to the doc.

masahi (approved):

PR #53, opened 2022-01-19
[RFC][Roadmap] microTVM roadmap

denise-k:
cc @areusch @Mousius @tqchen


areusch (approved):
LGTM, @mousius could you take a look too? this weekend would be a great time for me to begin assembling this roadmap if we are all good here.

Mousius (approved):
Let's get this setup :smile_cat:!

PR #54, opened 2022-01-20
[RFC][Roadmap] TVM Continuous Integration & Testing Roadmap

denise-k:
cc @areusch @leandron @tqchen

### DIFF
--- a/rfcs/0054-ci-testing-roadmap.md
+++ b/rfcs/0054-ci-testing-roadmap.md
@@ -1,0 +1,108 @@
+- Feature Name: TVM CI & Testing Roadmap
+- Start Date: 2022-01-19
+- RFC PR: [apache/tvm-rfcs#0054](https://github.com/apache/tvm-rfcs/pull/0054)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+- Co-Authors: Denise Kutnick ([**@denise-k**](https://github.com/denise-k)), Andrew Reusch 
+  ([**@areusch**](https://github.com/areusch)), Leandro Nunes ([**@leandron**](https://github.com/leandron))
+
+# [RFC] TVM Continuous Integration & Testing Roadmap
+
+## Background and Context
+
+**Roadmap Name:** TVM Continuous Integration & Testing Roadmap
+
+**Roadmap Maintainers: `@areusch, @denise-k, @leandron` (add more here)**
+
+**Roadmap Description:** This roadmap tracks all efforts related to Continuous Integration and Testing within TVM.
+
+## Scope
+
+**How are the tasks tracked in this roadmap grouped together? How can we think about this grouping distinct from those made in other roadmaps?**
+
+Tasks in this roadmap may be grouped together using **vertical** or **horizontal** tasks.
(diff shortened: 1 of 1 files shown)


areusch (approved):
LGTM, @leandron can you take a look here too?

denise-k:
@areusch can you get some additional reviews on this? thanks!

areusch:
@Mousius since @leandron is a bit busy would you like to review?

leandron (commented):
Sorry it took me a long time to review the RFC.

In general it looks good to me all the proposed ideas. The only thing I think we should add, is a "Security" item, pointing out that we should make sure our infra runs on actively supported versions of tools (meaning they receive security patches) and deprecation mechanisms are in place so that we don't keep old versions around after they reached end-of-life (EOL) status.

Examples of that are currently our usage of Python 3.6 and Ubuntu 16.04.

Apart from that, I'm happy for this RFC to be merged, pending your analysis of my suggestion above.

Mousius:
As an extension of @leandron 's comment above, we should apply the same to packages we install and use as part of TVM - at the bare minimum we should have tools to check packages haven't notified of major vulnerabilities and basic static analysis tools to catch easy to spot vulnerable code. As part of this we need some kind of mechanism for locking versions and performing updates on them regularly (I remember @areusch was keen on `poetry` at one point :wink:).

leandron (approved):
LGTM, I agree with the high level bullets about Security. Thanks for organising this @denise-k!

areusch:
@leandron @Mousius thanks for taking a look! @denise-k updated the RFC to address and scope security. I agree this is important. I think this covers the bit you're mentioning about CI security; I think given the themes of the roadmap, TVM security should fall more into a "release-oriented" roadmap. Currently we haven't specified a roadmap to hold any work around release infra. We could expand this one to hold it, but I'd rather merge this so we can make forward progress on adding the CI & Testing tasks we have now to the existing roadmap, and contemplate a release roadmap in a follow-on RFC. I do indeed want to continue hacking on my poetry-based Python dependency management thing soon.

areusch:
since we're out of cycles to upstream tasks today, @Mousius can you take another look and make sure this agrees with you, and feel free to approve/merge if so.

Mousius:
> @leandron @Mousius thanks for taking a look! @denise-k updated the RFC to address and scope security. I agree this is important. I think this covers the bit you're mentioning about CI security; I think given the themes of the roadmap, TVM security should fall more into a "release-oriented" roadmap. Currently we haven't specified a roadmap to hold any work around release infra. We could expand this one to hold it, but I'd rather merge this so we can make forward progress on adding the CI & Testing tasks we have now to the existing roadmap, and contemplate a release roadmap in a follow-on RFC. I do indeed want to continue hacking on my poetry-based Python dependency management thing soon.

Could you clarify how security is limited to a release? The tooling we use to automate detection of insecure packages and vulnerable code should be ran across all changes rather than checking it as part of a release. We should aim to keep our own CI and development environments secure as a general practice with CI automation to aid us.

Mousius:
@areusch helped me understand the contention here, there's a release process for managing things such as bugs, security vulnerabilities and other forms of fixes in a release cycle and accounting for all of those things in this roadmap is definitely out of scope :smile_cat: 

My confusion was that "Security problems with TVM dependencies" and "Security vulnerabilities in the TVM code itself" can be monitored by automation, which is what I'm suggesting to include as part of this roadmap. The actual tactics of remediation for issues would be defined elsewhere, likely alongside release processes or as part of a another roadmap.

PR #55, opened 2022-01-27
@slow test RFC

driazati:
Also see discussion in the pre-RFC: https://discuss.tvm.apache.org/t/rfc-ci-skip-slow-tests-on-prs/11910

@areusch @Mousius

### DIFF
--- a/rfcs/0055-slow-tests.md
+++ b/rfcs/0055-slow-tests.md
@@ -165,3 +165,5 @@ Over time as slow tests are manually debugged, `@slow` decorators could be remov
 - Better communication in Jenkins job pages of which tests ran, which did not, and why
 - Different levels of tests. `main` is the most frequent step, but longer running tests could be moved out to nightly or even release level testing (though this makes debugging failures more difficult).
 - Gather test coverage data to get a basic idea of what code is being tested
+- Track per-test durations for commits and report on CI runs what the runtime difference of each PR is (so developers can easily see the burden of their change)
+
--- a/rfcs/0055-slow-tests.md
+++ b/rfcs/0055-slow-tests.md
@@ -12,7 +12,7 @@ Add a Python decorator to skip tests on PRs but run them on branches (e.g. `main
 
 [motivation]: #motivation
 
-A small subset of tests take up a large portion of the total test runtime. This RFC proposes that we skip these tests on PRs where CI runtime is critical and run them only on main.
+A small subset of tests take up a large portion of the total test runtime in Pull Requests (PRs). This RFC proposes that we skip these tests on PRs where CI runtime is critical and run them only on `main`.
 
 # Guide-level explanation
 
--- a/rfcs/0055-slow-tests.md
+++ b/rfcs/0055-slow-tests.md
@@ -1,0 +1,110 @@
+- Feature Name: `slow_tests_decorator`
+- Start Date: 2022-01-26
+- RFC PR: [apache/tvm-rfcs#55](https://github.com/apache/tvm-rfcs/pull/55)
+
+# Summary
+
+[summary]: #summary
+
+Add a Python decorator to skip tests on PRs but run them on branches (e.g. `main`).
+
+# Motivation
+
+[motivation]: #motivation
+
+A small subset of tests take up a large portion of the total test runtime. This RFC proposes that we skip these tests on PRs where CI runtime is critical and run them only on main.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+CI runtime constantly plagues TVM developers with long iteration times, exacerbated by flakiness and difficult-to-reproduce steps. To reduce runtime, we can execute more work concurrently and increase usage of available resources, usually by parallelization within (apache/tvm#9834) or between (apache/tvm#9733) CI jobs. Another way is to do less work, which is what this RFC proposes. By running some tests on `main` only, we still get some measure of coverage provided by these tests without burdening PR developers.
+
+The runtime savings of this change are potentially significant, as this gives us a black-box knob which we can manually tune over time to trade off between PR test coverage and PR test runtime. [This gist](https://gist.github.com/driazati/e009f09ff44c6bc91c4d95a8e17fd6f1) shows a listing of TVM test runtime in descending order. And this list the potential time savings (across all jobs) in CI of cutting off tests above an arbitrary runtime (not to propose use a cutoff, but just to demonstrate in broad strokes the potential impact of this change):
+
+```
+[cutoff=10.0s] Total savings (m): 419.95m by skipping 695 tests
+[cutoff=20.0s] Total savings (m): 338.27m by skipping 320 tests
+[cutoff=30.0s] Total savings (m): 291.5m by skipping 205 tests
+[cutoff=40.0s] Total savings (m): 251.59m by skipping 135 tests
+[cutoff=50.0s] Total savings (m): 222.56m by skipping 96 tests
+[cutoff=60.0s] Total savings (m): 203.3m by skipping 75 tests
+[cutoff=70.0s] Total savings (m): 192.68m by skipping 65 tests
+[cutoff=80.0s] Total savings (m): 181.56m by skipping 56 tests
+[cutoff=90.0s] Total savings (m): 171.45m by skipping 49 tests
+[cutoff=100.0s] Total savings (m): 160.36m by skipping 42 tests
+```
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+A decorator `@tvm.testing.slow` will be added (see apache/tvm#10057) that implements
+the above behavior. Skipping slow tests would be an opt-in, rather than opt-out.
+This way developers who don't read this RFC won't have to adjust their workflows
+at all to run these tests locally. There is also a need to run slow tests on PRs
+in some cases, such as fixing reverted commits or if a developer suspects their
+change would have a wide reaching impact. In this case, `@ci run slow tests` can
+be added to the PR body before tests are run in order to disable skipping slow tests.
+
+Using a decorator has the advantage of being explicit compared to an automated system to detect and skip slow tests. Clicking through to the decorator’s short definition makes it clear what is going on so this shouldn’t add too much of a development burden. Slow tests run by default to minimize disruption but can be controlled by setting `SKIP_SLOW_TESTS=1`.
+
+An environment variable `SKIP_SLOW_TESTS=1` will be set in Jenkins on PRs. Branches,
+including `main` and `ci-docker-staging` will not have this flag set and will
+always run the full set of tests.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+The primary caveat is that tests that run on `main` may now fail due to PRs that were green when they were merged, so this will require some buy-in from all TVM developers. However, the runtime savings are significant (see below) enough to make this worth it. Developments like apache/tvm#9554 will make the revert process much smoother as well to minimize disruptions.
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This isn't a complete solution. Most PRs end up running lots of tests that the
+PR didn't affect at all. Ideally we would have a way a-la Bazel or Buck to
+walk the actual dependency graph of PRs (ignoring / banning some of Python's
+dynamism e.g. `importlib.import_module`) to determine what tests to run on a PR.
+This could also be implemented at the human level, with developers tagging their
+PRs based on what they think should run, though this has a higher potential to
+miss certain tests. However, this run-what-changed future would be difficult to
+achieve.
+
+Other efforts involve looking into tests themselves to determine why they are slow.
+Often TVM's tests are running much more work than they actually intend to test in
+more of an integration test than a unit test. Replacing these types of test with
+a framework that makes it easier to test TVM passes and functionality in smaller
+chunks is related but orthogonal to this work. It still has the same issue (coverage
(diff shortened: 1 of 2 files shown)


rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+  3. Who relies on this test? Do they understand the implications of `@slow`?
+
+- Who will monitor `main` for PR-related breakages? What is the SLA on fixes? Recent additions such as [messaging Discord on `main` failures](https://github.com/tlc-pack/ci-monitoring) and keeping track of the last known good commit (apache/tvm#10056) should make this easier.
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+- Better communication in Jenkins job pages of which tests ran, which did not, and why
+- Different levels of tests. `main` is the most frequent step, but longer running tests could be moved out to nightly or even release level testing (though this makes debugging failures more difficult).
```
Mousius, line 110: One of the concerns raised by @areusch is the change in coverage by removing slow tests, it would be good to investigate the difference in coverage both in Python and C++ to help guide us in re-introducing tests.

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+[summary]: #summary
+
+Add a Python decorator to skip tests on PRs but run them on branches (e.g. `main`).
+
+# Motivation
+
+[motivation]: #motivation
+
+A small subset of tests take up a large portion of the total test runtime. This RFC proposes that we skip these tests on PRs where CI runtime is critical and run them only on main.
```
Mousius, line 15:
```suggestion
A small subset of tests take up a large portion of the total test runtime in Pull Requests (PRs). This RFC proposes that we skip these tests on PRs where CI runtime is critical and run them only on `main`.
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+CI runtime constantly plagues TVM developers with long iteration times, exacerbated by flakiness and difficult-to-reproduce steps. To reduce runtime, we can execute more work concurrently and increase usage of available resources, usually by parallelization within (apache/tvm#9834) or between (apache/tvm#9733) CI jobs. Another way is to do less work, which is what this RFC proposes. By running some tests on `main` only, we still get some measure of coverage provided by these tests without burdening PR developers.
```
Mousius, line 21:
```suggestion
CI runtime constantly plagues TVM developers with long iteration times, exacerbated by flakiness and difficult-to-reproduce steps. To reduce runtime, we can execute more work concurrently and increase usage of available resources, usually by parallelization within ([apache/tvm#9834](https://github.com/apache/tvm/pull/9834)) or between ([apache/tvm#9733](https://github.com/apache/tvm/pull/9733)) CI jobs. Another way is to do less work, which is what this RFC proposes. By running some tests on `main` only, we still get some measure of coverage provided by these tests without burdening PR developers.
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+The runtime savings of this change are potentially significant, as this gives us a black-box knob which we can manually tune over time to trade off between PR test coverage and PR test runtime. [This gist](https://gist.github.com/driazati/e009f09ff44c6bc91c4d95a8e17fd6f1) shows a listing of TVM test runtime in descending order. And this list the potential time savings (across all jobs) in CI of cutting off tests above an arbitrary runtime (not to propose use a cutoff, but just to demonstrate in broad strokes the potential impact of this change):
```
Mousius, line 23: Can we include a snippet of this table, maybe the top 5 rows just in case you decide to clean up your gists one day? :smile_cat:

Mousius, line 23:
```suggestion
The runtime savings of this change are potentially significant, as this gives us a black-box knob which we can manually tune over time to trade off between PR test coverage and PR test runtime. [This gist](https://gist.github.com/driazati/e009f09ff44c6bc91c4d95a8e17fd6f1) shows a listing of TVM test runtime in descending order, showing the potential time savings (across all jobs) in CI of cutting off tests above an arbitrary runtime (not to propose use a cutoff, but just to demonstrate in broad strokes the potential impact of this change):
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+[cutoff=80.0s] Total savings (m): 181.56m by skipping 56 tests
+[cutoff=90.0s] Total savings (m): 171.45m by skipping 49 tests
+[cutoff=100.0s] Total savings (m): 160.36m by skipping 42 tests
+```
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+A decorator `@tvm.testing.slow` will be added (see apache/tvm#10057) that implements
```
Mousius, line 42:
```suggestion
A decorator `@tvm.testing.slow` will be added (see [apache/tvm#10057](https://github.com/apache/tvm/pull/10057)) that implements
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+the above behavior. Skipping slow tests would be an opt-in, rather than opt-out.
+This way developers who don't read this RFC won't have to adjust their workflows
+at all to run these tests locally. There is also a need to run slow tests on PRs
+in some cases, such as fixing reverted commits or if a developer suspects their
+change would have a wide reaching impact. In this case, `@ci run slow tests` can
+be added to the PR body before tests are run in order to disable skipping slow tests.
+
+Using a decorator has the advantage of being explicit compared to an automated system to detect and skip slow tests. Clicking through to the decorator’s short definition makes it clear what is going on so this shouldn’t add too much of a development burden. Slow tests run by default to minimize disruption but can be controlled by setting `SKIP_SLOW_TESTS=1`.
```
Mousius, line 50: Maybe mention that this can be extended to C++ tests using the same trigger variable?

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+An environment variable `SKIP_SLOW_TESTS=1` will be set in Jenkins on PRs. Branches,
+including `main` and `ci-docker-staging` will not have this flag set and will
+always run the full set of tests.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+The primary caveat is that tests that run on `main` may now fail due to PRs that were green when they were merged, so this will require some buy-in from all TVM developers. However, the runtime savings are significant (see below) enough to make this worth it. Developments like apache/tvm#9554 will make the revert process much smoother as well to minimize disruptions.
```
Mousius, line 60:
```suggestion
The primary caveat is that tests that run on `main` may now fail due to PRs that were green when they were merged, so this will require some buy-in from all TVM developers. However, the runtime savings are significant (see below) enough to make this worth it. Developments like [apache/tvm#9554](https://github.com/apache/tvm/pull/9554) will make the revert process much smoother as well to minimize disruptions.
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This isn't a complete solution. Most PRs end up running lots of tests that the
+PR didn't affect at all. Ideally we would have a way a-la Bazel or Buck to
```
Mousius, line 67: The Python parallel would be `testmon` as far as I know, may be a better reference? Bazel has a whole host of other idiosyncrasy to deal with.

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+walk the actual dependency graph of PRs (ignoring / banning some of Python's
+dynamism e.g. `importlib.import_module`) to determine what tests to run on a PR.
+This could also be implemented at the human level, with developers tagging their
+PRs based on what they think should run, though this has a higher potential to
+miss certain tests. However, this run-what-changed future would be difficult to
+achieve.
+
+Other efforts involve looking into tests themselves to determine why they are slow.
+Often TVM's tests are running much more work than they actually intend to test in
```
Mousius, line 76:
```suggestion
Often TVM's tests are running much more work than they actually intend to test (such as using entire off-the-shelf networks to test a few operators) in
```

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+more of an integration test than a unit test. Replacing these types of test with
+a framework that makes it easier to test TVM passes and functionality in smaller
+chunks is related but orthogonal to this work. It still has the same issue (coverage
```
Mousius, line 79: How is coverage still reduced if we put in place the better testing practices?

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,110 @@
...
+
+[unresolved-questions]: #unresolved-questions
+
+- What tests do we `@slow`? Based on discussion it seems like going down the list of slow tests one by one (at least to start), it is prudent to do a preliminary investigation to answer:
+
+  1. What is this test? Should it be slow?
+  2. How often does this test fail? If the test fails often, there is less of a case that it should be `@slow`-ed since it provides good signal to developers.
+  3. Who relies on this test? Do they understand the implications of `@slow`?
+
+- Who will monitor `main` for PR-related breakages? What is the SLA on fixes? Recent additions such as [messaging Discord on `main` failures](https://github.com/tlc-pack/ci-monitoring) and keeping track of the last known good commit (apache/tvm#10056) should make this easier.
```
Mousius, line 103:
```suggestion
- Who will monitor `main` for PR-related breakages? What is the SLA on fixes? Recent additions such as [messaging Discord on `main` failures](https://github.com/tlc-pack/ci-monitoring) and keeping track of the last known good commit ([apache/tvm#10056](https://github.com/apache/tvm/pull/10056)) should make this easier.
```

Mousius (changes requested):
Thanks for this @driazati, this is a great concise and meaningful improvement to TVM. I've put a few notes but they're mostly either wording or straight-forward questions.

My personal opinion here is that CI broken often enough on `main` we need to improve that before we can really complain about differences between tests. Alongside this, we still have PRs landing without test coverage, so until we can accurately define what we do and don't test, it seems better to remove friction from contributions - hopefully driving us to ask for more and better tests :smile_cat: 

We should also be prioritising a release schedule rather than treating TVM as ever-green which limits our ability to make larger changes between releases as a broken `main` shouldn't impact end users. This change actually moves us closer to this.

TLDR: Supportive + minor comments :smile_cat:

rfcs/0055-slow-tests.md
driazati: Right, I think that's where flipping the proposal from the pre-RFC from "disable by time cutoff" -> "start with the slowest tests, manually investigate and `@slow`" will help out, since we should have a better idea what's going on with the individual tests that way. Coverage data + proper test reports help out here too, we currently have https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/main/2414/tests but it's broken (doesn't report failures) and we don't have any tooling to scan across those vertically (i.e. from one commit to the next) to see changes

rfcs/0055-slow-tests.md
driazati: it's a bit of both I suppose, if we take an integration test that is running a whole model to test 1 pass and use infrastructure to test just that 1 pass we're making the test more narrow in what its trying to achieve, but in the end it's flexing fewer parts of the system (even if those parts 99% of the time don't matter for the tests).

rfcs/0055-slow-tests.md
```diff
@@ -0,0 +1,167 @@
...
+
+- Who will monitor `main` for PR-related breakages? What is the SLA on fixes? Recent additions such as [messaging Discord on `main` failures](https://github.com/tlc-pack/ci-monitoring) and keeping track of the last known good commit ([apache/tvm#10056](https://github.com/apache/tvm/pull/10056)) should make this easier.
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+- Better communication in Jenkins job pages of which tests ran, which did not, and why
+- Different levels of tests. `main` is the most frequent step, but longer running tests could be moved out to nightly or even release level testing (though this makes debugging failures more difficult).
+- Gather test coverage data to get a basic idea of what code is being tested
```
areusch, line 167: would add to future possibilities: longitudinal tracking of test time, possibly reporting the difference on the PR, so that folks can get an idea of how much additional burden they are adding.

areusch (commented):
this looks pretty good to me @driazati , thanks for the writeup! I'll defer to @Mousius for approval since he had more comments here.

areusch:
@Mousius can you take another look when you have a min?

rfcs/0055-slow-tests.md
Mousius: I just re-read this again, and I think that explicitly stating the roll out strategy as a manual approach of test investigation would clarify this to me - i.e. there's no automation involved albeit you've demonstrated we could automate it.

Mousius (changes requested):

rfcs/0055-slow-tests.md
driazati: Yeah I see what you mean, it all comes down to implementation/how it's used, but either way it's pretty immaterial to this RFC so I removed that language

rfcs/0055-slow-tests.md
driazati: Added a bit about implementation in the guide section

Mousius (approved):
Thanks for the clarification @driazati, you're clear to land :smile_cat:!

PR #57, opened 2022-02-18
[RFC]FamilySeer: A new search method for Auto-scheduler

noobotdj:
RFC topic in forum: https://discuss.tvm.apache.org/t/rfc-familyseer-a-new-search-method-for-auto-scheduler/11877

@comaniac @junrushao1994


comaniac:
Thanks for the RFC. Will review next week.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+    measure_callbacks=[auto_scheduler.RecordToFile(log_file)],
+)
+
+# start tuning
+#tuner.tune(tune_option) #add new parameter to tune function 
+tuner.tune(tune_option,search_policy="sketch.xgb.family_op")
+
+```
+
+The `tuner` loads the `tune_option` into the `tune` function. There are several parameters in the `tune` function (Refer to class [Taskscheduler](https://tvm.apache.org/docs/reference/api/python/auto_scheduler.html?highlight=taskscheduler#tvm.auto_scheduler.TaskScheduler)). Users can enable our method by changing the `search_policy` parameter to `sketch.xgb.family_<family_algorithm>`. We currently provide two family algorithms as an option: `op` refers to classifying subgraphs based on the core operation, and `hash` refers to classifying subgraphs based on operation sequence. We recommend using `op` to achieve better performance.
```
comaniac, line 54: By this description, `op` seems not an accurate term. IIUC, both `op` and `hash` target to subgraphs. The only different is the way to determine the similarity, so again, it would be good to have a better naming.

In addition, since you recommend to use `op` for better performance, could you elaborate the need of `hash`?

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+          first = task.workload_key.find("[\"") + 2
+          end = task.workload_key.find("\",")
+          task_hash = task.workload_key[first:end]
+          if task_hash not in family_group:
+              family_group[task_hash] = []
+              family_group[task_hash].append(idx)
+          else:
+              family_group[task_hash].append(idx)
+
+  elif class_type == "ind":
```
comaniac, line 103: Is `ind` the third version? I didn't find it in the previous section.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+          _, _, model_group = policy
+          _, class_type = model_group.split("_")
+      else:
+          raise ValueError("Invalid search policy: " + search_policy)
+      
+  family_group = {}
+  if class_type == "op":
+      for idx, task in enumerate(tasks):
+          task_layers = task.desc.split('_')
+          if task_layers[1] not in family_group:
```
comaniac, line 86: It seems to me that this is a bit too ad-hoc:
1. It relies on the task.desc, which is supposed to be used only for user reference and the format isn't guaranteed.
2. It identifies the second op (e.g., `task_layers[1]`) to be the anchor op, but this may not 100% hold.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+# Motivation
+[motivation]: #motivation
+
+Auto-scheduler (Ansor) uses code sketch and optimization rules to generate a large search space. The search space defined by Ansor has shown great opportunities and therefore the search quality and the search efficiency are determined by how we search the space.
+
+Ansor utilizes improved cost model and task scheduler to help explore the search space. The cost model analyzes and finds high-performance code transformations in the search space and the task scheduler allocates the time budget to different computation graphs. However, we find serval drawbacks to this approach:
+
+The accuracy of the cost model determines the search quality, but Ansor uses monolithic cost model to predict different computation graphs (subgraphs), resulting in an accuracy loss during tuning.
+
+The task scheduler allocates most of the time budget to subgraphs with most improving potential (i.e., those with the highest latency). This approach works well at the beginning of the autotuning. However, as the potential subgraph gradually reaches its peak performance with adequate time budget, other subgraphs have little time budget to reach its peak performance.
```
comaniac, line 24: IMHO, the solution to this problem is to improve the task scheduler strategy. Since the task scheduler predicts the peak performance of a task, as long as this prediction is accurate enough, it shouldn't spend more time on the task that has achieved the peak performance.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+
+After identifying similar subgraphs, We return `family_group` (a list of subgraph families) and build a cost model for each subgraph family.
+
+3.Foresee tuning
+
+```python
+def _tune_family_task(self, task_idx_groups,skip_measures_per_round):
+  """Tune the select family task for one round.
+  """
+  for task_idx in task_idx_groups:
```
comaniac, line 169: IIUC, in this case once TaskScheduler picks a family, all tasks in this family will be tuned for one round? If so, it seems not that efficient as some tasks may not need that many trials. Can we still let task scheduler pick one task at a time, but we just use different cost model based on the family it belongs to?

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+      callback.pre_tune(self, task_idx)
+
+    measure_inputs, measure_results = self.search_policies[task_idx].continue_search_one_round(
+      skip_measures_per_round, self.measurer
+    )
+
+    ……
+```
+
+The foresee tuning takes `task_idx_groups` (A list of subgraph families) and `skip_measures_per_round` as inputs and tunes all the subgraphs inside the list.
```
comaniac, line 181: What is `skip_measures_per_round`?

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+
+    ……
+```
+
+The foresee tuning takes `task_idx_groups` (A list of subgraph families) and `skip_measures_per_round` as inputs and tunes all the subgraphs inside the list. 
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+When searching on a larger search space (such as larger batch size), FamilySeer performs similarly or sometimes worse than Auto-scheduler. This is because a larger search space requires more time before the cost model can provide an accurate prediction. Deploying an inaccurate cost model on Foresee tuning may result in spending time budget on non-improvement code transformations.
```
comaniac, line 186: It seems to me that this can be improved somehow. For example, at the warmup stage (e.g., the first N trials), all cost models share the same training data from all tasks so that you have the same behavior as the current auto-scheduler. Afterward, you apply different data to each cost model to benefit from the task groups.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Auto-Scheduler generates a large enough search space, so searching the space with efficiency is important. With FamilySeer, Users can search for the same optimal code under less time budget. We hope that our search method can be an alternative option for those who expect to obtain better optimal code under a limited time budget.
+
+# Prior art
+[prior-art]: #prior-art
+
+Please refer to [this paper](https://arxiv.org/abs/2201.00194).
```
comaniac, line 196: This is not a "prior" art lol
You should just put this link to the summary section. And for this section, you should put other related work (e.g., Ansor, FlexTensor, AutoTVM, etc).

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Our search method is up for [discussion](https://discuss.tvm.apache.org/t/rfc-familyseer-a-new-search-method-for-auto-scheduler/11877).
```
comaniac, line 201: This is not an unresolved question. Unresolved questions should be the technical difficulty or the drawbacks of the proposed approach.

rfcs/0057-FamilySeer.md
```diff
@@ -0,0 +1,212 @@
+- Feature Name: (FamilySeer: A new search method for Auto-scheduler)
+- Start Date: (2021-01-07)
+- RFC PR: [apache/tvm-rfcs#57](https://github.com/apache/tvm-rfcs/pull/57)
+- GitHub Issue: [apache/tvm#9875](https://github.com/apache/tvm/pull/9875)
```
comaniac, line 4: The link you put is the pull request but not the tracking issue. Please open a new issue to track the PR progress and put the link here.

comaniac (commented):
Thanks again for the RFC. Please see my comments.
Moreover, it would be good to have a tracking issue with a PR plan, which could have a list of PRs to be filed. In addition to the major change and unit tests, you should also have a PR working on tutorials, documents, etc.

PR #60, opened 2022-03-08
Rfc uma

MichaelJKlaiber:
opening PR for UMA pre-RFC


rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,184 @@
...
+│   ├── backend.py
+│   ├── codegen.py
+│   ├── passes.py
+│   ├── patterns.py
+│   ├── schedules.py
+│   └── strategies.py
+└── accelerator_B
+    └── ...
+```
+A custom accelerator is added by inheriting the `UMABackend`. New elements (e.g., passes, schedules) are added using a registration machanism.
```
manupa-arm, line 136: I think this part should be Guide Level explaination as it explains to a "user" how this feature could be used.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,184 @@
...
+        # Relay to Relay function registration
+        self._register_pattern("conv1d_relu", conv1d_relu_pattern())
+
+        self._register_relay_pass(1, ConfigGenerator())
+        self._register_relay_pass(2, BufferScopeAnnotator())
+
+        # Relay to TIR function registration
+        self._register_operator_strategy("nn.conv1d", custom_conv1d_strategy, plevel=9)
+
+        self._register_tir_schedule(insert_extern_calls)
```
manupa-arm, line 153: Can you explain what is the type of "insert_extern_calls" here ?

manupa-arm, line 153: Also, I think it needs explanation/motivation why we need to two hooks ( _register_tir_schedule vs _register_tir_pass)

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,184 @@
...
+
+The figure below describes the UMA interface from a top level. An *Accelerator Partitioner* which is a specialization of the *UMA Partitioner* takes the Relay graph and matches for supported and unsupported operators. Unsupported operators are processed with the default TVM flow. Supported operator are processed with **UMA Pipeline**.
+In the following the tasks and the functionality of each block in the figure below is described:
+
+![](uma_toplevel.png)
+
+UMA Partitioner: 
+* Register relay passes
+* Register patterns - supported sub-graph operations
+* Order: pre-partitioning passes, Graph partitioning, post-partitioning passes
```
manupa-arm, line 67: Maybe for a new user/developer, it might beneficial explaining where to position a relay pass between "post-partitioning passes" and a "_register_relay_pass" -- which might seem not obvious who dont have deeper understanding of TVM . I think it is mainly because, post-partitioning passes run before OptimizeImpl(...) sequence of passes are run in the core compiler.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,184 @@
...
+    def target_name(self):
+        return "ultra_trail"
+```
+Once a `UMABackend` is registered, it hooks into the usual `relay.build` process to create the code for the target accelerator.
+```
+# Load model
+mod, params = relay.frontend.from_pytorch(scripted_model, [("input_data", input_shape)])
+
+# Register a UMA backend
+UltraTrailBackend().register()
```
manupa-arm, line 171: Is this the only thing the user need to do to register the backend ? 
Will it do something to the effect : 

```
TVM_REGISTER_TARGET_KIND("accelerator_B", kDLCPU)
    .set_attr<FTVMRelayToTIR>("RelayToTIR", relay::contrib::generic::RelayToTIR("accelerator_B"));
    .set_attr<FTVMTIRToRuntime>("TIRToRuntime", relay::contrib::generic::accelerator_B::TIRToRuntime);
```

manupa-arm (changes requested):
Thanks @MichaelJKlaiber for this work!

I have done a round of review.

One thing that is missing is how UMA Partition operators (in the Reference Level explaintation). It would be great to describe what exactly being done using the registered patterns.

One thing I'd like anwsered here what sort of control it will allow on the passes run there : MergeComposite, AnnotateTarget, MergeCompilerRegions and ParititionGraph.

For e.g. some backend might not want the compiler regions merged, how would that be controlled ?

Also how would one register post-partitioning passes ?

areusch:
cc @comaniac @csullivan @jroesch @tmoreau89 @masahi

cgerum:
> One thing I'd like anwsered here what sort of control it will allow on the passes run there : MergeComposite, AnnotateTarget, MergeCompilerRegions and ParititionGraph.

So far we had planned to standardize on MergeComposite, AnnotateTarget, MergeCompilerRegions and ParititionGraph. To get a better overview I extracted the partitioning flows of existing BYOC targets:

|BYOC Backend    | Pre Partition Passes                                                                    |  Partition                                                                   |  Post Partition Passes                                                                     |
|----------------|-----------------------------------------------------------------------------------------|------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------|
|arm_compute_lib | InferType,                                                                              |  MergeComposite, AnnotateTarget, PartitionGraph                              |                                                                                            |
|bnns            | InferType, FoldConstant, FoldScaleAxis, DynamicToStatic, AlterOpLayout, FoldConstant,   |  MergeComposite, AnnotateTarget, PartitionGraph                              |                                                                                            |
|cmsisnn         |                                                                                         |  MergeComposite, AnnotateTarget, PartitionGraph,                             | GenerateCMSISNNConstants, ScalarToTensorConstants, ExtractConstantsFromPartitionedFunction |
|cutlass         | SimplifyInference, FoldConstant, FoldScaleAxis,                                         |  MergeComposite, AnnotateTarget, PartitionGraph                              |                                                                                            |
|dnnl            |                                                                                         |  MergeComposite, AnnotateTarget, MergeCompilerRegions, PartitionGraph        |                                                                                            |
|ethosu          |                                                                                         |  MergeComposite, AnnotateTarget, MergeCompilerRegions, PartitionGraph,       |   preprocess_ext_io                                                                        |
|tensorrt        | RemoveDropoutPass, RemoveUnusedFunctions, ConvertLayout, FoldConstant,                  |  AnnotateTarget, MergeCompilerRegions, PartitionGraph                        |                                                                                            |
|vitis_ai        | RemoveUnusedFunctions, ConvertLayout, FoldConstant, InferType,                          |  ("VitisAIAnnotationPass"), MergeCompilerRegions, PartitionGraph,            |   RemoveUnusedFunctions, ConvertLayout, FoldConstant                                       |

Looking at the existing backends it might make sense to make MergeCompilerRegions optional. We probably do not want to support custom compiler annotations as used in `vitis_ai` target.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: For the pass registrations (`_register_relay_pass` and `_register_tir_pass`) we are following the idea of [phases](https://github.com/apache/tvm/blob/main/src/driver/driver_api.cc#L236), at which the passes are registered `def _register_relay_pass(self, phase: int, relay_pass: tvm.transform.Pass) -> None`. E.g. phase 0 would be pre-partitioning, phase 1 would be post-partitioning but before OptimizeImpl.

I agree, that the phases and their implications need proper documentation to help users decide where to place a pass.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
MichaelJKlaiber: You are right. Will move this part into the Guide Level Part

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
cgerum: > Is this the only thing the user need to do to register the backend ? Will it do something to the effect :

Yes. 
 
> ```
> TVM_REGISTER_TARGET_KIND("accelerator_B", kDLCPU)
>     .set_attr<FTVMRelayToTIR>("RelayToTIR", relay::contrib::generic::RelayToTIR("accelerator_B"));
>     .set_attr<FTVMTIRToRuntime>("TIRToRuntime", relay::contrib::generic::accelerator_B::TIRToRuntime);
> ```


`backend.register` interacts with TVM core in 2 ways.
1. It registers the target hooks as you said. 
2. It registers the target pattern tables using `tvm.relay.op.contrib.register.register_pattern_table` and ensures that both use the same compiler/target_name .

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: Thanks!

Should we use well defined enums instead ?

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: Ack. So this is function that defines something like ? : 

```
# Get block by its name
block_b = sch.get_block("B")
# Get loops surronding the block
(i,) = sch.get_loops(block_b)
# Tile the loop nesting.
i_0, i_1, i_2 = sch.split(i, factors=[2, 2, 2])
```

I think a reference explanation how the IRModule get transformed using an example (a single pass for each point of registeration) of  can go a long way :).

manupa-arm:
@cgerum thanks for detailed analysis!

Im wondering whether should we provide an optional partitioning hook as well -- so then it can be anything and let the default be a Sequential of MergeComposite, AnnotateTarget, MergeCompilerRegions, ParititionGraph. WDYT ?

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: The reason Im asking this the above tvm.tir.Schedule based transform can also be viewed as an implementation detail of S-TIR --> S-TIR. 

Anyway the hook will get S-TIR in the beginning. In an alternate method, one could use : 

_register_tir_pass(STIR_Pass_1) <---- this could/should use tvm.tir.Schedule inside to do the transformation
_register_tir_pass(STIR_Pass_2)
_register_tir_pass(STIR_Pass_3)
_register_tir_pass(NSTIR_Pass_4)
_register_tir_pass(NSTIR_Pass_5)
_register_tir_pass(NSTIR_Pass_6)

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: I think for the text here (where we make a decision at the end). I think we should enumerate the following options : 

P1. Int based : _register_relay_pass(self, phase: int, relay_pass: tvm.transform.Pass) -> None
P2. Enum based : _register_relay_pass(self, phase: tvm.transform.uma.Phase, relay_pass: tvm.transform.Pass) -> None
P3. Seperate registerations : 
             _register_pre_partition_relay_pass(self, relay_pass: tvm.transform.Pass)
             _register_post_partition_relay_pass(self, relay_pass: tvm.transform.Pass)
             _register_post_optimization_relay_pass(self, relay_pass: tvm.transform.Pass)

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: Hmmmmm.. TVM_REGISTER_TARGET_KIND is a C/C++ macro. Do we envision to have some sort of python decorator (as part of this work) to handle the registration?

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: Yes, exactly.

> I think a reference explanation how the IRModule get transformed using an example (a single pass for each point of registeration) of can go a long way :).

I agree. We will add simple examples for each registration.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: > Should we use well defined enums instead ?

Yes, we were talking about this internally as well. Enums are probably the preferred solution. We will update the section with the options and add our reasoning of choice. Thanks for the great input!

sunggg (commented):
Thanks for the great proposal! I'm interested in how we can customize pass pipeline for diverse contexts and have a question regarding this. Look forward to learn more about UMA!

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,184 @@
...
+├── backend.py
+├── api
+│   ├── codegen.py
+│   ├── lower.py
+│   ├── partitioner.py
+│   └── utils.py
+├── accelerator_A
+│   ├── backend.py
+│   ├── codegen.py
+│   ├── passes.py
```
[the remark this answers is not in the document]
MichaelJKlaiber: Hi @sunggg ,
UMA use the Relay-to-TIR hook to apply accelerator-specific passes. In the Figure find it in the UMALower step.
A pass relevant for one accelerator is registered via 
```UMABackend._register_pass()```.
A pass that would be relevant for all possible (or a vast majority) of all accelerators needs to be part of UMALower in this concept. I will be possible to activate/deactivate some passes, however, we want tot limit this, as the general idea of UMA is to reduce the space of all the possible descisions. (Will post an update version of the RFC later today to explain how to customize certain part of UMA using configurations parameters).

@sunggg , what pass are you thinking about when you talk about " generic accelerator passes"?  

CC: @cgerum  @PaulPalomeroBernardo

File: rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md (hunk 1 above)
Comment (MichaelJKlaiber) [reply rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md:116]: Done ... let me know if this is structured the way you would expect :)

File: rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md (hunk 4 above)
Comment (MichaelJKlaiber) [reply rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md:189]: @manupa-arm: This part was removed. The C++ code you mention is no longer needed for target registration.

CC: @cgerum

**Comment by cgerum** [comment]
> Im wondering whether should we provide an optional partitioning hook as well -- so then it can be anything (i.e. any Sequential) and let the default be a Sequential of MergeComposite, AnnotateTarget, MergeCompilerRegions, ParititionGraph. WDYT ?

Considering how partitioning is handled in #62 I would probably prefer a more declarative way of specifying different partitioning patterns. @MichaelJKlaiber @PaulPalomeroBernardo

File: rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md (hunk 4 above)
Comment (PaulPalomeroBernardo) [reply rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md:189]: Just a short addition regarding the implementation. We are using a global function `RegisterTarget` in C++, that takes the target name and registers the target together with the target hooks. `RegisterTarget` is called during the backend registration `UltraTrailBackend().register()`. To hide this process from the user we are not using a decorator, but I think it's a similar approach.

Comment (manupa-arm) [reply rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md:189]: That make sense to me! It might be worth adding this to the Reference explanation :) 

How/Do we deal with target specific options ?

E.g. : 
https://github.com/apache/tvm/blob/fe7b5d329a82f720a721356c40abd721cf1d780d/src/target/target_kind.cc#L373

**Review by lhutton1 — COMMENTED** [review]
Thanks for the great proposal @MichaelJKlaiber! While reading the RFC I picked up on a couple of small things, feel free to ignore them :)

One overall question I have is whether this proposal is strictly limited to accelerators or whether it could also be used by any back-end that leverages the target hook functionality? For example, it seems possible to register kernel libraries (e.g. CMSIS-NN) using a similar interface?

File: rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md [hunk 6]
```diff
@@ -0,0 +1,228 @@
...
+* API level:
+    * UMA Partitioner creates a wrapper API to TVM core-compiler APIs
+    * *UMAPartitioner* baseclass (Python only) has to be inherited by accelerator-specific Partitioners (e.g. Accelerator A Partitioner, etc)
+
+The figure below described the *UMA Pipeline*. The blocks are described below:
+
+![](uma_pipeline.png)
+
+UMA Pipelining:
+* Consists of UMALower and UMACogen, which implement the target hook Relay-to-TIR and TIR-to-Runtime (proposed in [TVM-RFC0010])
```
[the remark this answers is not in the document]
MichaelJKlaiber: fixed. thanks

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,228 @@
...
+  * Output:  NS-TIR(including tir.extern calls)
+  * UMALower baseclass (Python only) has to be inherited by accelerator-specific Lower classes (e.g. Accelerator A Lower, etc)
+* UMACodegen
+  * Input: NS-TIR(including tir.extern calls)
+  * Defaults to standard TVM codegen
+  * Intend is to provide a Python interface to insert/emit target code
+  * UMACodegen baseclass has to be inherited by accelerator-specific Codegen classes (e.g. Accelerator A Codegen, etc)
+  * Output: Target .c files
+
+The intention is to use TensorIR with MetaScheduler for optimization and Relax in later versions.
```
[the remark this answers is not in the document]
MichaelJKlaiber: thanks. Added it to the text

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,228 @@
...
+
+The **goal** of UMA is to establish two API layers with a different target groups of users:
+
+**Porcelain Layer**: UMA
+  - Straight-forward, *Python-only* and stable API wrapper of plumbing layer
+  - Easy and clearly-defined template for integration of accelerators
+  - Short learning period for hardware/software engineers new to TVM
+
+**Plumbing Layer**: 
+  - Collage-like API + other TVM APIs
```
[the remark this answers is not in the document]
MichaelJKlaiber: Wasn't available when we published the latest version, so that's for the hint! Link is added.

MichaelJKlaiber:
> One overall question I have is whether this proposal is strictly limited to accelerators or whether it could also be used by any back-end that leverages the target hook functionality? For example, it seems possible to register kernel libraries (e.g. CMSIS-NN) using a similar interface?

@lhutton1, I agree in general. The primary focus of UMA at the moment is accelerators. It might make sense to bear library integration in mind. In general, I see that it should be possible. The main difference might be the choice of configuration parameters needed.
@cgerum @PaulPalomeroBernardo  what are your thoughts here?

@lhutton1, if you have suggestion or concrete examples, feel free to share them 

- Michael

lhutton1:
Thanks @MichaelJKlaiber, that makes sense. So I was wondering if this is the case, perhaps in the future this interface is used by other backend's (not accelerators) we would need to think about renaming UMA to something more generic e.g. UMB _Universal Modular Backend_ - I'm not the best with names. I'm wondering if this could easily be done in the future?

MichaelJKlaiber:
> Thanks @MichaelJKlaiber, that makes sense. So I was wondering if this is the case, perhaps in the future this interface is used by other backend's (not accelerators) we would need to think about renaming UMA to something more generic e.g. UMB _Universal Modular Backend_ - I'm not the best with names. I'm wondering if this could easily be done in the future?

@lhutton1, I think for naming there are no limits for creativity, e.g. it could be _UMA: Universal Modular bAckend_ 😄

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: With commit 0dda395eda42e7c286e06c503f6215e30bd1a2f8 we added a bit of [documentation](https://github.com/boschresearch/tvm-rfcs/blob/rfc_uma/rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md#umabackend-references) for the API with small examples of what could be registered in each step.

PaulPalomeroBernardo: **TL;DR**
We decided to remove the `_register_tir_schedule` hook for the first version of the UMA API. So transforming S-TIR would work exactly as you described as part of a TIR-pass.

**Background**
Initially it was the plan to create a stable API that would, in the future, also support an easy integration of the MetaScheduler for design space exploration and optimization. This would require a separation of scheduling functions and different kind of transformation passes. For now, we decided to separate the MetaScheduler integration from this RFC and and move it to future work. As a result we removed the additional `_register_tir_schedule` hook.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: I added it to the Reference explanation [(link)](https://github.com/boschresearch/tvm-rfcs/blob/rfc_uma/rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md#uma-target-hooks).

Currently it is not possible to set target specific options. Since the UMA targets are essentially also "c" targets we did not see the need to deal with target specific options. Do you have a use-case in mind, for which this would be necessary?

manupa-arm: Well, we currently use such options to define accelerator variants that share the same lowering pipeline.
In the absense of that, we would need to resort to use the PassConfig, however, generally PassConfig is generally better suited to set a configuration to a specific Pass. In the above case, it would mean we need to set multiple Passes in the absense.

I would see the newly registered targets as extensions of the "c" target and Im a bit keen on not ending up having to dump a union of UMA target options to "c" target. 

Following your proposal, is there a reason why we wont be able to use RegisterTarget ? We could consider including AttrDict to that effect.

Adding to this, there are two variants of these : 
 
 * relay.ext.<backend>.options
 
 Which define the options for the lowering. This is inherited by the original BYOC design and we still use it with Target Hooks. This is partly due to the seperate existence kCompiler strings and actual targets.
 
 * target_kind options
 
 This is what I alluded to in the previous comment.
 
Ideally, since UMA is wrapping Target Hooks, I suppose if we want to add this, we would want to proceed with the second option -- hence the suggestion.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
manupa-arm: I think is resolved now.

manupa-arm: looks great to me! resolving...

areusch:
@MichaelJKlaiber is this one ready for another look?

MichaelJKlaiber:
> @MichaelJKlaiber is this one ready for another look?

Sure 👍

manupa-arm:
@MichaelJKlaiber, I did another pass. It is looking good!

From my side, Ideally, there are only two things that we need to agree upon on the design (which relates two outstanding conversations above).

A1. The rationale to use the int-based phase in registering passes.
A2. The methodology to attach target-specific options that affects multiple passes in the lowering.

For A2, it would be bit important to reach a conclusion where we could point future contributors/users who want to use UMA for upstream/downstream development, respectively.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
cgerum: > Adding to this, there are two variants of these :
> 
> * relay.ext.<backend>.options
> 
> Which define the options for the lowering. This is inherited by the original BYOC design and we still use it with Target Hooks. This is partly due to the seperate existence kCompiler strings and actual targets.

Yes this not exactly user friendly.

> * target_kind options
> 
> This is what I alluded to in the previous comment.
> 
> Ideally, since UMA is wrapping Target Hooks, I suppose if we want to add this, we would want to proceed with the second option -- hence the suggestion.

Adding them to the TargetKind would obviously be the preferred solution. As would potentially be a preferred way of doing things. It would allow for a clean `target = Target("ethos-u -version=ethosu-256", host=Target("c"))`  but this makes the current partitioning Flow somewhat clumsy as we still need to run `.partition` for the target hooks. On the other hand it could serve as a starting point for deprecating `partition_for` and providing a unified API for Collage and non-collage flows. @mbs-octoml @areusch would that make sense?

areusch:
for A2, we should consider the type of options present:
- options that serve to specify how a pass runs should eventually find their way into the active `PassContext`
- options which are constraints can be attached to `Target` or similar. 

I think we could use [Target-registered compilation flow customization](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0010-target-registered-compiler-flow-customisation.md) to hold these options. What do you think?

areusch:
@MichaelJKlaiber @cgerum any more thoughts on @manupa-arm 's two questions above? would be great to get this RFC in!

manupa-arm:
@areusch and @MichaelJKlaiber, I agree with using [Target-registered compilation flow customization](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0010-target-registered-compiler-flow-customisation.md).

I am struggling to how to connect that with : 
```
TVM_REGISTER_GLOBAL("relay.backend.contrib.uma.RegisterTarget")
    .set_body_typed([](String target_name){
        ::tvm::TargetKindRegEntry::RegisterOrGet(target_name)
        .set_name()
        .set_device_type(kDLCPU)
        .add_attr_option<Array<String>>("keys")
        .add_attr_option<String>("tag")
        .add_attr_option<String>("device")
        .add_attr_option<String>("model")
        .add_attr_option<Array<String>>("libs")
        .add_attr_option<Target>("host")
        .add_attr_option<Integer>("from_device")
        .set_attr<FTVMRelayToTIR>("RelayToTIR", relay::contrib::uma::RelayToTIR(target_name))
        .set_attr<FTVMTIRToRuntime>("TIRToRuntime", relay::contrib::uma::TIRToRuntime);
    });
```

What would be the API to add an option ? 

e.g. from here 

```
class UltraTrailBackend(UMABackend):
    def __init__(self):
        super(UltraTrailBackend, self).__init__()

        # Relay to Relay function registration
        self._register_pattern("conv1d_relu", conv1d_relu_pattern())

        self._register_relay_pass(1, ConfigGenerator())
        self._register_relay_pass(2, BufferScopeAnnotator())

        # Relay to TIR function registration
        self._register_operator_strategy("nn.conv1d", custom_conv1d_strategy)

        self._register_tir_pass(0, CodegenGenerateConfig())
        self._register_tir_pass(0, CodegenGenerateExternCalls())

        # TIR to runtime function registration
        self._register_codegen(format="c", includes=gen_includes, replace_call_extern=None)

    @property
    def target_name(self):
        return "ultra_trail"
```

manupa-arm:
In order to help to progress with A2,

I can think of two solutions : 

A2.1 :  We could register a sub_target string (similiar to a mtriple in LLVM) and a decoder (e.g. : _register_sub_target_decoder(ultra_trail_target_decoder(str)) --> AttrDict/Dict. We could register the attr_dict as uma_attrs to the templated UMA target registration.

A2.2 : We could directly register an attr_dict. i.e. _register_target_attr(AttrDict/Dict)

I have a slight preference for A2.1 as it will only expose a string to the user in the actual compilation run.

PS : 
@Mousius highlights the motivation behind making the target attributes discoverable in this new recent RFC : https://github.com/apache/tvm-rfcs/blob/1901b7361891fc7235b3bb8e80c89130cfe2c91d/rfcs/0070-target-preprocessing.md

manupa-arm:
For A1,

My personal preference is that we go for Enum based approach rather than int s.

Unless, of course, we have a good reason not to do that -- which I think should be outlined in the RFC for future reference.

areusch:
for A1, what do people think of an ordering spec e.g.
```
self._register_relay_pass(ConfigGenerator(), before="FuseOps", after=("MergeComposite", "abcOtherPass"))
```

my concern is that whether int or enum, people are really expressing a dependency graph here and while we hope it is not terribly complicated, it's hard to intuit the meaning from an enum/int.

for A2, i agree with @manupa-arm 's question, but i think it would be awesome to see the prototype and we could discuss from there.

cgerum:
> The main downside I see with this, is that all attributes are treated as strings since the type is hardcoded. However, I'm not sure if we can avoid this at all.

This could probably be solved by adding `type` and/or `default` arguments to the argument parser, e.g.:

```
  self._register_target_attr("ultra_trail_attr_1", default=False)
```

For v1 I would prefer to only support default values, and restrict supported dtypes to string, int and bool.

> For A1: I would like to keep the phases. They definitely need proper documentation, but I think a handfull of phases (e.g., `PRE_PARTITIONING`, `POST_PARTITIONING`, ...) provide more orientation for new users than having to explicitly define the dependencies to other passes. We could think of also supporting the `before` and `after` options to provide more flexibility for experienced users.

I agree with @PaulPalomeroBernardo on the user perspective. 
Implementing before and after options, also goes beyond the current Scope of UMA in my opinion. If we were to expose it in the UMA-API, pass dependencies should probably be implemented in TVM core.

MichaelJKlaiber:
Thanks @cgerum and @PaulPalomeroBernardo . I agree, this totally makes sense like this.

@manupa-arm @areusch is this sufficiently detailed for you? I propose to discuss outstanding topics in meeting to settle for UMA-v1. We could use the Community Meeting on May 25th. Or if these discussions are too specific for a broader audience, then we can setup a separate meeting.

What are your thoughts?

manupa-arm:
```
They can be used during target creation similar to other sub_target strings

ut_target = tvm.target.Target("ultra_trail -ultra_trail_attr_1=attr1 -ultra_trail_attr_2=attr2")
```
```
This could probably be solved by adding type and/or default arguments to the argument parser, e.g.:

  self._register_target_attr("ultra_trail_attr_1", default=False)
```

This aligns with A2.2 -- directly registering each attribute. I think this is fine for UMA-v1 and aligns with state of TVM targets today. Should we just put a note that for future considerations, to include a registration for string preprocessor (A2.1) to extract attributes ?

```
For A1:
I would like to keep the phases. They definitely need proper documentation, but I think a handfull of phases (e.g., PRE_PARTITIONING, POST_PARTITIONING, ...) provide more orientation for new users than having to explicitly define the dependencies to other passes. We could think of also supporting the before and after options to provide more flexibility for experienced users.
```

Again, I think phase approach is fine for v1 as we already have that in the core compiler (which is also int based) but I'd appreciate if we can put a "name" to ease the reasoning in future. Similarly, we could also note as future work to define dependencies on passes -- if and when the TVM core compiler improve its pass infrastructure we could be able to use that information.

PaulPalomeroBernardo:
> This aligns with A2.2 -- directly registering each attribute.

@manupa-arm Then just for clarification a few questions because I might have misunderstood your initial idea. For A2.1 you were thinking about registering an attribute preprocessor to the target `Target().add_attrs_preprocessor(Preprocessor)` that would operate on a predefined attribute (e.g., `-uma_attrs=<string>`) by processing the `<string>` and creating a Dict/Map from it?

So a user would only write `tvm.target.Target("ultra_trail -uma_attrs=<my custom attr string>")` and in code you would access the target via `target.attrs["uma_attrs"]["attr1"]`, `target.attrs["uma_attrs"]["attr2"]`, ect.?

manupa-arm:
> So a user would only write tvm.target.Target("ultra_trail -uma_attrs=<my custom attr string>") and in code you would access the target via target.attrs["uma_attrs"]["attr1"], target.attrs["uma_attrs"]["attr2"], ect.?

More or less yes -- maybe we could (re) use "mattr" instead of "uma_attrs" looking at other target kinds -- but in principle that is what I meant.

areusch:
ok for A1 i'm good with named phases and we can modify as necessary. i think the A2.2 solution of directly registering target attrs makes sense to me. is that the direction we're aligned on here?

we can discuss this next week at the community meeting, or if we're in alignment on these two items, i think all that remains is to update the RFC to reflect the discussion here and we can approve/merge.

mbs-octoml:
Apologies for not following the conversation in detail in real time. Here are some thoughts on how we can make sure an UMA-integrated accelerator is also a Collage-supported 'backend'.

 - The registration of patterns will need to support the existing triple of (pattern name, pattern, predicate) since the predicates are necessary to control support based on dtypes, shapes, backend version, etc. No big deal.
 - I'm assuming those triples will continue to end up in either the global pattern table registry, or can be otherwise retrieved by a system like Collage which wishes to bypass the 'eager' UMA partitioning with it's own search. But again no big deal, just need to know where to look.
 - Though not significant to Collage, I assume the order of application of the partitioning patterns matches the registration order?
 - Collage requires external codegen compiler names to be 1:1 with already registered target kinds with the same kind name. It also requires instances of those targets to be provided in the build targets list, even if those instances are nothing other than Target("my_backend") with no extra attributes. But the target kinds may also support additional attributes, and the various transitions into external codegen code have been changed to ensure the matching Target instance has been pushed as the Target.current() so that codegen can retrieve and extract any attributes to guide compilation. I think that matches some of the conversation above, except that the attributes can be fetched by Target.current().get_attr("foo"), but I might have missed the point in that sub-thread.
 - Collage assumes a regular build of an IRModule will respect any existing "Compiler" attributed functions already in the module. I think all that means is that the UMA partitioner should respect existing partitions, but otherwise trigger the appropriate custom downstream compilation, and given the partitioner uses the existing passes I think that should all Just Work.
 - Collage assumes it can do it's partitioning before any other backend-specific passes. I'm assuming however that some of the Relay pass phases mentioned can be before partitioning. If so I'm guessing we'd need to first apply those pre-partitioning phases in deterministic order in the hope that they sensibly compose, then partition using Collage, then run the post-partitioning phases as usual.
 - Collage uses the list of available Targets to guide it's search, but if I understand correctly UMA uses the registration of backends to enforce a fixed partitioning order. Perhaps this suggests the Collage partitioner should be integrated as a user-controlled alternative to the default  'eager' partitoner supplied by UMA (presumably as a loop of the usual Relay MergeComposite/AnnotateTarget/MergeCompilerRegions?/PartitionGraph passes for each backend). That way the user can use the same construct-and-register-backends-of-interest API.
 - I'm surprised by the emphasis on going via TIR. Are we explicitly saying any BYOC integrations which don't need/want to go via TIR don't fall under the UMA integration API? If so that will make Collage/UMA integration harder since Collage would have to account for both UMA-style and original-style integrations.

Thanks,
-m

mbs-octoml:
One more collage/uma overlap aspect: Collage distinguishes 'registered' backends (ie just TargetKinds) from 'activated' backends (ie Target objects in the provided build targets). I think though the proposal here is the act of registration is also activation? I need help understanding how this will look from the user's pov in combination with targets.

PaulPalomeroBernardo:
Thanks @mbs-octoml for this detailed explanation. Being a Collage-supported backend is definitely something we want to achieve for UMA-integrated backends.

> The registration of patterns will need to support the existing triple of (pattern name, pattern, predicate) since the predicates are necessary to control support based on dtypes, shapes, backend version, etc. No big deal.

We will add this to the pattern registration.

> I'm assuming those triples will continue to end up in either the global pattern table registry, or can be otherwise retrieved by a system like Collage which wishes to bypass the 'eager' UMA partitioning with it's own search. But again no big deal, just need to know where to look.

They are registered in the global pattern table registry during backend registration but can also be accessed directly over the backend object if necessary.

> Though not significant to Collage, I assume the order of application of the partitioning patterns matches the registration order?

Correct.

> Collage requires external codegen compiler names to be 1:1 with already registered target kinds with the same kind name. It also requires instances of those targets to be provided in the build targets list, even if those instances are nothing other than Target("my_backend") with no extra attributes. But the target kinds may also support additional attributes, and the various transitions into external codegen code have been changed to ensure the matching Target instance has been pushed as the Target.current() so that codegen can retrieve and extract any attributes to guide compilation. I think that matches some of the conversation above, except that the attributes can be fetched by Target.current().get_attr("foo"), but I might have missed the point in that sub-thread.

I think, this works well. After the backend registration (e.g., `UMABackend.register()`) the target kind, which matches the required codegen compiler name, is available. From there, a target can be created (with or without attributes) and passed to the build target list.

> Collage assumes a regular build of an IRModule will respect any existing "Compiler" attributed functions already in the module. I think all that means is that the UMA partitioner should respect existing partitions, but otherwise trigger the appropriate custom downstream compilation, and given the partitioner uses the existing passes I think that should all Just Work.

I agree.

> Collage assumes it can do it's partitioning before any other backend-specific passes. I'm assuming however that some of the Relay pass phases mentioned can be before partitioning. If so I'm guessing we'd need to first apply those pre-partitioning phases in deterministic order in the hope that they sensibly compose, then partition using Collage, then run the post-partitioning phases as usual.

Yes, we were planning to include a pre-partitioning pass phase. Passes within one pass phase should always be executed in order of their registration.

> Collage uses the list of available Targets to guide it's search, but if I understand correctly UMA uses the registration of backends to enforce a fixed partitioning order. Perhaps this suggests the Collage partitioner should be integrated as a user-controlled alternative to the default 'eager' partitoner supplied by UMA (presumably as a loop of the usual Relay MergeComposite/AnnotateTarget/MergeCompilerRegions?/PartitionGraph passes for each backend). That way the user can use the same construct-and-register-backends-of-interest API.

Currently a user needs to explicitly call `partition()` on the registered backend to perform the usual MergeComposite/AnnotateTarget/MergeCompilerRegions?/PartitionGraph passes plus the relevant relay pass phases (e.g., pre-partitioning).

```
backendA= MyUMABackendA()
backendB= MyUMABackendB()

backendA.register()
backendB.register()
mod = backendA.partition(mod)
mod = backendB.partition(mod)
```

As you described this would eagerly partition the graph depending on the call order of `.partition()`. This would actually give the user the opportunity to skip this partitioning and directly go for the Collage approach. I am not sure if this is the best solution though.

> I'm surprised by the emphasis on going via TIR. Are we explicitly saying any BYOC integrations which don't need/want to go via TIR don't fall under the UMA integration API? If so that will make Collage/UMA integration harder since Collage would have to account for both UMA-style and original-style integrations.

As it is now, they would not fall under the UMA integration API. With UMA we wanted to wrap one specific BYOC integration into an easy-to-use interface and we decided to go with the target hooks via TIR (`relay_to_tir`, `tir_to_runtime`). However, if there is enough motivation we could think about adding `relay_to_runtime` as a second path. This would require greater changes to the current architecture so I don't see it as part of UMA v1 but we can take this into account for future development.

> One more collage/uma overlap aspect: Collage distinguishes 'registered' backends (ie just TargetKinds) from 'activated' backends (ie Target objects in the provided build targets). I think though the proposal here is the act of registration is also activation? I need help understanding how this will look from the user's pov in combination with targets.

There are three steps required to make use of UMA as a user.
1. Create and instantiate a UMA backend `backend = MyUMABackend()`
2. Register the backend `backend.register()`
3. Apply the standard partitioning (might not be necessary with Collage)

`backend.register()` is registering the target kind, a pattern table, and global functions required by the UMA lowering. I think this is more or less equivalent with the Collage 'registration'. Only when the partitioning annotates a subgraph for the backend, it is 'activated'.

PaulPalomeroBernardo:
> ok for A1 i'm good with named phases and we can modify as necessary. i think the A2.2 solution of directly registering target attrs makes sense to me. is that the direction we're aligned on here?
we can discuss this next week at the community meeting, or if we're in alignment on these two items, i think all that remains is to update the RFC to reflect the discussion here and we can approve/merge.

@manupa-arm @areusch I think, we are aligned on this. We decided to go with the enum-based approach for A1 and use A2.2 for UMA v1. I updated the RFC accordingly ([Pass Phases](https://github.com/boschresearch/tvm-rfcs/blob/rfc_uma/rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md#pass-phases), [Target Hooks](https://github.com/boschresearch/tvm-rfcs/blob/rfc_uma/rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md#uma-target-hooks)).

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+|Parameter|Description|
+|---------|-----------|
+|phase|The phase at which the pass is registered.|
+|relay_pass|The relay pass to be registered.|
+
+Example usage:
+```python
+self._register_relay_pass(0, MyPassA)
```
manupa-arm, line 269: nit : might need to fix the example to use PassPhase

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,388 @@
...
+
+A custom accelerator is added by inheriting the `UMABackend`. New elements (e.g., passes) are added using a registration machanism. Below example shows a backend that makes use of all available registration functions.
+```python
+"""UMA backend for the UltraTrail accelerator"""
+
+class UltraTrailBackend(UMABackend):
+    def __init__(self):
+        super(UltraTrailBackend, self).__init__()
+
+        # Relay to Relay function registration
```
manupa-arm, line 121: Should we add an _register_target_attr here as well ?

manupa-arm (commented):
Looks good to me (bar @mbs-octoml comments related to Collage).
Two nits to adjust the text but the design looks good to me.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: Thanks, fixed.

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: Yes, I've added an example.

manupa-arm (approved):
LGTM!

I ll let @areusch and @mbs-octoml  to cover Collage related concerns here ?

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
```diff
@@ -0,0 +1,390 @@
...
+        self._register_tir_pass(PassPhase.TIR_PHASE_0, CodegenGenerateExternCalls())
+
+        # TIR to runtime function registration
+        self._register_codegen(format="c", includes=gen_includes, replace_call_extern=None)
+
+    @property
+    def target_name(self):
+        return "ultra_trail"
+```
+
```
areusch, line 142: could you guys add a brief example of how to use this once you implement this backend class?

areusch (commented):
@PaulPalomeroBernardo @MichaelJKlaiber great! I think we are basically aligned here. I've asked for one more clarification given the discussion around use cases, then I think we can merge.

would you guys still like to discuss this on Wednesday? We have a few different topics, so I'm wondering if we may just need a brief 10 minutes to do an overview of the changes here? i think we can merge this RFC as soon as my one comment is addressed

rfcs/00xx_UMA_Unified_Modular_Accelerator_Interface.md
PaulPalomeroBernardo: This is an override of the abstract property defined in the base class `UMABackend`
```
    @property
    @abstractmethod
    def target_name(self) -> str:
        """Name of the hardware target.

        Returns
        -------
        out : str
            The hardware target name.
        """
        ...
```
It's primarily used internally (e.g., target kind, target related global function names).

areusch: @PaulPalomeroBernardo sorry i meant--can you add an example of how you might call tvm.relay.build() here, just so folks can understand it from user guide perspective?

PaulPalomeroBernardo: @areusch Ahh right, so basically move [this section](https://github.com/boschresearch/tvm-rfcs/blob/rfc_uma/rfcs/0060_UMA_Unified_Modular_Accelerator_Interface.md#umabackend-class) up to the guide-level explanation? I think that makes sense

areusch: yes exactly! i think i'm not seeing that change otherwise i'd hit merge. can you ping again when that's done? and then i think we're good here

MichaelJKlaiber: @areusch, done :)

areusch (approved):

areusch:
thanks @MichaelJKlaiber @PaulPalomeroBernardo @cgerum and others! the RFC is now merged. Please open a tracking issue and link it from this thread for discoverability.

PR #62, opened 2022-03-15
Collage RFC

mbs-octoml:
In rendered form:
https://github.com/mbs-octoml/mbs-tvm-rfcs/blob/mbs-rfcs-collage/rfcs/0062-collage.md


rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+```
+with tvm.transform.PassContext(config={"relay.fallback_device_type": 2, "relay.collage.enable_collage": True}):
+    host_target = tvm.target.Target("llvm")
+    generic_target = tvm.target.Target("cuda", host_target)
+    cutlass_target = tvm.target.Target("cuda -compiler=cutlass", host_target)
+    tensorrt_target = tvm.target.Target("cuda -compiler=tensorrt", host_target)
+    cudnn_target = tvm.target.Target("cuda -libs=cudnn", host_target)
+    cublas_target = tvm.target.Target("cuda -libs=cublas", host_target)
+    targets = [generic_target, cutlass_target, tensorrt_target, cudnn_target, cublas_target]
+    exe = tvm.relay.vm.compile(mod, target=targets)
```
leeexyz, line 159: The targets list here is a kind of `composite` target.  Can we just use it?
https://github.com/apache/tvm/blob/0b6125637058dd080371399e913314bc5f132af3/src/target/target_kind.cc#L406-L407

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+```
+
+(Note that `cudnn` and `cublas` are not yet supported in the 'v2' prototype.)
```
mbaret, line 162: I think further consideration is required around whether cuBLAS should use the partitioning mechanism - it implements just linear algebra functions which would cover a limited amount of Relay.

mbs-octoml: That's actually good -- we want to explore the possibility that a some Relay ops (with some shapes, dtypes) are better implemented by a target-specific library. It's perfectly fine if that support is tiny.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+Pending.
+
+## Success Metrics
+
+1. Collage offers at least a 10% latency improvement for a selection of standard ONNX models and NVIDIA hardware using
+   targets which include the CuDNN and CuBlas libraries, the CUTLASS library (with tuning, via BYOC), the TensorRT
+   compiler (via BYOC), and (obviously!) TVM native.
+2. Collage does not require new per-target or per-model patterns or rules to be implemented independently of the BYOC
+   integrations.
+3. Collage with just the native TWM and a single BYOC toolchain enabled is never worse than using the
```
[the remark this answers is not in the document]
mbs-octoml: Oh darn, now I've gone and leaked it. You didn't see anything.

rfcs/xxxx-collage.md
mbs-octoml: Thanks. Ideally I'd not need to innovate in target handling so if there's an easier path I'm all for it. I do see https://discuss.tvm.apache.org/t/rfc-composite-target/7744/10 which explains the intent is similar to what I need here, but I don't see any downstream handling of it -- do you have any pointers? 

There's also been quite a bit of discussion lately about handling multiple targets which I've not followed, perhaps there's a summary of that?

leeexyz: As far as I know, there is no such handling for the `composite` target yet. I think we do not have a summary either.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+  additive, plus a small launch penalty. However cache effects can dominate measured latency, particularly for 'light'
+  kernels. Thus there may be a **additive error** in the final result:
+
+  > additive_error = measured_latency(collage_partitioning) - sum_{kernel} (estimated_latency(kernel) + penalty)
+
+  The evolutionary search explored by the Collage paper can help here since it uses measured end-to-end model latency as
+  its cost function, but we're deferring that to future work.
+
+- **Limited search space**: Naively exploring all sub-graphs is O(n!), so we need to constrain the search. The easiest
+  approach is just to limit candidate kernels to sub-graphs of just a few operators. This can mean significatly faster
```
cgerum, line 633: This might be very bad on Hardware Accelerators that use extensive Fusion, e.g. EthosU. In these cases it would be probably be preferrable, to limit the search space by using a maximum number of splits of potentially fusible regions. Heuristics might be selected as an attribute of `CombineByPrimitivesFusionRule`.

cgerum (commented):
Thanks for the great proposal. My first comments are all more or less future possibilities. 

I will try to sync up with the UMA proposal, and add a second round of comments ASAP.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+- Classic pattern-based BYOC with `MergeComposite`/`AnnotateTarget`/`PartitionGraph` passes:
+
+```
+     DFPatternFusionRule(pattern1)  ...  DFPatternFusionRule(patternn)
+                |                                    |
+                v                                    v
+      CompositeFusionRule(label1)   ...   CompositeFusionRule(labeln)
+                        \                     /
+                         v                   v
+                            UnionFusionRule
```
mbaret, line 467: What's the UnionFusionRule?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+  retrieving schedules for all the anticipated sub-graphs from the cache.) However the Collage 'v2' prototype collapses
+  all these phases. This lets us lazily explore the implied search graph (nodes = partially rewritten models, edges =
+  selected of sub-graph and toolchain as a candidate kernel, cost = estimated sum of kernel costs plus launch penalties)
+  , and thus only pay the cost of tuning candidate kernels which could possibly influence the final partitioning.
+- **No non-local optimization**: Though Collage can explore the choice of sub-graph and toolchain, it cannot explore any
+  choices which require the arguments and/or result of the sub-graph to be rewritten, or the overall `IRModule` to be
+  changed. Thus Collage **cannot** be used to search over:
+    - choice of layout for arguments/results (may require insertion of layout transforms),
+    - choice of memory scope for arguments/results (may require insertion of device copies),
+    - choice of device on which to host the kernel (ditto),
```
mbaret, line 613: It would be valuable to consider whether in future there would be a viable way to include device exploration in Collage without major design changes. I think there are a number of potentially very significant opportunites to explore with device tuning, particularly in the mobile space.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+
+### Phase 2
+
+Most of the hard work for this phase is carried by the `AllCandidateKernels` implementations of the `FusionRule`s. The
+main driver simply needs to index all the found `CandidateKernels` by their minimum 'inside' `PostDfsIndex`
+for rapid retrieval during the shortest path search.
+
+### Phase 3
+
+We find it most natural to use Dijkstra to find the optimal partitioning. A `SearchState` is:
```
mbaret, line 538: I think the cascader has a similar algorithm as part of the 'proposal generator', but not using Dijkstra's (just a top-down dynamic programming solution). It'd be interesting to work out if that'd be applicable here, and if so whether it's faster or not.

mbaret:
@manupa-arm would be great to get your view on this too

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+- **No non-local optimization**: Though Collage can explore the choice of sub-graph and toolchain, it cannot explore any
+  choices which require the arguments and/or result of the sub-graph to be rewritten. Thus Collage **cannot** be used to
+  search over:
+    - choice of layout for arguments/results (may require insertion of layout transforms),
+    - choice of memory scope for arguments/results (may require insertion of device copies),
+    - choice of device on which to host the kernel (ditto)
+      since all those choices can require changes beyond the candidates sub-graph.
+- the choice of layout for a kernel since any choice other than the model's default must be
+  'corrected' for by the inserted layout transformations. To support this efficiently we'd need to abandon the
+  simple-minded but fast `SubGraph` representation we describe below in favor of something like an EGraph
```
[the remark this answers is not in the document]
mbs-octoml: Yeah agree, there's so much pressure to support search over rewrites it's hard to imagine we won't be going there soon. This proposal is by far not the last word here and it's fine if it's just a small stepping stone.

rfcs/xxxx-collage.md
mbs-octoml: Though it's buried in a TODO I do intend to coalesce small candidates which ended up on the same target before continuing with the rest of the build. However that would be after collage has found what it thinks is the optimal partitioning, which has been measured w.r.t. these possibly too small partitions, and so may be too late. I'd like to experiment with the limits based on empirical measurements against common models and targets before speculating too far ahead here.

rfcs/xxxx-collage.md
mbs-octoml: My hope was the search bloom with Dijkstra would save us from measuring candidates which ultimately wont be in the optimal partitioning. However we not yet ready to measure the bloom on actual largish models.

I spent the last few years working on route optimization for worldwide transport networks so that particular hammer is most familiar to me :-)

mbs-octoml:
Hey all, heads up I'm taking a close look at the 'pass ordering' problem hiding in this RFC. That is, as written and prototyped, CollageFuseOps runs just before the current FuseOps so that it can 'see' rewrites which guide TVM's native fusion rules. However some of those rewrites are target specific, and some are very oriented towards TVM and would likely interfere with existing BYOC patterns and custom lowering function. So we want CollageFuseOps to run both before and after quite a few passes so somethings got to give. I'll see if I can put a clear strawman in place to address this.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+                      \         CompositeFusionRule(label1)  ...  CompositeFusionRule(labeln)
+                       \                    |                        /
+                        v                   v                       v
+                                      UnionFusionRule
+                                            |
+                                            v
+                             CombineByPrimitivesFusionRule (with default TVM primitive rules)
+```
+
+### FusionSpec
```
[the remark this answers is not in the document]
mbs-octoml: I'm warming up to your suggestion (thank you!) to characterize this as 'partitioning' rather than 'fusion', in which I'd rename throughout (FusionSpec -> PartitionSpec, CandidateKernel -> CandidatePartition, CollageFuseOps -> CollagePartition etc). As part of that I could move CollagePartition earlier in the standard pass prefix, and be prepared to run target-specific passes after partitioning but before lowering. This is all more TVM surgery than I was anticipating but I'll give it a go in the prototype and report back.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+  use can be arbitrary. We probably want to i) validate our variance estimator is accurate, ii) choose a percentile
+  slightly above 50% for the estimated candidate kernel latency, and iii) fall back to hard-coded priorities when the
+  measured variance is too high.
+- **Repeated FuseOps**: Some passes (eg `ManifestAlloc`) introduce new calls to primitive function which must be fused
+  and lowered, even though the main work of fusion and lowering has already occurred. We'll need to either
+  retain `FuseOps`, or ensure `CollageFuseOps` retains the efficiency and handling of `FuseOps` when there's no
+  toolchain ambiguity.
+- **Explainability**: It's easy to show the user the final partitioning and estimated times for each kernel, but harder
+  to show why that partitioning won out over all others during search.
+- **Does not subsume `partition_for_<toolchain>`**: We don't have any plan to deprecate the existing patterns of each
```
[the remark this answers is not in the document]
mbs-octoml: I'm a bit stuck on what do about the global rewrite passes many of these BYOC integrations want to include. So I was thinking the UMA effort can solves the API aspect (the user does not need to know about all the ad-doc functions they need to invoke for their intended target) without worrying about the search/turning aspect, Collage can support search/tuning but possibly only for 'well behaved' BYOC integrations, and once both are stable we see where we stand about the UMA partitioning (ie call the usual passes exactly as done today) being subsumed by the Collage partitioning.

Hope that makes sense.

rfcs/xxxx-collage.md
mbs-octoml: Yes, same point by Christoph above. I'd like to, but for now trying to draw the line at not requiring search over rewrites, just partitionings.

rfcs/xxxx-collage.md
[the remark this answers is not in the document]
cgerum: Yeah, the pre-partitioning passes are a real problem. There are two main reasons for these:

1. Simplifying pattern matching e.g. running FoldScaleAxis to handle BatchNorm fusion I assume these would be run before Collage Partitioning
2. Layout transforms, these need to be run pre partitioning for both PatternMatching and to avoid Layout transforms inside the Matched Subgraphs.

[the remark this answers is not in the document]
mbs-octoml: I added a rough section on 'Pass Ordering' to try to get my head around the constraints. I've been pretty focused on a very narrow set of BYOC & library targets so this whole issue caught me off guard.

I think if passes like FoldScaleAxis can be considered 'Safe global' (using my made-up taxonomy from that section) then we can agree to move them into the standard prefix to run before collage. However, in the TVM flow that pass is currently run quite late. So for all of these passes we'd need to:
 - decide all BYOC patterns will be written w.r.t. the canonical form enforced by the pass
 - fix any BYOC patterns which break because of that change
 - check the pass can moved backwards over other passes we wish to leave on the TVM-only side
 I'd been hoping I could stay out of that game.

Another approach is to generalize the patterns so they are pre-canonicalization w.r.t. those passes, then move the passes into the lowering function (on the assumption they are needed downstream). 

And finally, a third approach is to just declare that only some BYOC targets are 'collage ready'. I think that's where we've been all along with the prototyping but just didn't realize it. :-)

Anyway, I'm going to switch back to prototyping mode then come back to the RFC with what I've learned.

rfcs/xxxx-collage.md
mbs-octoml: Note to self to expand this a bit. Search over layout, scope and device placement could technically be done since those choices can be encoded with additional transforms/copies/etc on call arguments and results. However there's little use for that since the estimated model latency won't account for any Relay optimization to remove redundant transforms/copies/etc.

I think global optimization for layout, device placement etc should be done before Collage. TQ reminded me yesterday we don't need to solve all the problems in one pass and composable passes can be added before and after Collage.

rfcs/xxxx-collage.md
mbs-octoml: Marking as 'resolved', and added Appendix G to give more detail.

rfcs/xxxx-collage.md
mbs-octoml: Marking as resolved. I added a sentence in the intro about graph- vs library-style integrations, but let me know if i'm being too obtuse.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+- The [Cascading Scheduler](https://github.com/apache/tvm-rfcs/blob/main/rfcs/0037-arm-ethosu-cascading-scheduler.md) combines i) dynamic-programming
+  to find an optimal grouping of TE sub-expressions, ii) an analytic model of cost to guide the search,
+  and iii) cascading scheduling of the TE sub-expressions so as to reduce memory high-watermark. By contrast
+  Collage i) also uses dynamic-programming, but to find an optimal grouping of Relay sub-expressions, ii)
+  uses measurement to guide the search and iii) assuming the toolchain will 'do its best' with the
+  sub-graph offered to it.
+- The [Universal modular Accelerator Interface](https://github.com/apache/tvm-rfcs/pull/60) proposal
+  adds a layer on top of the existing and separate TVM BYOC, operator strategy, operator scheduling,
+  target-specific passes and target-specific code generation extension points. Collage currently relies
+  only on the global pattern registry and global `relay.ext.<toolchain>` function to integrate with BYOC
```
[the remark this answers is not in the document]
mbs-octoml: Good point. I've moved the TODOs into Appendix D and added an item to connect with the RFC10 machinery in the prototype and write it up if anything interesting comes out.

rfcs/xxxx-collage.md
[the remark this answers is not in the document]
mbs-octoml: Marking as resolved -- added a para about how we could pretty easily do local rewrites but there's little use without the search understanding the overall model latency may dramatically change after further relay passes have optimized based on those rewrites (eg remove redundant device copies).

rfcs/xxxx-collage.md
mbs-octoml: Added section explaining why the 'partitioning' rules should, in a way, predict the downstream 'fusion' rules so as to not waste time partitioning across kernel boundaries while still keeping the maximum depth manageable.

I think we're still a bit speculative here since we're not yet deep into running 'v2' Collage on large models.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+the existing `FuseOps` pass. The new pass effects the invariant:
+
+> All Relay sub-graphs in all global functions which are to be lowered to a kernel are replaced by calls to an inline
+> `"Primitive"` `Function`. Functions which are to be lowered by a BYOC-provided toolchain are given
+> `"Compiler"` and `"global_symbol"` attributes. The bodies of those function may contain calls to inlined
+> `"Composite"` annotated functions to further direct lowering within the kernel.
+
+The `CollageFuseOps` pass proceeds in four phases:
+
+- **Phase 1**: The available `Target`s are scanned to build a list of `FusionSpec`s. Each `FusionSpec` is built from
```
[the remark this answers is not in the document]
mbs-octoml: Rejigged to be an overview without getting bogged down in the specific class names etc. Let me know if you think I need a more gentle overview, i'm probably too close to this now to gauge what's obvious.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+- Classic operator-based BYOC with `AnnotateTarget`/`MergeCompilerRegions`/`PartitionGraph` passes:
+
+```
+      OpPredicateFusionRule
+                |
+                v
+   CombineByPrimitivesFusionRule (with join anything primitive rule)
+```
+
+- Classic pattern-based BYOC with `MergeComposite`/`AnnotateTarget`/`PartitionGraph` passes:
```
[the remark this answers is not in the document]
mbs-octoml: Will account for in above comment.

rfcs/xxxx-collage.md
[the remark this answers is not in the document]
mbs-octoml: I ended up covering this in Appendix C.

For the 'I just want to partition for BYOC x' case I think that's where UMA can take away all the manual calls to ad-hoc functions.
Eventually UMA could defer to Collage to actually do the partitioning, but only after we're confident there's no seach overhead if the user want to do things manually. I think we can tackle that later.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+  measurement of Collage `CandidateKernel`s with minor changes.
+- (Internal to OctoML) We need an implementation connecting to the internal OctoML kernel tuning workflow and production
+  cache. Ideally this would be the same implementation as for the MetaSchedule system.
+- Collage converts measured 50th %ile latencies to costs in seconds. We may need to consider taking a slightly higher
+  %ile to be more robust against variance on small kernels. We need to validate the estimated variance reflects true
+  variance.
+- For TVM-native targets, we would like the `Estimate` call to perform any TVM tuning required for a novel candidate
+  kernel.
+
+### Easier Library Integration
```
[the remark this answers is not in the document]
mbs-octoml: Though moved into Appendix B I 'd still like to keep a placeholder for this issue since it motivates how I'm proposing to support "Composite" functions fairly generically. I'll be happy to delete the 'One possibility is...' section once you've got something to replace it.

rfcs/xxxx-collage.md
mbs-octoml: Will cover in comment above.

rfcs/xxxx-collage.md
mbs-octoml: Going to resolve. Expanded on the difficulty (yes can rewrite, but unlikely to help since Collage cost won't account for downstream global optimizations enabled by such.)

rfcs/xxxx-collage.md
mbs-octoml: Going to resolve but we can adjust search algorithm  once we come up against the reality of search on large models.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,833 @@
...
+- `OpPredicateFusionRule`: Given an attribute name, produces a candidate for every call to a primitive Relay operator
+  where the operator has predicate bound to that attribute which returns true given the call sub-expression. Generally
+  this will result in a singleton sub-graph containing only the call, but it may pull in constant arguments to the call
+  should they be required. This is the foundation for operator-based BYOC integrations, though we should consider
+  retiring this mechanism in favor of pattern-based alone.
+- `OpCallByKindFusionRule`: Uses the `"TOpPattern"` attribute provided for every Relay operator to produce a candidate
+  for every call to a 'fusable Relay operator'. This can be used as the foundation for generic fusion patterns which
+  work over all Relay operators with particular properties (elementwise, broadcast, injective, reductive, anchor).
+
+The combinator rules implemented so far:
```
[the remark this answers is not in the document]
mbs-octoml: Yeah I glossed over this pretty quickly, sorry! I've reworked this section a little bit, but it is still just a copy/paste from the code comment. Leaving this comment open till I can come back and expand this whole section.

mbs-octoml:
Bumped version to 0.8 based on extensive reworking due to excellent comments above -- thanks!.

I still need to expand the PartitionRule section, should be ready tomorrow.

rfcs/xxxx-collage.md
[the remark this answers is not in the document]
mbs-octoml: Ok put a bit more love into this section. Let me know if that helps.

mbs-octoml:
PTAL all, thanks.

jwfromm (approved):
Awesome RFC, Mark! It seems like all comments so far have been well addressed. If there are no additional comments in the next few days, we'll officially approve and merge this RFC.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+- **Phase 1**: The available `Target`s are scanned to build a list of rules describing how to find possible partitions (
+  see `PartitionSpec` and `PartitionRule` below). Depending on the `Target` the rules may incorporate entries from the
+  BYOC pattern table. (The remaining phases execute on each global function separately.)
+- **Phase 2**: A dataflow graph is constructed for the global function (which is just an `IndexedGraph<Expr>`). The
+  available rules from phase 1 are evaluated on the dataflow graph to yield a (possibly overlapping) set of candidate
+  partitions for each target (see `CandidatePartition` below). Each candidate efficiently describes a sub-graph of the
+  global function's body without the need to construct any new expressions (see `SubGraph` below).
+- **Phase 3**: A shortest path is found in the following (implicit and lazily constructed) search graph:
+    - Search Nodes: The set of dataflow nodes which have already been assigned to a candidate partition in all paths to
```
manupa-arm, line 261: [Clarification] are these intersection or union of all dataflow nodes that have been assigned by all paths to the node ?
A set notation might really help here to disambiguate

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+      the node.
+    - Search Edge X->Y: A candidate partition can be applied to node X to give node Y. The candidate is disjoint from
```
manupa-arm, line 263: Im not sure I follow what is mean by "give" here ? I assume between the search nodes candidate partitions change.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+The `CollagePartitioner` pass has four phases:
+
+- **Phase 1**: The available `Target`s are scanned to build a list of rules describing how to find possible partitions (
+  see `PartitionSpec` and `PartitionRule` below). Depending on the `Target` the rules may incorporate entries from the
+  BYOC pattern table. (The remaining phases execute on each global function separately.)
+- **Phase 2**: A dataflow graph is constructed for the global function (which is just an `IndexedGraph<Expr>`). The
+  available rules from phase 1 are evaluated on the dataflow graph to yield a (possibly overlapping) set of candidate
+  partitions for each target (see `CandidatePartition` below). Each candidate efficiently describes a sub-graph of the
+  global function's body without the need to construct any new expressions (see `SubGraph` below).
+- **Phase 3**: A shortest path is found in the following (implicit and lazily constructed) search graph:
```
manupa-arm, line 260: If we are modelling the search as a path-finding problem, I think we need to define the start and the end (success criteria) too.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+    - Search Nodes: The set of dataflow nodes which have already been assigned to a candidate partition in all paths to
+      the node.
+    - Search Edge X->Y: A candidate partition can be applied to node X to give node Y. The candidate is disjoint from
+      all dataflow nodes already assigned in X. To avoid an unnecessary search space explosion the candidate must also
+      include the next yet-to-be-assigned dataflow node in X.
+    - Edge cost: The estimated latency of the candidate partition, plus a partition transition penalty. Note that though
```
manupa-arm, line 266: [Clarification] why is estimated latency associated with an edge ? (partition transition penalty make sense)

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+      we need to be able to extract the candidate's sub-graph in order to build a function representing the candidate,
+      we do not yet need to partition the overall function body expression.
+
+  Other search algorithms are certainly possible, eg the paper uses an evolutionary search to refine the partitioning
+  found by the dynamic-programming search. We can easily abstract away the search interface to support multiple
```
manupa-arm, line 271: If this is an aspiration, we need to define a bit higher-level interface where path-finding formulation becomes a specific implementation.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+  implementations in the future.
+- **Phase 4**: The function body is partitioned according to the candidate kernels on the shortest path.
```
manupa-arm, line 273: A small example would really help here :)

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+The candidates are allowed to overlap, and ultimately it is the job of the Collage searcher to find a selection of
+candidates which cover the whole Relay expression without overlap. There may be many thousands of candidates in flight
+during the Collage search.
+
+We currently have three distinct flavors of partitions:
+
+- For pattern-based BYOC integrations, individual `DFPattern`s are used to select the `"Composite"` functions to
+  offload, and those are grouped into a `"Primitive"` Relay function with a `"Compiler"` attribute.
+- For operator-based BYOC integrations, per-operator predicates indicate operators to offload, and again those are
```
manupa-arm, line 368: Can we do a simple utility and covert these into patterns as well in a very early layer in Collage ? -- so that we don't have to maintain yet another rule ?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+  call sub-expression. Generally this will result in a singleton sub-graph containing only the call, but it may also
+  pull in constant arguments to the call should they be required. Used to bring BYOC operator predicates into the
+  Collage framework.
+- `OpCallByKindPartitionRule`: Uses the `"TOpPattern"` attribute provided for every Relay operator to produce a
+  candidate for every call to a 'fusable Relay operator'. Used to select the operators which `FuseOps` will consider
+  parts of kernels.
+
+The combinator rules are:
+
+- `CompositePartitionRule`: Indicates all sub-candidates matched by the sub-rule should be wrapped by a `"Composite"`
```
manupa-arm, line 421: How about we use base-candidates and combinator-candidates rather than introducing another term "sub-candidates" ?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+  existing TVM `FuseOps` pass, though: i) all candidates are found rather than just the largest, ii) the starting
+  set of candidates can be provided by any other rule, and iii) we rely on `SubGraph` validity checking to weed out
+  infeasible candidates.
+- `OnlyValidPartitionRule`: Given a `SubGraphConfig`, ignores candidates with 'invalid' sub-graphs. Used to limit the
+  maximum candidate depth, the number of independent outputs, and whether intermediate 'taps' are allowed.
+- `HostPartitionRule`: Produces candidates for all Relay expressions which could be
+  'left behind' for execution by the host (eg on the VM). This rule lets us simplify the overall Collage search
+  algorithm.
+
+Here are some typical ways to combine `PartitionRules` for different partition flavors:
```
manupa-arm, line 441: How are these rules populated ?

Can they be "registered" to avoid unnecessary search ?
The registration could be allowed or a dis-allowed list.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+    OnlyValidPartitionRule
+      UnionPartitionRule
+        DFPatternPartitionRule(pattern1)
+                     :
+        DFPatternPartitionRule(patternn)
+  ```
+
+### PartitionSpec
+
+A `PartitionSpec` pairs a a `PartitionRule` with one or more `Target`s.
```
manupa-arm, line 519: I suppose this is related to the above question.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+Most of the hard work for this phase is carried by the `AllCandidates` implementations of the `PartitionRule`s. The main
+driver simply needs to index all the found `CandidatePartitions` by their minimum 'inside' `PostDfsIndex`
+for rapid retrieval during the shortest path search.
+
+### Phase 3
+
+We find it most natural to use Dijkstra to find the optimal partitioning. A `SearchState` is:
+
+- An `IndexSet` of the dataflow nodes already 'covered' by candidates on the best path to this state. This is the
```
manupa-arm, line 559: This point somewhat contradicts/mis-defined above as the node in the Search, where it was defined as all nodes in all paths and here it is best path -- which kind of make sense.

Is the SearchState == node in the search graph ?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+The starting state has no covered nodes. The final state has all nodes covered.
+
+When expanding a state we could choose any `CandidatePartition` collected from phase 2 provided it doesn't overlap with
+the state's covered set. However, a search path applying candidates C then D is equivalent to one applying D then C, so
+we only consider candidates which intersect the next yet-to-be-covered dataflow node. For each such candidate we use
+the `CostEstimator` (with it's assumed cache) to get the candidate's cost, build the successor state, and 'relax' the
+successor state in the usual way. (See Appendix E for more details on `CostEstimator`.)
+
+The `HostPartitionRule` is used to allow some dataflow nodes to be 'left behind' for execution by the host.
```
manupa-arm, line 573: Why host treated differently?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+### Phase 4
+
+The overall Relay expression is partitioned over all the `CandidatePartition`s on the shortest path 'in parallel'. Since
```
manupa-arm, line 577: nit : maybe use least-cost path ?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+all the candidates are expressed using `SubGraph`s w.r.t. the original dataflow graph, we must be careful not to
+invalidate yet-to-be-partitioned candidates as we go. Working backwards in dataflow order avoids this problem.
+
+## Known Limitations
+
+- **Some BYOC boilerplate changes required**: TVM's current BYOC integration API only requires the 'lowering/codegen'
+  function to be registered to a well-known global function name. Everything else is up to the BYOC author.
+    - Collage requires pattern-based BYOC integrations to register their patterns in the global pattern table. Some BYOC
+      integrations use the table but many do not, but it's an easy fix.
+    - Collage requires the BYOC lowering function to yield a valid `runtime::Module` without requiring any additional
```
manupa-arm, line 587: Will Collage work with Target Hooks ?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+- **Explainability**: It's easy to show the user the final partitioning and estimated times for each kernel, but harder
+  to show why that partitioning won out over all others during search.
+
+- **Does not subsume `partition_for_<toolchain>`**: We don't have any plan to deprecate the existing patterns of each
+  BYOC integration supplying a `partiion_for_<toolchain>` function. If the user has a specific toolchain in mind then
+  making the partition explicit enjoys both faster compilation and can incorporate global optimization passes which
+  Collage cannot currently account for (eg enforcing a particular layout).
+
+## Appendix A: Differences between the paper's implementation and this proposal
```
manupa-arm, line 671: Note to myself : reviewed up to here.

manupa-arm (changes requested):
Thanks @mbs-octoml for the detailed RFC.

I did a first pass. There is lot to unpack here.

I think we would need to follow the template for this RFC : https://github.com/apache/tvm-rfcs/blob/main/0000-template.md

This will allow the readers to focus on UX/UI implications and technical deeper-details separately. I suppose just moving the content around to match the template is what's required here.

I have left comments. questions -- mostly to understand what is being proposed.

I have few broad questions : 

1.) How can a user export the search done by collage ? i.e. similar to loading tuning logs where ApplyBestHistory is done.

2.) Is there a model being developed to estimate latency of the chosen partitioning? or is it always need to run on the actual hardware ? -- it would be great explore if we can abstract this at that at level so a user could plugin a Relay/TIR based performance model to avoid lengthy "tuning" time.

manupa-arm:
> 1.) How can a user export the search done by collage ? i.e. similar to loading tuning logs where ApplyBestHistory is done.

Having thought about this more, If there can be a way to define a PartitionSpec that is tied with DFS IndexSet that could be exported and imported, that might work out.

I guess my question is there a way to reproduce the partitioning w/o needing to undergo a tuning phase everytime.

If that is so from a design perspective, it would be better to decouple the collage tuner and collage partitioner, one to produce exportable paritioning spec and what to consume such a partitioning spec. 

Let me know if you have any thoughts around this :) @mbaret @mbs-octoml

rfcs/xxxx-collage.md
mbs-octoml: Thanks Manupa, I'll hold off working through these until you've finished a first pass.

mbs-octoml:
> Is there a model being developed to estimate latency of the chosen partitioning? 

No. As far as Collage is concerned it just calls the abstract CostEstimator::Estimate interface for each candidate partition, and can remain ignorant as to where those costs come from. In the prototype it is hard coded to tune, build and run locally to help us get going. Here at OctoML we'll need  to instantiate the interface to connect to our production tuning/running/caching systems. In principle the interface could also be instantiated to some analytical models a la the cascading schedule, though we're not planning to build one.

I'm thinking the CostEstimator object can just be passed into the CollagePartitioner to leave this extension point open.

> How can a user export the search done by collage ?

In an early draft I had support for that. Basically it just needs some materialization of the 'optimal' partitioning as an  Array<CandidatePartition>. The method CandidatePartition::ParallelRewrite can take that array and rewrite the whole expression, exactly as the Collage pass would have done. So splitting between search and rewrite is pretty easy.

But I ended up dropping it all in favor of just relying on the CostEstimator to cache all measurements (which it needs to do anyway given all the sharing opportunities). Firstly, it's not yet clear if there's any significant compile time advantage to bypassing the collage search if every candidate partition to estimate results in a cache hit. I figured I'd at least measure that before adding a fix. But secondly, if someone (the service, the user) is going to go to the trouble of caching the optimal partitioning for a particular (model, targets) pair, why not just cache the built artifact directly and skip all the bother?

However let me know if I over simplified and I can add that part back.

> think we would need to follow the template

Ok.

rfcs/xxxx-collage.md
mbs-octoml: It will need to, yes, I've just not gotten around to figuring out the encoding so that the hooks make the correct redirection, and how to tell  the difference between using Compiler=<toolchain> vs Target=.... But I think it's just leg work.

rfcs/xxxx-collage.md
mbs-octoml: As the search proceeds we may encounter Relay subexpressions (eg an if) which no BYOC backend is interested in and which do not fit into TVM's notion of what can go in a kernel. At first I handled all of this inside the search algorithm, but then realized a 'HostPartitionRule' can move that special casing out of the core algorithm and into a very trivial rule implementation.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+3. Uses the existing convention for `"Primitive"`, `"Composite"` and `"Compiler"` attributes on Relay `Function`s to
+   encode partitioning choices.
+4. Does not treat library integrations (eg for `CuDnn`) differently from toolchain integrations (eg for `TensorRT`). See
+   Appendix B for a sketch of the issues.
+5. Supports all of Relay.
+6. Is implemented almost entirely in C++.
+
+However:
+
+6. We have only re-implemented the 'op-level' dynamic-programming based search strategy from the paper. Though the paper
```
manupa-arm, line 730: It might be worth clarifying that here op-level will consider sub-graphs of operators as well. Having read the paper, I can understand what is meant here, however, it has the tendency to confuse the reader to assume the current implementation only works single relay op-level, which I think is not true.

Or Am I also confused :p ?

manupa-arm (commented):
Alright I completed the pass.

The work being proposed is really interesting. kudos to all involved in the effort.

manupa-arm:
> No. As far as Collage is concerned it just calls the abstract CostEstimator::Estimate interface for each candidate partition, and can remain ignorant as to where those costs come from. In the prototype it is hard coded to tune, build and run locally to help us get going. Here at OctoML we'll need to instantiate the interface to connect to our production tuning/running/caching systems. In principle the interface could also be instantiated to some analytical model a la the cascading scheduler, though we're not planning to build one.

I'm thinking the CostEstimator object can just be passed into the CollagePartitioner to leave this extension point open.

Ack.

>  I figured I'd at least measure that before adding a fix. But secondly, if someone (the service, the user) is going to go to the trouble of caching the optimal partitioning for a particular (model, targets) pair, why not just cache the built artifact directly and skip all the bother? However let me know if I over simplified and I can add that part back

This might enable bring us to down to get the cost estimation but it will still explore the combinatorial partitioning possibilities of subgraphs x backends, right ? (lmk, if I am wrong here).

I suppose having the abstraction might help a user to play around with what Collage generates as a base. I would consider that to be a useful feature -- even guide the next generation algorithm we might want to put there. In the absense of that, the user need to tweak the CostEstimator as a proxy.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+> *Collage: Automated Integration of Deep Learning Backends*  
+> Byungsoo Jeon, Sunghyun Park, Peiyuan Liao, Sheng Xu, Tianqi Chen, Zhihao Jia
+
+(See Appendix A for a comparison of this proposal and the paper's implementation. See Appendix D for TODO items in the '
+v2' prototype.)
+
+This tuning approach contrasts with TVM's existing "greedy" and "manual" approaches to partitioning:
+
+- Greedy: Currently only the largest possible supported sub-graphs are used for partitions, irrespective of their
+  execution time. With Collage many more candidate sub-graphs are explored, and it is possible for two smaller
```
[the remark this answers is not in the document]
mbs-octoml: This terminology is favored by our PMs since it conveys a useful intuition for how both FuseOps and BYOC approachs tend to gobble up sub-graphs without concert for what comes later. I understand MergeCompilerRegions. A small lie saves a thousand truths?

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+  sub-graphs to yield better overall latency than one large sub-graph if they mix toolchains.
+- Manual: Currently the TVM user must commit to a BYOC toolchain and invoke the corresponding
+  `partition_for_<toolchain>` function before the main TVM compilation flow begins. With Collage the choice of toolchain
+  can be automated based on measured latency. Collage will also explore mixing and matching between multiple BYOC
+  toolchains as well as TVM's native backend.
+
+When Collage is enabled it subsumes the existing `MergeComposite`/`AnnotateTarget`/`MergeCompilerRegions`/
```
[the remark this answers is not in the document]
mbs-octoml: Though I've not follow every section to the letter I did rearrange to match most of the template.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+`PartitionGraph` passes embedded within each `partition_for_<toolchain>` function with a single new
+`CollagePartitioner` pass. The pass is guided by the list of available `Target`s and three existing sources:
+
+1. The `"TOpPattern"` attributes provided for every Relay operator and used by TVM's built-in `FuseOps`.
+2. The BYOC `"target.<toolchain>"` operator predicates provided for some operator/toolchain pairs by
+   'operator-based' BYOC integrations.
+3. The BYOC operator pattern/predicates (usually) registered in the pattern table by 'pattern-based' BYOC integrations.
```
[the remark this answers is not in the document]
mbs-octoml: Left a TODO to revisit once the TRT conversion is in main. Fingers crossed!

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+  execution time. With Collage many more candidate sub-graphs are explored, and it is possible for two smaller
+  sub-graphs to yield better overall latency than one large sub-graph if they mix toolchains.
+- Manual: Currently the TVM user must commit to a BYOC toolchain and invoke the corresponding
+  `partition_for_<toolchain>` function before the main TVM compilation flow begins. With Collage the choice of toolchain
+  can be automated based on measured latency. Collage will also explore mixing and matching between multiple BYOC
+  toolchains as well as TVM's native backend.
+
+When Collage is enabled it subsumes the existing `MergeComposite`/`AnnotateTarget`/`MergeCompilerRegions`/
+`PartitionGraph` passes embedded within each `partition_for_<toolchain>` function with a single new
+`CollagePartitioner` pass. The pass is guided by the list of available `Target`s and three existing sources:
```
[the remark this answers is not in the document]
mbs-octoml: Added line and forward reference.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+
+1. The `"TOpPattern"` attributes provided for every Relay operator and used by TVM's built-in `FuseOps`.
+2. The BYOC `"target.<toolchain>"` operator predicates provided for some operator/toolchain pairs by
+   'operator-based' BYOC integrations.
+3. The BYOC operator pattern/predicates (usually) registered in the pattern table by 'pattern-based' BYOC integrations.
+
+Only some boilerplate aspects of existing BYOC integrations need to be adjusted to support Collage (and we will make
```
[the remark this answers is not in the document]
mbs-octoml: Done.

rfcs/xxxx-collage.md
mbs-octoml: Yes, made that a bit clearer. The covered set is actually all nodes included in partitions on any path to the state, of which the best is one. Updated text.

rfcs/xxxx-collage.md
mbs-octoml: yes, done.

rfcs/xxxx-collage.md
mbs-octoml: Added a TODO for this.

rfcs/xxxx-collage.md
mbs-octoml: Added an extra sentence here.

mbs-octoml:
PTAl, bumped to 0.81, thank you @manupa-arm,  @mbaret and @cgerum, you've been a great help.

manupa-arm:
Hi @mbs-octoml ,

I think you may have missed the comments that is hidden by Github (happens to me all the time :) ).

mbs-octoml:
> I think you may have missed the comments that is hidden by Github
D'oh! 
Hidden resolved conversations != Hidden because, you know, looks ugly if the page is too long, or something.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,987 @@
...
+See Appendix H for some frequently asked questions.
+
+## Success Metrics
+
+1. Collage offers at least a 10% latency improvement for a selection of standard ONNX models and NVIDIA hardware using
+   targets which include the CuDNN and CuBlas libraries, the CUTLASS library (with tuning, via BYOC), the TensorRT
+   compiler (via BYOC), and (obviously!) TVM native.
+2. Collage does not require new per-target or per-model patterns or rules to be implemented independently of the BYOC
+   integrations.
+3. Collage with a `Target` list enabling just one BYOC toolchain is never worse than using the the existing
```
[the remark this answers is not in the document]
mbs-octoml: Added a 'without loss of generality' comment.

rfcs/xxxx-collage.md
mbs-octoml: Cleaned those comments up, thanks.

rfcs/xxxx-collage.md
mbs-octoml: done

rfcs/xxxx-collage.md
mbs-octoml: I'm still following the paper where this is considered a 'kernel launch penalty', which is uniquely determined by the partition, which is captured in the search graph edges. I don't have any experience with this aspect, and don't yet know whether i) we'll need to measure these penalties, ii) their impact on optimal placements and iii) whether they should be determined by the pair of successive partitions rather than just one partition. So I'm leaving this a bit vague.

rfcs/xxxx-collage.md
mbs-octoml: I think that's best deferred till we have the second example. At this point however I think IRModule -> Array<CandidatePartition> looks to be about as specific as we can get.

rfcs/xxxx-collage.md
mbs-octoml: I'd take those to mean 'candidates produced by a {base,combinator} rule' which isn't the intent of the term. In particular, 'combinator' rules need to be nested, so the 'sub-candidates' of one combinator may be the result of running some nested combinator. Let me add a more explicit definition of that term to see if that helps, otherwise the text gets pretty unweildy.

rfcs/xxxx-collage.md
mbs-octoml: Added a forward reference to the 'Phase 1' section. Yeah, I'd like to allow the Target to directly specify the rule to use by a "partition_rule" attribute, but since that's not a simple atomic scalar I'm not sure how the ergonomics will work out. Otherwise there's logic to, effectively, 'import' the existing BYOC patterns etc by inspecting the global pattern table (and UMA could help make this a little more structured I think).

rfcs/xxxx-collage.md
mbs-octoml: As soon as https://github.com/apache/tvm/pull/10759 is in I'll terminate all of this special casing (with extreme prejudice!).

rfcs/xxxx-collage.md
mbs-octoml: It's really just here for book-keeping, all the real action takes place with PartitionRules.

rfcs/xxxx-collage.md
mbs-octoml: Ok, thanks for pushing me -- I added some pictures for some of the key concepts using (mini) MNIST as an example throughout. Hopefully that helps!

mbs-octoml:
PTAL. Barring more github incompetence on my part I think I covered your comments @manupa-arm. Thanks again.

rfcs/xxxx-collage.md
```diff
@@ -0,0 +1,863 @@
...
+  choices which require the arguments and/or result of the sub-graph to be rewritten, or the overall `IRModule` to be
+  changed. Thus Collage **cannot** be used to search over:
+    - choice of layout for arguments/results (may require insertion of layout transforms),
+    - choice of memory scope for arguments/results (may require insertion of device copies),
+    - choice of device on which to host the kernel (ditto),
+    - choice of quantization scheme,
+  
+  since all those choices can require changes beyond the candidate's sub-graph. To support this efficiently we'd need
+  to abandon the simple-minded but fast `SubGraph` representation we describe here in favor of something like an EGraph
+  representation, which seems like a very large change for TVM.
```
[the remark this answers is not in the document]
mbs-octoml: I think your 'Plan' matches my 'CandidatePartition', which has a 'SubGraph'. I think it would be pretty straightforward for SubGraphs to be able to represent things like 'oh, by the way, this partition is on device X so needs some device copies on the way in and out'. But I remain skeptical that will do what we want unless we somehow account for post-partitioning optimizations to remove redundant copies etc.

jwfromm:
@manupa-arm can you take a look at Mark's revisions and let us know if this RFC now looks good to you?

rfcs/0062-collage.md
```diff
@@ -0,0 +1,1041 @@
...
+    tensorrt_target = tvm.target.Target("cuda -compiler=tensorrt", host_target)
+    cudnn_target = tvm.target.Target("cuda -compiler=cudnn", host_target)
+    cublas_target = tvm.target.Target("cuda -compiler=cublas", host_target)
+    targets = [generic_target, cutlass_target, tensorrt_target, cudnn_target, cublas_target]
+    exe = tvm.relay.vm.compile(mod, target=targets)
+```
+
+(Note that `cudnn` and `cublas` are not yet supported in the 'v2' prototype, see Appendix B.)
+
+After the `CollagePartitioner` pass, the intermediate `"main"` global function could resemble the following
```
manupa-arm, line 174: One small question -- would we be able to export IRModule after the pass and make sure if we pass that IRModule to relay.build(...) the compilation works as if it were tuned by Collage ?

manupa-arm (approved):
Thanks @mbs-octoml for detailed explanation as figures.

I do have one more small question, hope that could be done by default.

rfcs/0062-collage.md
mbs-octoml: Yes! Though I had to fix a few ill-behaved passes to make it work, it is possible to take the partitioned IRModule and call build on it with Collage disabled, and all the primitive/compiler annotations are respected. I use this to build test cases.

However if Collage is enabled then it will attempt to re-tune the 'TVM' partitions, since there's currently nothing in the Collage output IRModule to explicitly encode those partitions (eg by wrapping in a primitive function as is done for the BYOC partitions). I'd prefer a clearer way to encode partitions but don't want to introduce unnecessary churn on existing BYOC machinery.

jwfromm:
Thank you Mark for this excellent RFC, and thank you everyone for your feedback. This RFC is now merged. I'm really excited to see the results of this project.

areusch:
We discussed this RFC at the TVM Community meeting. Here are some notes beyond the [presentation content](https://docs.google.com/presentation/d/1sRz8hVy_619VmbSRwJPGmcMfgBzaqMOAVaVNxlSgGFM/edit#slide=id.p):
- @mbs-octoml notes that this RFC isn't particularly set in stone and it may grow/change as this effort proceeds. It's a bit speculative right now.
- note: "on TVM" means running regular Relay operators on TVM
- in the autotuning case you'd do autotuning on each path from beginning to end?
  - currently we are doing autotuning on the fly. for autotvm this isn't too bad. for new metaschedule, every candidate kernel will be treated as its own tuning task, so there are a lot more candidate kernels to explore (tvm's present FuseOps pass is greedy and always combines kernels, but MetaScheduler does not necessarily
- will we still use the cache mechanism to avoid re-tuning identical operators? 
  - yes, and we hope to get a good hit rate from the cache.
- by cache do you mean something you've created yourself? or is it an online service or tuning log on github?
  - @mbs-octoml: there's an abstract cost-estimator interface (given an IRModule, return a double). in prototype, there's only 1 instantiation of such interface which runs using the TVM runners and the standard benchmarking machinery in Python. Inside OctoML, we'll have different instantiations of this interface which will consult our internal infrastructure. the caching in the prototype works by a naïve in-memory cache coupled with a bit of a hack to populate it with standard AutoTVM tuning records. 
- @manupa-arm: after the candidate partitioning is searched, will collage consider merging adjacent offloaded subgraphs to the same compiler?
  - yes, there is a cleanup pass that will merge adjacent subgraphs. this could happen due to heuristics in how partitions are chosen. this is a little bit like how MergeComplierRegions works now. generally speaking we expect this to be a bit orthogonal to the results of the search, because we expect transition latency to be small. if we do see a difference, it means Collage was not trying to offload the right size of subgraphs.
- if we identify
  - when i say partitioning, what do I mean? can you explore serial vs parallel? what about notions of inlining (e.g. a reshape shared 38 times)? should I be exploring doing the reshape and sharing the result? none of this is being done right now.
- the changes needed are just to expand the search model and the way we compute end-to-end measurement latency?
  - the constraint we need to stay in here is that the transform you want to try is local--the changes can be confined to a subgraph and you can time that subgraph. beyond that, you get outside the bounds of dynamic programming.

mbs-octoml:
Thanks @areusch.
> (tvm's present FuseOps pass is greedy and always combines kernels, but MetaScheduler does not necessarily
Just a small clarrification: TVM's FuseOps is greedy but CollagePartitioner may require measuring (and thus tuning) many more TVM partitions/kernels. For autotvm this isn't a big deal since autotvm task extraction only considers single ops. But for metaschedule whatever is presented as a Primitive is considered the unit of tuning.

PR #66, opened 2022-04-08
[RFC] A proposed update to the Docker images ci_* tag pattern

leandron:
This RFC proposes a new format to keep our Docker images used in CI, moving away from our current incremental numbering to a more meaningful format that contains a timestamp and the latest hash from the repository used to generate such images.

cc @areusch @driazati @Mousius @konturn @junrushao1994 @masahi @gromero @mehrdadh for visibility and reviews

### DIFF
--- a/rfcs/0000-docker-image-tags.md
+++ b/rfcs/0000-docker-image-tags.md
@@ -1,0 +1,52 @@
+- Feature Name: docker_image_tag_format
+- Start Date: 2022-04-08
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+
+Defines a format for TVM CI Docker images tag.
+
+# Motivation
+
+In the current format used to identify our Docker images, it is impossible to link back a given version of the image to which version of the repository that represents.
+
+# Guide-level explanation
+
+For years we have been using a patern to version our CI Docker images, which is composed by v0.INCREMENTAL_NUMBER, e.g. tlcpack/ci_cpu:v0.80. Every time an image is updated, we bump the minor number, just as a way to release a new version.
+
+As the project is growing and given we recently have been improving the Docker images building process, it is time to consider using a more meaningful tagging scheme, so that it is easier to identify what is included in the images being used in our CI.
+
+When looking/using a Docker image, usually we would be interested in:
+
+1. How long ago was this image generated?
+2. What is the last change added in the current images being used in production?
+
+None of these questions can be answered by our current numbering scheme, and to get that information we would usually need to do some digging and inspecting the image to see what is in there.
+
+# Reference-level explanation
+
+As an improvement to the current situation, this RFC proposes the adption of a tagging scheme currently used in the `tlcpackstaging` (https://hub.docker.com/u/tlcpackstaging) repository, composed by:
+
+* a timestamp YYYYMMDD-HHMMSS
+* the last short git hash added in that image
+
+One example would be tlcpackstaging/ci_arm:20220201-115323-2af42ba8e. This tells us when this image was generated and up to which point in the repository is included.
+
+The advantage with the proposed scheme, is that we can just look back on our own repository to discover which changes to the Docker setup scripts (under docker/install/* in the TVM repository) as a way to check whether a given change is expected to be in the image.
+
+# Drawbacks
+
+Pointing to "the previous image" is obvious today, because it is just "the current version minus one". To revert the image in the proposed tagging scheme, we will need to adapt this process and use git history to discover which was the previous version in the repository. However, reverting images is quite rare so it should be something the will consume lots of time very often.
+
+# Rationale and alternatives
+
+Many other formats could be used to describe the tag, such as `git describe`, however, it lacks the timestamp, which is a quite useful piece of information to give an impression of how long ago an image was updated.
+
+# Prior art
+
+The proposed format is used in our automated daily Docker image rebuild jobs at https://ci.tlcpack.ai/job/docker-images-ci/.
+
+# Future possibilities
+
+As this is already in discussion, there is an opportunity, given we evaluate properly the impact in the current CI job runs, to rebuild Docker images for each PR.


rfcs/0000-docker-image-tags.md
```diff
@@ -0,0 +1,52 @@
...
+* a timestamp YYYYMMDD-HHMMSS
+* the last short git hash added in that image
+
+One example would be tlcpackstaging/ci_arm:20220201-115323-2af42ba8e. This tells us when this image was generated and up to which point in the repository is included.
+
+The advantage with the proposed scheme, is that we can just look back on our own repository to discover which changes to the Docker setup scripts (under docker/install/* in the TVM repository) as a way to check whether a given change is expected to be in the image.
+
+# Drawbacks
+
+Pointing to "the previous image" is obvious today, because it is just "the current version minus one". To revert the image in the proposed tagging scheme, we will need to adapt this process and use git history to discover which was the previous version in the repository. However, reverting images is quite rare so it should be something the will consume lots of time very often.
```
Lunderberg, line 40: This might be a benefit, rather than a drawback.  In the current naming scheme, if a rollback has occurred, then the previous image is no longer the current version minus one.  This could cause an issue in the hypothetical situation below.

1. Docker image is updated from stable `N` to updated `N+1`.
2. Version `N+1` requires rollback, version number is decremented to stable `N`.
3. Docker image is updated from stable `N` to updated `N+2`.
4. Version `N+2` also requires rollback, version number is erroneously decremented to `N+1`, rather than being reverted to the stable `N`.

Avoiding the error in step 4 would require checking the git history to determine the previous stable version.  By preventing "the current version minus one" from being used as a rule of thumb, it forces somebody to check the git history anyways and avoids this failure mode.

leandron: Agreed. I think the original text is more regarding the impact for people who are used to do the current process and will need to adjust to the proposed here.

Thanks for the comment.

driazati (approved):
thanks for sending this in, looks good and should go along nicely with the work in https://github.com/apache/tvm/issues/10646

rfcs/0000-docker-image-tags.md
areusch: it seems like the rollback process should be rolling back the `Jenkinsfile`, not tagging a new Docker image, so it would avoid this case. `git revert` should handle this, i think.

areusch (approved):
thanks @leandron this looks great!

rfcs/0000-docker-image-tags.md
```diff
@@ -0,0 +1,52 @@
...
+
+Defines a format for TVM CI Docker images tag.
+
+# Motivation
+
+In the current format used to identify our Docker images, it is impossible to link back a given version of the image to which version of the repository that represents.
+
+# Guide-level explanation
+
+For years we have been using a patern to version our CI Docker images, which is composed by v0.INCREMENTAL_NUMBER, e.g. tlcpack/ci_cpu:v0.80. Every time an image is updated, we bump the minor number, just as a way to release a new version.
```
Mousius, line 16:
```suggestion
For years we have been using a pattern to version our CI Docker images, which is composed by v0.INCREMENTAL_NUMBER, e.g. tlcpack/ci_cpu:v0.80. Every time an image is updated, we bump the minor number, just as a way to release a new version.
```

rfcs/0000-docker-image-tags.md
```diff
@@ -0,0 +1,52 @@
...
+When looking/using a Docker image, usually we would be interested in:
+
+1. How long ago was this image generated?
+2. What is the last change added in the current images being used in production?
+
+None of these questions can be answered by our current numbering scheme, and to get that information we would usually need to do some digging and inspecting the image to see what is in there.
+
+# Reference-level explanation
+
+As an improvement to the current situation, this RFC proposes the adption of a tagging scheme currently used in the `tlcpackstaging` (https://hub.docker.com/u/tlcpackstaging) repository, composed by:
```
Mousius, line 29:
```suggestion
As an improvement to the current situation, this RFC proposes the adoption of a tagging scheme currently used in the `tlcpackstaging` (https://hub.docker.com/u/tlcpackstaging) repository, composed by:
```

Mousius (changes requested):
This looks like a much better and more consistent format @leandron, could we also move to `-` or `_` everywhere when we roll this out? (not `tlcpackstaging/ci_cpu` and `tlcpack/ci-cpu`)

Other than that, I spotted a few typos 😿

areusch:
@Mousius i support `-` instead of `_` (when running `docker/bash.sh`, you can then use `_` as a sort of rule of thumb to determine if you're referencing a variable-driven revision or a specific one)

mehrdadh (approved):
LGTM!
Thanks for working on this!

rfcs/0000-docker-image-tags.md
gromero, line 40: @leandron I follow here but I think there is a typo in `However, reverting images is quite rare so it should be something the will consume lots of time very often.` i.e. s/`the`/`that`/? But even with that change I think that this phrase can be improved, because: "revert is quite rare but it will consume lots of time very often" seems a contraction in this context?

gromero (changes requested):
@leandron Just nitpicked what looks like a typo. Please see my comment inline.

Otherwise `Rationale and alternatives` makes sense to me and overall LGTM!

rfcs/0000-docker-image-tags.md
leandron: Agreed. I'll change it.

Mousius:
@leandron, looking at Docker Hub (https://hub.docker.com/_/hello-world) it would appear the convention for image names is to use `-` there as well (i.e. `tlcpack/ci-cpu` rather than `tlcpack/ci_cpu`) can we go for that one?

leandron:
> @leandron, looking at Docker Hub (https://hub.docker.com/_/hello-world) it would appear the convention for image names is to use `-` there as well (i.e. `tlcpack/ci-cpu` rather than `tlcpack/ci_cpu`) can we go for that one?

Sure. I’ll push an updated version with this and @gromero’s suggestion as well.

areusch:
@leandron should we merge this? I think the change has landed in Jenkinsfile now

areusch:
cc @Mousius @gromero can you approve?

gromero:
> cc @Mousius @gromero can you approve?

@leandron Hi. Are you still planing to change https://github.com/apache/tvm-rfcs/pull/66#discussion_r854052265 ?

rfcs/0000-docker-image-tags.md
manupa-arm: Hmmm, there is another perspective here.

Ideally, we should be verifying a new-built image using the dockerfile in the repo in every CI run.
Therefore, the revert should really be seen as a revert to the dockerfile.

However, due to various (practical) reasons we are sampling the dockerfile history instead.
Therefore, I'd say going backwards chronologically in the samples (i.e. "the current version minus one") should not be the solution.

I'd go further and say we should be really using an image before dockerfile change and this change allows do that seamlessly. This will allow us to keep changes that happened before the offending change.

WDYT ?

driazati: fyi if my work in https://github.com/apache/tvm/pull/11329 is successful we will be rebuilding the Docker images on every build and using docker hub only as a cache so reverts wouldn't be necessary outside the usual git process

gromero:
@Mousius I don't have any further comments on it. Are you ok with the current state of this RFC? If so, could you please approve the changes so I can merge it? Thanks!

Mousius:
> > @leandron, looking at Docker Hub (https://hub.docker.com/_/hello-world) it would appear the convention for image names is to use `-` there as well (i.e. `tlcpack/ci-cpu` rather than `tlcpack/ci_cpu`) can we go for that one?
> 
> Sure. I’ll push an updated version with this and @gromero’s suggestion as well.

@gromero the above change is still pending as far as I can see?

gromero:
> > > @leandron, looking at Docker Hub (https://hub.docker.com/_/hello-world) it would appear the convention for image names is to use `-` there as well (i.e. `tlcpack/ci-cpu` rather than `tlcpack/ci_cpu`) can we go for that one?
> > 
> > 
> > Sure. I’ll push an updated version with this and @gromero’s suggestion as well.
> 
> @gromero the above change is still pending as far as I can see?

@Mousius hrm right. So we'll need @leandron to make the next move here.

PR #67, opened 2022-04-22
[RFC] Quarterly Releases

driazati:
Cross-posted to discuss here: https://discuss.tvm.apache.org/t/rfc-quarterly-releases/12629

cc @areusch @leandron @Mousius

### DIFF
--- a/0067-quarterly-releases.md
+++ b/0067-quarterly-releases.md
@@ -1,0 +1,53 @@
+- Feature Name: Release Schedule
+- Start Date: 2022-04-21
+- RFC PR: [apache/tvm-rfcs#67](https://github.com/apache/tvm-rfcs/pull/67)
+
+# Summary
+This RFC proposes that TVM move to a quarterly release schedule. Releases would happen every 3 months or so on a schedule set well in advance, independent of individual feature development in TVM.
+
+# Motivation
+Releases are essential to the usage of TVM, especially now that are beginning to work on publishing binary packages for TVM under PyPi. Making TVM releases frequent forces the release process to become well documented and simple rather than bespoke and only achievable by a small group. Users benefit from releases by seeing that the the project is still under active development and providing an easy way to get new features. As of this RFC it has been five months since the last release. It could easily confuse new users when they expect some TVM feature that was only developed recently but is not present in the latest official release.
+
+
+# Guide-level explanation
+TVM has [release process documentation](https://tvm.apache.org/docs/contribute/release_process.html). This RFC proposes that the release candidate vote thread be abolished in favor of a mechanical schedule where releases happen roughly every three months. A release branch will be cut, evaluated for a period of two weeks, then a release published. Publishing a release entails:
+
+* Gathering and organizing release notes since the last release
+* Posting a source code release on GitHub
+* Uploading the source code to the Apache SVN repository
+* Uploading binary packages to tlcpack
+* Uploading binary packages to PyPi
+
+# Reference-level explanation
+
+Prior to a release a new, lightweight vote would instead be used to nominate a release manager, a committer who will be responsible for guiding along the release process. Releases will roughly match the quarterly [calendar dates](https://en.wikipedia.org/wiki/Calendar_year#Quarters) shifted two weeks earlier (Q1: mid March, Q2: mid June, Q3: mid September, Q4: mid December). The release manager will have final say on all dates and consideration should be given for their personal schedule. The timeline will be as follows:
+* Three weeks prior to the release
+    * The release manager will cut a release branch and create a new tag
+    * The release manager will open a GitHub issue announcing the release branch cut and target date for the release and state that any further inclusions in the release must be manually cherry-picked
+    * The release manager will create a PR targeted to merge into the release branch with the necessary changes to make a release (i.e. changing version numbers)
+    * Contributors that wish changes to be cherry-picked should comment on the announcement issue with the relevant PRs and commits and their reasoning. The release manager has final say on which changes should be included but should aim to be inclusive at this stage
+    * The release manager begins gathering release notes
+
+* One week prior to release
+    * Cherry picks become limited to critical changes only
+    * The release manager begins building binaries and testing them against TVM's test suite
+
+* Day of the release
+    * The release manager publishes the relevant binaries
+    * The release manager closes the release issue
+    * The release manager makes a GitHub release and updates an in-repo file `RELEASE.md` on both `main` and the release branch with the release notes
+
+Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+# Prior Work
+
+* This [proposal](https://discuss.tvm.apache.org/t/pre-rfc-switch-to-time-based-releases/4245) suggests a similar thing for TVM and links to a great [write-up](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan) from the Kafka project describing their plan.
+
+# Drawbacks
+
+* This requires a high level of commitment from a single individual (the release manager). If someone is not able to fill or execute this role effectively the release will be stalled.
+
+# Rationale and alternatives
+
+* The main alternative is waiting until specific features have been developed in order to make a release. This has caveats as TVM has many sub-projects within it, so it's not clear which is significant enough to warrant a release.
+
--- a/rfcs/0067-quarterly-releases.md
+++ b/rfcs/0067-quarterly-releases.md
@@ -10,27 +10,34 @@ Releases are essential to the usage of TVM, especially now that are beginning to
 
 
 # Guide-level explanation
-TVM has [release process documentation](https://tvm.apache.org/docs/contribute/release_process.html). This RFC proposes that the release candidate vote thread be abolished in favor of a mechanical schedule where releases happen roughly every three months. A release branch will be cut, evaluated for a period of two weeks, then a release published. Publishing a release entails:
+TVM has [release process documentation](https://tvm.apache.org/docs/contribute/release_process.html). This RFC proposes that a release candidate vote thread sent on a schedule roughly every three months. A release branch will be cut, evaluated for a period of two weeks, evaluated by the PMC, then a release published. Publishing a release entails:
 
 * Gathering and organizing release notes since the last release
 * Posting a source code release on GitHub
 * Uploading the source code to the Apache SVN repository
 * Uploading binary packages to tlcpack
 * Uploading binary packages to PyPi
+* Getting an approval vote from the PMC
 
 # Reference-level explanation
 
 Prior to a release a new, lightweight vote would instead be used to nominate a release manager, a committer who will be responsible for guiding along the release process. Releases will roughly match the quarterly [calendar dates](https://en.wikipedia.org/wiki/Calendar_year#Quarters) shifted two weeks earlier (Q1: mid March, Q2: mid June, Q3: mid September, Q4: mid December). The release manager will have final say on all dates and consideration should be given for their personal schedule. The timeline will be as follows:
 * Three weeks prior to the release
     * The release manager will cut a release branch and create a new tag
+    * The release manager will audit the licenses of all project dependencies to ensure they are compatible with Apache
     * The release manager will open a GitHub issue announcing the release branch cut and target date for the release and state that any further inclusions in the release must be manually cherry-picked
     * The release manager will create a PR targeted to merge into the release branch with the necessary changes to make a release (i.e. changing version numbers)
     * Contributors that wish changes to be cherry-picked should comment on the announcement issue with the relevant PRs and commits and their reasoning. The release manager has final say on which changes should be included but should aim to be inclusive at this stage
     * The release manager begins gathering release notes
 
-* One week prior to release
+* Two weeks prior to release
     * Cherry picks become limited to critical changes only
     * The release manager begins building binaries and testing them against TVM's test suite
+    * The release manager sends the release notes and plan to the PMC for approval (feedback thread will be open for one week)
+    * The release manager performs another license audit of all project dependencies to ensure they are compatible with Apache
+
+* One week prior to release
+    * The release manager sends the release and binaries to the PMC for an approval vote (thread open for one week). If rejected the building and test process must be started anew and repeated until the PMC approves.
 
 * Day of the release
     * The release manager publishes the relevant binaries
@@ -38,6 +45,12 @@ Prior to a release a new, lightweight vote would instead be used to nominate a r
     * The release manager makes a GitHub release and updates an in-repo file `RELEASE.md` on both `main` and the release branch with the release notes
 
 Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release.
+
+This RFC will not go into effect until we resolve the API around `relay.build` since that may introduce significant churn in the community. See [the discussion](https://discuss.tvm.apache.org/t/pre-rfc-compilation-configuration-representation/11372) for details.
+
+Even with this RFC the ultimate decision to releases rests with the community. If no one wants to do a release, then it will be skipped or delayed until the community agrees and a release manager can be selected.
 
 # Prior Work
 
--- a/rfcs/0067-quarterly-releases.md
+++ b/rfcs/0067-quarterly-releases.md
@@ -21,7 +21,7 @@ TVM has [release process documentation](https://tvm.apache.org/docs/contribute/r
 
 # Reference-level explanation
 
-Prior to a release a new, lightweight vote would instead be used to nominate a release manager, a committer who will be responsible for guiding along the release process. Releases will roughly match the quarterly [calendar dates](https://en.wikipedia.org/wiki/Calendar_year#Quarters) shifted two weeks earlier (Q1: mid March, Q2: mid June, Q3: mid September, Q4: mid December). The release manager will have final say on all dates and consideration should be given for their personal schedule. The timeline will be as follows:
+Prior to a release a new, lightweight vote would instead be used to nominate a release manager, a committer who will be responsible for guiding along the release process. Releases will be published roughly every three months, with one in January, April, July and October. The release manager will have final say on all dates and consideration should be given for their personal schedule. The timeline will be as follows:
 
 * Three weeks prior to the release
     * The release manager will cut a release branch and create a new tag
@@ -56,7 +56,7 @@ When publishing a new release, the release manager should also check the previou
 ## Versioning
 The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release. The release manager will be responsible for maintaining the release until the next release, which mainly entails putting up a patch release if necessary. The release manager may also delegate this responsibility to another party if both agree.
 
-Releases will be supported and get patch updates so long as that minor version is the most recent. For example, version 0.N.0 will get patches until version 0.(N + 1).0 is released. Critical patch fixes may also be backported to older releases at the community and release manager's descretion. The release manager is responsible for patch releases for bug fixes as well, though they may delegate to another if they so choose.
+Releases will be supported and get patch updates so long as that minor version is the most recent. For example, version 0.N.0 will get patches until version 0.(N + 1).0 is released. Critical patch fixes may also be backported to older releases at the community and release manager's descretion. The release manager is responsible for patch releases for bug fixes as well, though they may delegate to another if they so choose. When asking for support on forums, one of the first steps for users should be to reproduce their bug on most recently nightly build of TVM.
 
 # Prior Work
 
--- a/rfcs/0067-quarterly-releases.md
+++ b/rfcs/0067-quarterly-releases.md
@@ -54,6 +54,8 @@ Even with this RFC the ultimate decision to releases rests with the community. I
 ## Versioning
 The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release. The release manager will be responsible for maintaining the release until the next release, which mainly entails putting up a patch release if necessary. The release manager may also delegate this responsibility to another party if both agree.
 
+Releases will be supported and get patch updates so long as that minor version is the most recent. For example, version 0.N.0 will get patches until version 0.(N + 1).0 is released. Critical patch fixes may also be backported to older releases at the community and release manager's descretion. The release manager is responsible for patch releases as well, though they may delegate to another if they so choose.
+
 # Prior Work
 
 * This [proposal](https://discuss.tvm.apache.org/t/pre-rfc-switch-to-time-based-releases/4245) suggests a similar thing for TVM and links to a great [write-up](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan) from the Kafka project describing their plan.
--- a/rfcs/0067-quarterly-releases.md
+++ b/rfcs/0067-quarterly-releases.md
@@ -33,11 +33,11 @@ Prior to a release a new, lightweight vote would instead be used to nominate a r
 * Two weeks prior to release
     * Cherry picks become limited to critical changes only
     * The release manager begins building binaries and testing them against TVM's test suite
-    * The release manager sends the release notes and plan to the PMC for approval (feedback thread will be open for one week)
+    * The release manager sends the release notes and plan to a vote thread for approval (feedback thread will be open for one week)
     * The release manager performs another license audit of all project dependencies to ensure they are compatible with Apache
 
 * One week prior to release
-    * The release manager sends the release and binaries to the PMC for an approval vote (thread open for one week). If rejected the building and test process must be started anew and repeated until the PMC approves.
+    * The release manager sends the release and binaries to a vote thread for approval (thread open for one week). If rejected the building and test process must be started anew and repeated until the PMC approves.
 
 * Day of the release
     * The release manager publishes the relevant binaries
@@ -46,7 +46,7 @@ Prior to a release a new, lightweight vote would instead be used to nominate a r
 
 Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
 
-The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release.
+The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release. The release manager will be responsible for maintaining the release until the next release, which mainly entails putting up a patch release if necessary. The release manager may also delegate this responsibility to another party if both agree.
 
 This RFC will not go into effect until we resolve the API around `relay.build` since that may introduce significant churn in the community. See [the discussion](https://discuss.tvm.apache.org/t/pre-rfc-compilation-configuration-representation/11372) for details.
 
--- a/rfcs/0067-quarterly-releases.md
+++ b/rfcs/0067-quarterly-releases.md
@@ -51,10 +51,12 @@ Developers are heavily encouraged to use feature flags and Python API convention
 
 Even with this RFC the ultimate decision to releases rests with the community. If no one wants to do a release, then it will be skipped or delayed until the community agrees and a release manager can be selected.
 
+When publishing a new release, the release manager should also check the previous release version's branch to see if any commits have been added since the last release and publish a new minor version if so.
+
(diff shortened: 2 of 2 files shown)


areusch:
cc @apache/tvm-committers

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
+- Feature Name: Release Schedule
+- Start Date: 2022-04-21
+- RFC PR: [apache/tvm-rfcs#67](https://github.com/apache/tvm-rfcs/pull/67)
+
+# Summary
+This RFC proposes that TVM move to a quarterly release schedule. Releases would happen every 3 months or so on a schedule set well in advance, independent of individual feature development in TVM.
```
Mousius, line 6: Just clarifying, I assume that even though the features themselves are independent of the release we'd eventually want to pick a versioning scheme that reflects the changes made?

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Motivation
+Releases are essential to the usage of TVM, especially now that are beginning to work on publishing binary packages for TVM under PyPi. Making TVM releases frequent forces the release process to become well documented and simple rather than bespoke and only achievable by a small group. Users benefit from releases by seeing that the the project is still under active development and providing an easy way to get new features. As of this RFC it has been five months since the last release. It could easily confuse new users when they expect some TVM feature that was only developed recently but is not present in the latest official release.
```
Mousius, line 9: Regarding "As of this RFC it has been five months since the last release", we could've released far earlier but there was some discomfort around evolving the `relay.build` API multiple times - with these regular releases, are we suggesting that we can evolve the API over a number of releases?

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Guide-level explanation
+TVM has [release process documentation](https://tvm.apache.org/docs/contribute/release_process.html). This RFC proposes that the release candidate vote thread be abolished in favor of a mechanical schedule where releases happen roughly every three months. A release branch will be cut, evaluated for a period of two weeks, then a release published. Publishing a release entails:
```
Mousius, line 13: Doesn't this RFC just provide a schedule for following the release process? The release candidate vote thread is part of that process and ensures we don't mistakenly ship a broken release, why do we want to abolish it as part of this rather than incorporate it into the two week review window?

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
+* Gathering and organizing release notes since the last release
+* Posting a source code release on GitHub
+* Uploading the source code to the Apache SVN repository
+* Uploading binary packages to tlcpack
+* Uploading binary packages to PyPi
+
+# Reference-level explanation
+
+Prior to a release a new, lightweight vote would instead be used to nominate a release manager, a committer who will be responsible for guiding along the release process. Releases will roughly match the quarterly [calendar dates](https://en.wikipedia.org/wiki/Calendar_year#Quarters) shifted two weeks earlier (Q1: mid March, Q2: mid June, Q3: mid September, Q4: mid December). The release manager will have final say on all dates and consideration should be given for their personal schedule. The timeline will be as follows:
```
Mousius, line 23: Could you motivate why we shift the release two weeks earlier than the calendar quarters?

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
+Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+# Prior Work
+
+* This [proposal](https://discuss.tvm.apache.org/t/pre-rfc-switch-to-time-based-releases/4245) suggests a similar thing for TVM and links to a great [write-up](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan) from the Kafka project describing their plan.
+
+# Drawbacks
+
+* This requires a high level of commitment from a single individual (the release manager). If someone is not able to fill or execute this role effectively the release will be stalled.
```
Mousius, line 48: Interestingly, the Kafka document states that if noone is interested they just don't release - that seems a fair way to decide to skip the release and come back in 3 months rather than stalling it which could create the scenario we have now.

Mousius (changes requested):
Thanks for formalising this RFC @driazati, this is a great improvement to how releases are currently done in TVM. I've highlighted some concerns but the sooner we start doing this the sooner we'll be able to figure out how to minimize the friction of releasing 😸

tqchen:
cc @apache/tvm-committers for general visibility

0067-quarterly-releases.md
tqchen: Thanks @driazati for proposing the change and bringing up the same goal of making release streamlined (which we all share).

Just to clarify this particular point(of release voting) as it is something that is related to the policy. Overall release policy is something that apache handles (sometimes extra) carefully overall. Release voting thread is mandatory per  policy and it is not something that we can change project-wide. 

The rationale is that the release voting thread brings broad community awareness and overall license requirements.  A release needs at least three PMC vote to be shipped. We can however, keep the schedule streamlined and improve the overall efficiency driven by a release manager.

The original release policy guide is a good reference pt -- about the GPG key needs and where to upload, posts.

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Motivation
+Releases are essential to the usage of TVM, especially now that are beginning to work on publishing binary packages for TVM under PyPi. Making TVM releases frequent forces the release process to become well documented and simple rather than bespoke and only achievable by a small group. Users benefit from releases by seeing that the the project is still under active development and providing an easy way to get new features. As of this RFC it has been five months since the last release. It could easily confuse new users when they expect some TVM feature that was only developed recently but is not present in the latest official release.
+
+
+# Guide-level explanation
+TVM has [release process documentation](https://tvm.apache.org/docs/contribute/release_process.html). This RFC proposes that the release candidate vote thread be abolished in favor of a mechanical schedule where releases happen roughly every three months. A release branch will be cut, evaluated for a period of two weeks, then a release published. Publishing a release entails:
+
+* Gathering and organizing release notes since the last release
+* Posting a source code release on GitHub
```
tqchen, line 16: would be useful to clarify what is the "official release". The official release is the source code tarball posted to the TVM webpage download section(with the corresponding GPG keys that can be verified) and the apache svn repo.

The binary packages, and source code tag on github are "convenient by product" that is not part of the official release. I know it sounds a bit strange. ASF have a quite pedantic approach towards release and licensing. The main retionale is that for people who want to be extra "safe" on licensing implications, they want to go and download from the webpage -- where the source code comes clean without binary dependencies and things can be built from scratch. This pedantic view, however, is also why many organizations trust apache software (e.g. they can be applied without review)

Of course for most people the ability to quickly use and try things out is more important.

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+* Three weeks prior to the release
+    * The release manager will cut a release branch and create a new tag
+    * The release manager will open a GitHub issue announcing the release branch cut and target date for the release and state that any further inclusions in the release must be manually cherry-picked
+    * The release manager will create a PR targeted to merge into the release branch with the necessary changes to make a release (i.e. changing version numbers)
+    * Contributors that wish changes to be cherry-picked should comment on the announcement issue with the relevant PRs and commits and their reasoning. The release manager has final say on which changes should be included but should aim to be inclusive at this stage
+    * The release manager begins gathering release notes
+
+* One week prior to release
+    * Cherry picks become limited to critical changes only
+    * The release manager begins building binaries and testing them against TVM's test suite
```
tqchen, line 33: Let us add a dependency license check to quickly verify the new dependencies added since last release, and ensure [license](https://github.com/apache/tvm/tree/main/licenses) files are inplace and they are compatible with ASF policy

0067-quarterly-releases.md
tqchen: I agree that skip the release would be a good lightweight option for now. We can also do out of schedule release if necessary if there is a strong need to do so. But the goal is to keep a regular cadence where things happens most of the time

0067-quarterly-releases.md
```diff
@@ -0,0 +1,53 @@
...
+
```
tqchen, line 34: Add a feedback(one week) and vote period(one week) on the release candidate. The feedbacks are based on the rc tarball(with GPG keys and sha512) to be verifiable to avoid possible errors during the release shipping(e.g. mistakenly packge in something or ignores something into the source tarball). There is also a process on how new rc should be generated -- following the same process of updating the tarball and code if there is feedbacks.

0067-quarterly-releases.md
driazati: I'm imagining it as 0.9 -> 0.10 -> 0.11, etc until we decide to 1.0, I added some info about this in the reference section

0067-quarterly-releases.md
driazati: We certainly can change APIs over releases, though maintaining as much backwards compatibility and consistency as we can is good. @areusch and I discussed this specific issue and we were thinking we'd wait until that is resolved to start doing this RFC, I added another part to the reference about this issue specifically

0067-quarterly-releases.md
driazati: Ah, I wasn't aware of the rules on the vote thread. I think it is fine to keep in that case but move it to the end of the process, so the release manager does most of the work and testing but sends it to the PMC for approval before publishing anything

0067-quarterly-releases.md
driazati: This list has both uploading the source to GitHub and Apache as a task for the release, do we just need to clear on GitHub that the officially released code can be found on the Apache site?

0067-quarterly-releases.md
driazati: mainly just to avoid conflicting with common holidays (e.g. new year for U.S.), the time interval between is the most important part here so it'd be fine to move the start date to any time if anyone has preferences

0067-quarterly-releases.md
driazati: good notes, I added a small section about this to the reference. Practically it makes sense as well, since there are always people in the loop deciding whether to proceed or not

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,66 @@
...
+
+* This [proposal](https://discuss.tvm.apache.org/t/pre-rfc-switch-to-time-based-releases/4245) suggests a similar thing for TVM and links to a great [write-up](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan) from the Kafka project describing their plan.
+
+# Drawbacks
+
+* This requires a high level of commitment from a single individual (the release manager). If someone is not able to fill or execute this role effectively the release will be stalled.
+
+# Rationale and alternatives
+
+* The main alternative is waiting until specific features have been developed in order to make a release. This has caveats as TVM has many sub-projects within it, so it's not clear which is significant enough to warrant a release.
```
leandron, line 65: On this point of "_waiting until specific features have been developed in order to make a release_", I think as a very-active project, we always have quite a few new features being developed, so I don't think we should have as a general approach to *wait* for features to do releases.

In my opinion, we should have as a general policy that any major feature should have a toggle, to be disabled by default while it is in development stage, so that releases can be taken without partial/broken features in it.

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,66 @@
...
+
+Even with this RFC the ultimate decision to releases rests with the community. If no one wants to do a release, then it will be skipped or delayed until the community agrees and a release manager can be selected.
+
+# Prior Work
+
+* This [proposal](https://discuss.tvm.apache.org/t/pre-rfc-switch-to-time-based-releases/4245) suggests a similar thing for TVM and links to a great [write-up](https://cwiki.apache.org/confluence/display/KAFKA/Time+Based+Release+Plan) from the Kafka project describing their plan.
+
+# Drawbacks
+
+* This requires a high level of commitment from a single individual (the release manager). If someone is not able to fill or execute this role effectively the release will be stalled.
```
leandron, line 61: Perhaps we should also have a deputy, who's aware about what is going on, and in the absence of the release manger can take over.

0067-quarterly-releases.md
tqchen: no need to explicit send to PMC as PMC members normally vote on the same public voting thread

0067-quarterly-releases.md
u99127: A project need releases so that end users have a way of consuming the work done since the last release including bug fixes, no releases means that as a community we are saying that end users don't need to consume the work done in the last 3 months including fixes for bugs  and it lessens the value of releases and we leave it to people checking out random git SHAs and working from that. Not very sustainable for end users. 

For the TVM community to be more end user focussed releases need to happen on a cadence rather than it being dependent on someone's interest. If there is no volunteer , maybe the PMC "volunteers" someone ?

rfcs/0067-quarterly-releases.md
driazati: agreed that's a good idea but I don't think we need to explicitly codify that, if there is a problem with the release manager's time they should communicate that to the community and figure something out from there

rfcs/0067-quarterly-releases.md
driazati: both good points though I would say using feature flags and how sub-teams do planning is outside the scope of this RFC. In general nothing should ever be broken on `main` and we can hopefully establish some processes to ensure API compatibility doesn't break in crazy ways between releases as part of the release testing

u99127: Hmm, I am not sure I agree fully with the fact that this is out of scope for the RFC.  

This RFC is to set out the policy and set out a timeline for releases. 

One of the current drawbacks of the TVM project is that industrial contributors find it difficult to see how their customers would consume the open source project without the existing standard basic practices of regular releases and release trains. 

The consequence of that is that industrial contributors use these timelines to plan for their activities and would adjust other activities they have to do in order to make sure that features are landed in a sensible fashion . If TVM needs to remain relevant to industry, then these fundamental software engineering practices need to be put in place and remain.

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,66 @@
...
+
+Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release. The release manager will be responsible for maintaining the release until the next release, which mainly entails putting up a patch release if necessary. The release manager may also delegate this responsibility to another party if both agree.
+
+This RFC will not go into effect until we resolve the API around `relay.build` since that may introduce significant churn in the community. See [the discussion](https://discuss.tvm.apache.org/t/pre-rfc-compilation-configuration-representation/11372) for details.
```
u99127, line 51: I don't see quite understand why this has a place in the RFC since the RFC is about release time lines. Equally leaving something so open-ended is a bit unfair to the end users.

 Playing devils advocate, if it takes a year to fix this - what about bug fixes to 0.8 ?

tqchen: I agree that the particular discussion should be left out. We take cautions to evolve the API and the decision of evolution can happen before/after a release as long as it is clearly communicated

Mousius: As the author of the post, I strongly agree we shouldn't block on it given how long it's gone with no sign of resolution. We're currently releasing `0.x` releases, which are typically considered to be unstable, and any versioning scheme we moved to should also allow for multiple breaking changes in a row (such as repeatedly incrementing the major version in [semantic versioning](https://www.semver.org)).

rfcs/0067-quarterly-releases.md
leandron: > both good points though I would say using feature flags and how sub-teams do planning is outside the scope of this RFC.

How teams will organise to implement such feature toggles is their decision - agreed.

To be more specific, what I think worth to be explicit in this RFC is that such mechanisms should be mandatory, as a way to keep viable releases going. Often - for technical reasons - features are still making their way into the repository, so it would be good to create a culture in which even with partial features in the repo, we can still release if we need/want to.

rfcs/0067-quarterly-releases.md
driazati: agreed since there's not been a lot of movement on it lately, removed

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,67 @@
...
+
+* Day of the release
+    * The release manager publishes the relevant binaries
+    * The release manager closes the release issue
+    * The release manager makes a GitHub release and updates an in-repo file `RELEASE.md` on both `main` and the release branch with the release notes
+
+Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+The release manager will use the release notes and discussions with developers to determine the next version number. Releases will continue the current versioning scheme of `major.minor.patch`, with a typical release involving a bump of the minor release version. Patch versions will be used for follow up releases onto a quarterly release, but not for the next quarter's release. The release manager will be responsible for maintaining the release until the next release, which mainly entails putting up a patch release if necessary. The release manager may also delegate this responsibility to another party if both agree.
```
hogepodge, line 50: This looks like semver, and will likely be interpreted as semver, but is not. I would strongly encourage considering officially adopting and adhering to semver to help manage API expectations for groups trying to use TVM in production. Time-based releases don't always fit perfectly with semantic versioning, but careful consideration of API changes and additions should make adherence possible. https://semver.org

hogepodge: Another alternative is adopting a date-based version number such as `year.quarter.minor`, which will meet the needs for resolving dependencies but does not imply the strict API compatibility requirements across releases that semver has.

areusch: i agree we are not semver here. however, we are also not 1.0, so it might be worth reconsidering our versioning scheme once we do hit 1.0. I'd like to see some more infrastructure and processes developed (e.g. feature flags and a more codified set of expectations for large projects like metascheduler which are currently developed in-tree and merged in an effort-specific way) before we can commit to semver.

we could consider e.g. `0.202206.0`, though it's a bit hard to parse. i don't think we should propose to rev the major version past 1.0 at this time, though.

areusch (commented):
thanks @driazati ! left some comments

areusch:
@u99127 these are great points! i have some thoughts:

> Lifecycle of a release , what happens to a release x.y.0 after the release point ? The community has 2 choices, use frugal release manager resources in making sure the next (x.y.0)+1 release happens on schedule and on time . Or continues to improve the quality of x.y.0 by fixing bugs and making sure that x.y.1 / x.y.2 keep happening - but then after a year we'll have too many releases and given where we are in the lifecycle of TVM, this may not be necessary.

It seems like there may be a difference between a 1.0 release and a 0.x release in terms of support. Perhaps a 0.x release is patch-supported for a short time (e.g. until the next release cycle starts)?

> At what point do we declare an old release something that the community will not support further ? i.e. x.y.0 is valid for 6 months or a year after which there will be no releases from the branch and the community will not take any bug fixes into the branch ?

yeah this is a great question. it seems like right now our policy is to "support" the previous release and no further back. also something i'd expect to grow or perhaps we'd arrive at an LTS-style support policy depending on development cadence.

> Are release managers appointed for the life cycle of a 0.x release - i.e is it the responsibility to keep 0.x until its end of life ?

it could be good to state that a release manager is expected to shepherd questions (though the entire set of committers and community is expected to try and answer without the need for shepherding) about the release on the forum. i'd propose we start with a shorter support period so it is not overly daunting to be a release manager.

driazati:
I agree with @areusch that a lot changes pre and post 1.0. This RFC's main focus is getting code out the door often, so with that I think it makes the most sense to only support the current version and only publish a new release mid-cycle for critical bug fixes and security issues.

Ostensibly the release manager should be responsible for the release they published (and in the best case this would mean a release needs no patches and thus requires no extra work), but this could get passed off or reassigned as we move along based on who's available.

rfcs/0067-quarterly-releases.md
driazati: agree semver is probably too much for us right now and we should revisit at 1.0. Increasing version numbers seem simplest to think about which is what this RFC proposes and anecdotally date releases aren't as common for Python packages (though they are listed in [PEP 440](https://peps.python.org/pep-0440/))

rfcs/0067-quarterly-releases.md
hogepodge: To help mitigate with this, several projects stagger release managers, with an outgoing release manager and an incoming shadow release manager. This spreads out the responsibility and gives an opportunity to formally train someone on release management duties.

0067-quarterly-releases.md
Mousius: Is the aim to start a release in this time period or to finish a release? I'm wondering if it's better to aim for the fairly standard January/April/July/October with the flexibility given to the release manager rather than trying to optimise for holidays in a global project. For example, many UK companies reset holidays in April, so there's often a lot of people taking it in mid March.

areusch:
We discussed this RFC briefly this morning at the [Community Meeting](https://discuss.tvm.apache.org/t/next-tvm-community-meeting-may-11/12714/3). Here are notes:
- @Mousius notes that because we're 0.x version numbers, we don't need to worry too much about breaking APIs
- @driazati notes that many of our APIs are stable and we should still think about this, but that we shouldn't spend a ton of time on those issues and should get releases out the door.

0067-quarterly-releases.md
areusch: I guess the proposal is to start a release. I do see that we grant flexibility here to the release manager, but I don't know if that's sufficient or not here. It seems like doing an April release 2 weeks later would be ok, if someone was volunteering to be release manager. On the other hand it could hamper participation in the release. Perhaps we can try this a few times and see if we need to make changes?

areusch:
I think the main outstanding thing is the question of support here. I'd love for a few more folks to weigh in, tagging a few folks who may have ideas: @tqchen @jroesch @kparzysz-quic @u99127 @Mousius @leandron @comaniac @zhiics @Hzfengsy @ZihengJiang @yzhliu @masahi @icemelon 

broadly, the question is: once we release version `0.X.0`, should we discover a bug, we should likely fix it on that branch and on `main` and release `0.X.Y`. how long shall we commit to doing that? If a bug is discovered on `0.X-1.0`, will we fix it?

my vote is the following (for now, we can revisit at 1.0):
- we agree we will fix major bugs on `0.X.Y` where `0.X.Y` is the latest release.
- when we make a new minor version release e.g. `0.(X+1).0`, we will stop maintaining releases starting with `0.X.` at that time.
- committers/pmcs should exercise their judgement when determining if something warrants a fix. 
- the same policy should apply on the forum--if someone asks a question on an old minor release version, it should be acceptable to ask them to reproduce it on the latest release.

Mousius:
> my vote is the following (for now, we can revisit at 1.0):
> 
> * we agree we will fix major bugs on `0.X.Y` where `0.X.Y` is the latest release.
> * when we make a new minor version release e.g. `0.(X+1).0`, we will stop maintaining releases starting with `0.X.` at that time.
> * committers/pmcs should exercise their judgement when determining if something warrants a fix.
> * the same policy should apply on the forum--if someone asks a question on an old minor release version, it should be acceptable to ask them to reproduce it on the latest release.

Just checking for my understanding, this would be:
* Release 0.9.0 next
* We find a bug, release 0.9.1
* 0.9.1 is now the only supported release
* 0.10.0 comes out
* 0.10.0 is now the only supported release

Who does the patch releases? I'm assuming the nominated release person from `0.9.0` ?

0067-quarterly-releases.md
Mousius: I think the simplest thing is "we aim to deliver a finished release in Jan/Apr/Jul/Oct" as that's straight forward and gives people the flexibility of a month window?

driazati:
thanks for summarizing the meeting! I agree with @areusch's points on support and have added some text about it

Mousius:
> @Mousius that's my suggestion. i don't know who should do it--perhaps at least the nominated person is responsible for finding someone to do it and ensuring it's done?

I'd suggest we make it clear it may be required when taking the role, but equally if they find someone to swap with then that's ok 😸

u99127:
> I think the main outstanding thing is the question of support here. I'd love for a few more folks to weigh in, tagging a few folks who may have ideas: @tqchen @jroesch @kparzysz-quic @u99127 @Mousius @leandron @comaniac @zhiics @Hzfengsy @ZihengJiang @yzhliu @masahi @icemelon
> 
> broadly, the question is: once we release version `0.X.0`, should we discover a bug, we should likely fix it on that branch and on `main` and release `0.X.Y`. how long shall we commit to doing that? If a bug is discovered on `0.X-1.0`, will we fix it?
> 
> my vote is the following (for now, we can revisit at 1.0):
> 
> * we agree we will fix major bugs on `0.X.Y` where `0.X.Y` is the latest release.

Ack. 

> * when we make a new minor version release e.g. `0.(X+1).0`, we will stop maintaining releases starting with `0.X.` at that time.

When we make a new minor version release 0.(x+1).0 , we do a 0.X.(Y+1) and close the branch, otherwise there might be bug fixes since 0.x.y that never see a release or the light of day. 

I'd take it with that minor adjustments.


> * committers/pmcs should exercise their judgement when determining if something warrants a fix.

s/fix/backport

> * the same policy should apply on the forum--if someone asks a question on an old minor release version, it should be acceptable to ask them to reproduce it on the latest release.

Hmm, I'd suggest that we say all bugs / questions should be reproduced with trunk / main and the latest release rather than an unsupported release. 

We probably need to update the "Releases tab" on gh to make this work.

This is looking like good small steps forward. Very good to see this coming together.

Ramana

areusch:
ok cool, so i think we have these actionables:
- note that when we do a minor release, we check the previous minor release branch and also do a revision release if there are unreleased changes there
- make it clear that the release manager is responsible to ensure that bugs raised with the release are addressed via patch/revision release, but that such tasks are largely expected to be delegated to members of the community.

u99127:
> I think the main outstanding thing is the question of support here. I'd love for a few more folks to weigh in, tagging a few folks who may have ideas: @tqchen @jroesch @kparzysz-quic @u99127 @Mousius @leandron @comaniac @zhiics @Hzfengsy @ZihengJiang @yzhliu @masahi @icemelon
> 
> broadly, the question is: once we release version `0.X.0`, should we discover a bug, we should likely fix it on that branch and on `main` and release `0.X.Y`. how long shall we commit to doing that? If a bug is discovered on `0.X-1.0`, will we fix it?
> 
> my vote is the following (for now, we can revisit at 1.0):
> 
> * we agree we will fix major bugs on `0.X.Y` where `0.X.Y` is the latest release.

Agreed. 

> * when we make a new minor version release e.g. `0.(X+1).0`, we will stop maintaining releases starting with `0.X.` at that time.

Agreed. 

> * committers/pmcs should exercise their judgement when determining if something warrants a fix.
> * the same policy should apply on the forum--if someone asks a question on an old minor release version, it should be acceptable to ask them to reproduce it on the latest release.

I agree and I'd be stronger that the bug needs to be reproducible on trunk / main and the latest release. 

How do we get this moving forward ? 

R

Mousius (approved):
Thanks for the updates @driazati, I think we've gotten a good amount of input here from folks and we can move forwards with this for a July release.

Will leave this open until end of week unless someone has more comments 😸 

cc @hogepodge @leandron @areusch @tqchen @u99127

areusch:
Great. let's move this to a PMC voting thread.

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,72 @@
...
+    * The release manager will cut a release branch and create a new tag
+    * The release manager will audit the licenses of all project dependencies to ensure they are compatible with Apache
+    * The release manager will open a GitHub issue announcing the release branch cut and target date for the release and state that any further inclusions in the release must be manually cherry-picked
+    * The release manager will create a PR targeted to merge into the release branch with the necessary changes to make a release (i.e. changing version numbers)
+    * Contributors that wish changes to be cherry-picked should comment on the announcement issue with the relevant PRs and commits and their reasoning. The release manager has final say on which changes should be included but should aim to be inclusive at this stage
+    * The release manager begins gathering release notes
+
+* Two weeks prior to release
+    * Cherry picks become limited to critical changes only
+    * The release manager begins building binaries and testing them against TVM's test suite and running on relevant hardware
```
kparzysz-quic, line 36: What is the test suite?  Is it anything beyond what the Jenkins tests already do?

driazati: Not currently and this RFC intentionally doesn't specify it, we can iterate on release-specific tests more as we go through the process

rfcs/0067-quarterly-releases.md
```diff
@@ -0,0 +1,72 @@
...
+
+* Day of the release
+    * The release manager publishes the relevant binaries, including the source `.tar` to Apache
+    * The release manager closes the release issue
+    * The release manager makes a GitHub release via a tag and updates an in-repo file `RELEASE.md` on both `main` and the release branch with the release notes
+
+Much of this can be automated via GitHub Actions on the apache/tvm repo. Eventually (though maybe not for the upcoming release) these will handle all the building, testing, and publishing of releases so the job of the release manager will become simpler over time.
+
+Developers are heavily encouraged to use feature flags and Python API conventions (i.e. prefixing unstable or private APIs with an underscore) to enable in-progress features. It is important both for users to be able to access new or experimental behavior but also that they can get a stable interface, so these kinds of changes should be hidden behind opt-in flags.
+
```
csullivan, line 51: > Developers are heavily encouraged to use feature flags ... to enable in-progress features. It is important both for users to be able to access new or experimental behavior but also that they can get a stable interface, so these kinds of changes should be hidden behind opt-in flags.

Thanks @driazati, I think this specifically is a very positive move for TVM to make. I have encountered perception among developers that designs must be near-perfect before making it into TVM. 

I'm happy to see explicit language around experimental feature flags which explicitly encourage iterative design. I wonder about your and others opinion on whether we need a separate RFC to formalize the process around introducing experimental changes and their path to graduation into stable releases?

Nonetheless this is great first step, thank you for outlining it.

areusch:
Thanks all for the comments, the vote has passed and we'll merge this RFC.

PR #72, opened 2022-05-12
Create 0072-dynamic-autoscheduler.md

ArmageddonKnight:
@comaniac @junrushao1994

### DIFF
--- a/rfcs/0072-dynamic-autoscheduler.md
+++ b/rfcs/0072-dynamic-autoscheduler.md
@@ -1,0 +1,207 @@
+- Feature Name: DietCode: An Auto-Scheduler for Dynamic Tensor Programs
+- Start Date: (2022-05-10)
+- RFC PR: [apache/tvm-rfcs#xx](https://github.com/apache/tvm-rfcs/pull/xx)
+- GitHub Issue: [apache/tvm#yy](https://github.com/apache/tvm/pull/yy)
+
+# Summary
+[summary]: #summary
+
+We propose to integrate DietCode, an auto-scheduler for dynamic tensor programs,
+to AutoTIR. DietCode offers the following features:
+- A shape-generic search space to cover possible shapes in dynamic shape
+  workloads.
+- A dynamic-shape aware cost model to judge the quality of schedule candidates.
+- Enhancement to the TVM CUDA codegen for imperfect tiling.
+
+DietCode has been published by MLSys 2022 so please see [the
+paper](https://proceedings.mlsys.org/paper/2022/hash/fa7cdfad1a5aaf8370ebeda47a1ff1c3-Abstract.html)
+for more details and evaluations. Meanwhile, the latest DietCode codebase is also publicly
+available [here](https://github.com/UofT-EcoSystem/DietCode).
+
+# Motivation
+[motivation]: #motivation
+
+Achieving high performance for compute-intensive operators in machine learning
+workloads is a crucial but challenging task. Many machine learning and system
+practitioners rely on vendor libraries or auto-schedulers to do the job. While
+the former requires significant engineering efforts, the latter in TVM only supports
+static-shape workloads in existing works. It is difficult, if not impractical,
+to apply the existing auto-scheduler directly to **dynamic-shape workloads**, as
+this leads to extremely long tuning time.
+
+We observe that the key challenge faced by existing auto-schedulers when
+handling a dynamic-shape workload is that they cannot construct a conclusive search
+space for all the possible shapes of the workload, because their search space is
+shape-dependent. To address this, this RFC aims to add dynamic-shape supports to
+AutoTIR by integrating DietCode framework, which constructs **a shape-generic
+search space and cost model** to auto-schedule dynamic-shape workloads
+efficiently.
+
+Our evaluation shows that DietCode has the following key strengths when
+auto-scheduling an entire model end-to-end: 
+
+1. reduces the auto-scheduling time by up to 5.88x less than the current
+   auto-scheduler on 8 uniformly sampled dynamic shapes, and
+1. improves performance by up to 69.5% better than the auto-scheduler and 18.6%
+   better than the vendor library. All these advantages make DietCode an
+   efficient and practical solution for dynamic-shape workloads.
+
+
+# Guide-Level Explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The existing experiments are largely conducted with auto-scheduler. However,
+having been syncing with the AutoTIR team for quarters, we plan to integrate
+this RFC to MetaSchedule (AutoTIR), because it provides more systematic
+interface and cleaner integration path with less hacks.
+
+To provide an example of additional information users are required to feed the
+system:
+
+```python
+# A symbolic shape constraint
+T = tir.ShapeVar('T’)
+# The candidate values of `T`
+T_vals = list(range(1, 128))
+
+task = Task(func=Dense,
+            args=(16*T, 768, 2304),
+            shape_vars=(T,),
+            wkl_insts=(T_vals,)
+            wkl_inst_weights=([1. for _ in T_vals],))
+```
+
+To enable auto-scheduling for dynamic shape workloads, users only need to:
+1. Have `ShapeVar` in the TE/TensorIR compututation.
+2. Specify the weight/distribution of each shape value.
+
+Notes:
+1. Symbolic constraint is required additional in Relay, but could be inferred
+   automatically after Relax is introduced;
+2. The proposed interface does not change any existing functionality.
+
+# Reference-Level Explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Here is an overview of the DietCode framework design.
+
+<img src="https://raw.githubusercontent.com/UofT-EcoSystem/DietCode/main/docs/figures/DietCode.jpg" width="61.8%" />
+
+- We construct **a shape-generic search space that consists of micro-kernels**,
+  an incomplete program that carries out a tile of the complete computation, to
+  efficiently support dynamic-shape workloads. 
+  
+  We use the hardware constraints (e.g., the maximum number of threads, the
+  amount of shared and local memory) rather than the shape information to
+  determine the micro-kernel candidates. Those candidates serve as the building
+  blocks and are executed repeatedly to carry out a workload instance (defined
+  as an static-shape instance of the dynamic-shape workload).
+- We build a **micro-kernel-based cost model**. The key insight is that the cost
+  of a complete program *P* that is made up of a micro-kernel *M* can be
+  decomposed into two parts: 
+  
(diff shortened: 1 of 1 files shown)


comaniac:
As one of the co-authors, it glad to see this RFC finally out :)
cc @masahi @ZihengJiang @Laurawly @Hzfengsy @MasterJH5574 @jinhongyii please help review and share your thoughts. Thanks.

junrushao1994:
We are more than happy to collaborate on AutoTIR side to make good things happen :-)

junrushao1994 (approved):
LGTM

areusch:
hey guys, thanks for the RFC! would you guys be up for doing a short walkthrough of it at a Community Meeting?

ZihengJiang (approved):

comaniac:
@ArmageddonKnight we might present/demo DietCode and introduce this RFC in a community meeting. Maybe in 5/25 or 6/1?

ArmageddonKnight:
@areusch @comaniac Sure. I would be happy to. I will be traveling on 6/1 so 5/25 looks good.

areusch:
@ArmageddonKnight @comaniac ok--let me do some agenda shuffling (we now have 3 topics for that meeting) and hopefully we can accommodate your travel schedule here.

areusch:
@junrushao1994 yep, we will post meeting notes here per the TVM Community Meeting [guidelines](https://docs.google.com/document/d/179l1WzqDiLXjjLmLopvxBSr0ylFS2plyXbeTtB8CteM/edit#heading=h.yjcjro6m0xi3).

areusch:
@ArmageddonKnight just confirming, we have on the agenda to discuss tomorrow

ArmageddonKnight:
@areusch BTW, may I ask whether there is anything I should prepare for the meeting? Alternatively, any particular thing that the community might be interested in? How much time am I given for the presentation?

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+1. Symbolic constraint is required additional in Relay, but could be inferred
+   automatically after Relax is introduced;
+2. The proposed interface does not change any existing functionality.
+
+# Reference-Level Explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Here is an overview of the DietCode framework design.
+
+<img src="https://raw.githubusercontent.com/UofT-EcoSystem/DietCode/main/docs/figures/DietCode.jpg" width="61.8%" />
```
areusch, line 88: could you describe each of these components in a bit better detail? here are some questions to prompt you
- which interface will you use to influence auto_scheduler's search space?
- is there additional metadata that's output from the compiler? if so, how is that communicated to the runtime?
- is the dispatcher fully contained within a PackedFunc that can be invoked from Graph/AoT/VM executors? Which have you tested with so far?

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+  ```CUDA
+  __global__ void default_function(float* X, float* W, float* Y,
+                                   const int T)
+                                   // Note the `T` here.
+  ```
+  Our evaluations indicate that this program has at least 5% worse performance
+  compared with the static-shape alternatives. Hence, we decide to sacrifice the
+  binary size for the runtime performance, which can potentially be problematic
+  when the hardware resources are limited.
+
```
areusch, line 129: could you also add to this section a description of how you plan to test this component?
- which parts of this should be done in CI vs which should be done offline? 
- do you require T4's to be introduced to CI to catch performance regressions? if not, what's the strategy for verifying this tuner in the long term?
- do we intend to do any of this testing pre-release? if so, how might you like the release process to look (i.e. we need a week before the TVM release is made final to do offline testing)?

areusch (commented):
adding a couple of questions here based on my understanding from the community meeting this morning

areusch:
We discussed this at the [TVM Community Meeting](https://discuss.tvm.apache.org/t/next-tvm-community-meeting-may-25/12807) this morning. Here are notes (thanks to mbs-octoml for taking these):
 - Most discussion was around whether the approach could help the related
   problem of long training time and a large number of distinct kernels
   for fully statically-shaped models. Eg resnet50 with its dozens of
   static shape instances for conv2d.

   It is useful to consider kernel compilation in two steps:
    - Start with the N instances of the same operator, with different
      shapes.
    - After tuning those independently, may end up with M << N schedules.
    - However, each schedule is instantiated for the original static shape
      so that no dimension variables are left behind at run time,
      thus we are back at N implementations.

   There's a significant perf impact to having only M kernel implementation
   which retain their parameterization on dimension variables. However, TVM
   could support not doing that inlining today.

   DietCode could be used to reduce M. The learned decision tree would dispatch
   from a fully shape polymorphic implementation to the specific schedule
   implementation. But at this stage the focus is on dynamic shape workloads.

 - No changes to Relay are required. Though Relay's dynamic shape functions
   support data-dependent output shapes, currently DietCode does not search
   over output shape and assumes only input shapes need to be inspected. But
   sounds like this could be addressed in future work.

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+            args=(16*T, 768, 2304),
+            shape_vars=(T,),
+            wkl_insts=(T_vals,)
+            wkl_inst_weights=([1. for _ in T_vals],))
+```
+
+To enable auto-scheduling for dynamic shape workloads, users only need to:
+1. Have `ShapeVar` in the TE/TensorIR compututation.
+2. Specify the weight/distribution of each shape value.
+
```
cgerum, line 77: Does DietCode have relay integration? If it does could you provide an example of how dynamic shapes, and wkl_insts are identified in relay. Otherwise this should probably mentioned in the current drawbacks and/or future possibilities.

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+# Upstream Milestones
+[upstream-milestones]: #upstream-milestones
+
+We propose the following milestones for upstreaming, where each bullet point
+corresponds to a PR with unit tests of roughly several hundred lines.
+
+- [ ] Code Generation Support
+  - Local Padding
+  - Loop Partitioning
+- [ ] Auto-Scheduler
```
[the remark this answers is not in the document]
ArmageddonKnight: Auto-Scheduler `->` Meta-Scheduler

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+
+```python
+# A symbolic shape constraint
+T = tir.ShapeVar('T’)
+# The candidate values of `T`
+T_vals = list(range(1, 128))
+
+task = Task(func=Dense,
+            args=(16*T, 768, 2304),
+            shape_vars=(T,),
```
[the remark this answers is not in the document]
ArmageddonKnight: Changed the example from single shape variable to multiple variables.

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+- Evaluate more operator use cases.
+- CPU Support
+
+# Upstream Milestones
+[upstream-milestones]: #upstream-milestones
+
+We propose the following milestones for upstreaming, where each bullet point
+corresponds to a PR with unit tests of roughly several hundred lines.
+
+- [ ] Code Generation Support
```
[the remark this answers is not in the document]
ArmageddonKnight: Will do.

rfcs/0072-dynamic-autoscheduler.md
```diff
@@ -0,0 +1,207 @@
...
+   auto-scheduler on 8 uniformly sampled dynamic shapes, and
+1. improves performance by up to 69.5% better than the auto-scheduler and 18.6%
+   better than the vendor library. All these advantages make DietCode an
+   efficient and practical solution for dynamic-shape workloads.
+
+
+# Guide-Level Explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The existing experiments are largely conducted with auto-scheduler. However,
```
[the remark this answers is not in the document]
ArmageddonKnight: The PoC design: https://github.com/UofT-EcoSystem/DietCode/tree/MLSys2022_AE is included in the updated version. Please refer to the `ops` folder for the workflow.

rfcs/0072-dynamic-autoscheduler.md
ArmageddonKnight: I have added more details to the workflow. Specifically
- The workflow for auto-scheduling dynamic-shape workloads will only be enabled if the frontend users provide `shape_vars` and `wkl_insts`. If these are not provided, then the current workflow will be used.
- There are metadata (i.e., `if` checks on shape variables that decide which kernel to be dispatched). Those are embedded as part of the host code.
- Yes. It is fully contained in a single `PackedFunc` and compatible with Relay.

rfcs/0072-dynamic-autoscheduler.md
ArmageddonKnight: We have added more details in the upstreaming section. We do NOT require any extra hardware and will be comparing the performance numbers delivered by our workflow on dynamic-shape workloads versus those delivered by the meta-scheduler on equivalent static-shape workloads. The expectation is that the difference between those numbers should be smaller than 5%.

rfcs/0072-dynamic-autoscheduler.md
ArmageddonKnight: It is possible for DietCode to be integrated as part of Relay. This is similar to the current auto-scheduling workflow: one needs to traverse through the Relay IR and records the workloads to be auto-scheduled and then export them to DietCode. We might add Relay integration after the full auto-scheduling workflow has been upstreamed.

ArmageddonKnight:
@areusch @cgerum Thank you for your feedback. I have addressed your feedback. Please review it again.

areusch (approved):
thanks for addressing my comments @ArmageddonKnight ! @cgerum could you take another look here? I think otherwise we should be good to merge

cgerum:
@areusch It's good to merge from my side

renfeier:
@ArmageddonKnight 
hello, i have a question looking forward to you reply.
Is it workable to use DietCode + vm compiler + vm runtime to implement graphlevel dynamic shape inference ?
do you have any suggestions?

ArmageddonKnight:
It is workable. The VM can gather the workloads and pack them as a single dynamic-shape workload and forward it to DietCode. DietCode does the auto-scheduling and send the generated code back to the VM.

renfeier:
> It is workable. The VM can gather the workloads and pack them as a single dynamic-shape workload and forward it to DietCode. DietCode does the auto-scheduling and send the generated code back to the VM.
Thank you very much for your prompt reply. :-D, b（￣▽￣）d

PR #74, opened 2022-05-25
[RFC] TUNIP: TVMScript Unified Printer

yelite:
[Rendered](https://github.com/yelite/tvm-rfcs/blob/tunip/rfcs/0074-tvmscript-unified-printer.md)

### DIFF
--- a/rfcs/0074-tvmscript-unified-printer.md
+++ b/rfcs/0074-tvmscript-unified-printer.md
@@ -2,6 +2,9 @@
 - Start Date: 05/25/2022
 - RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
 - GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+- Co-Authors: Lite Ye ([**@yelite**](https://github.com/yelite)), Yong Wu
+  ([**@yongwww**](https://github.com/yongwww)), Yuchen Jin
+  ([**@YuchenJin**](https://github.com/YuchenJin))
 
 # Summary
 [summary]: #summary
--- a/rfcs/0074-tvmscript-unified-printer.md
+++ b/rfcs/0074-tvmscript-unified-printer.md
@@ -1,0 +1,472 @@
+- Feature Name: TUNIP: TVMScript Unified Printer
+- Start Date: 05/25/2022
+- RFC PR: [apache/tvm-rfcs#0000](https://github.com/apache/tvm-rfcs/pull/0000)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes to modularize and infrastructuralize the existing TVMScript
+printer, to develop unified printing mechanism across TVM stack, where TIR,
+Relax and any future vendor-specific IR are all treated equally as dialects and
+could be printed together without potential conflict in engineering.
+
+# Motivation
+[motivation]: #motivation
+
+TVMScript, as a roundtrippable python-based text format, is
+the central piece of TVM performance productivity. As the frontend of TVM, it
+enables end users to directly construct the TVM IR, either TIR or Relax, in a
+pragmatic approach. From Relax to MetaSchedule and TIR, TVMScript enables
+inspectability and reproducibility at any level of compilation and
+optimization. Furthermore, based on TVMScript, developers are empowered to
+intercept, manipulate and customize the compiler behavior in a principled way.
+
+While TVMScript is gaining traction and buy-in from the open source community,
+the TVMScript printer suffers from multiple profound design issues:
+- Not supporting IR fragment printing requires users to jump in-between
+  TVMScript syntax and TIRText syntax 
+- The lack of modularity leads to practical inability to scale up to and
+  maintain multiple IRs without engineering conflicts 
+- Enhancing co-existence of multi-level IRs often leads to re-engineering of
+  existing features.
+
+**Goal.** This RFC introduces Tvmscript UNIfied Printer (TUNIP), a systematic
+redesign to
+address those engineering, usability and scalability issues above. The goal of
+this re-design includes:
+
+**Goal 1 [Unified Representation].** Become the unified roundtrippable
+representation of TIR and Relax, allowing systematic mixing of IRs or IR
+fragments (Relax + TIR) in the same IRModule in the target language (for
+example, python, C++). 
+
+Currently TVMScript priner is designed specifically for TIR, and printing
+multiple dialects together was not a design goal at that time. Therefore,
+supporting Relax requires ad-hoc hack around the system (for
+instance, [relax#149](https://github.com/tlc-pack/relax/pull/149) added support
+of printing `T.cast` and `T.max` in an ad-hoc way, without reusing the printing
+code for TIR). The unified printer in this RFC addresses this issue by having a
+unified approach for printing IR tree to TVMScript. Engineers will be able to
+implement a fully-fledged printer for Relax, TIR and any potential IR in the
+future with minimal effort.
+
+The folder structure that we want to pursue is:
+```bash
+include/tvm/script/printer/
+└── ... # Public headers for the core infra
+src/script/printer/
+├── core # Core infra, which is IR-agnostic
+│   ├── ir_docsifier.cc
+│   └── ...
+├── tir # TIR dialect 
+│   ├── expr.cc
+│   ├── stmt.cc
+│   └── ...
+└── relax # Hypothetical Relax dialect (not part of our RFC)
+    └── ...
+```
+
+**Goal 2 [Third-Party IRs in Multi-Stage Compilation].** Modularize and
+infrastructuralize the printer to support more future IRs or third-party IRs at
+any level with maintainability, for example, IRs at lower-level than TIR, or
+Relax VM executable.
+
+The current TVMScript printer is tightly coupled with TIR by being a subclass
+of TIR-specific functors
+([link](https://github.com/apache/tvm/blob/main/src/printer/tvmscript_printer.cc#L129)).
+This design isn’t scalable when we want to support more IRs. More importantly,
+it’s impossible for the current approach to support third-party IR bteing
+registered in a dynamic library.
+
+**Goal 3 [Reproducibility and Error Reporting].** Expand reproducibility and
+flexible rendering of diagnostic messages during any level of IR
+transformation.
+
+For example, the following snippet runs and produces an error.
+
+```py
+import tvm
+
+@T.prim_func
+def func_a(A: T.Buffer[(1,), "int32"]):
+    A[0] = 0
+
+@T.prim_func
+def func_b(A: T.Buffer[(8,), "int32"]):
+    A[0] = 0
+
+tvm.ir.assert_structural_equal(func_a, func_b)
+```
+
+The current error message indicates what the difference was, but not where it
+occurred.  This can sometimes be inferred from a stack trace, but becomes
+increasingly difficult with larger IR graphs.
+
+```
+ValueError: StructuralEqual check failed, caused by lhs:
+1
+and rhs:
+8
+```
+
+TUNIP should enable individual utilities and IR passes to have error messages
+directing the user to exact locations in the IR representation.
+
+```
+ValueError: StructuralEqual check failed, first delta highlighted below
+
+@T.prim_func
+def func_a(A: T.Buffer[(1,), "int32"]) -> None:
+                       ^^^^
+    A[0] = 0
+
+@T.prim_func
+def func_b(A: T.Buffer[(8,), "int32"]) -> None:
+                       ^^^^
+    A[0] = 0
+```
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This section introduces the design philosophy of the printer, and demonstrates
+the proposed user-facing APIs where users means IR developers.
+
+## Two-Stage Translation
+
+Traditionally in TVM stack, printing is a single-stage process. The printer
+assumes certain syntax of the target language, and therefore, so far there are
+3 different printers all for TIR: ReprPrinter, TIRTextPrinter,
+TVMScriptPrinter.
+
+We extend the idea of the existing Doc class at
+[src/printer/doc.h#L67](https://github.com/apache/tvm/blob/main/src/printer/doc.h#L67)
+to allow better consistency and scalability. An IR, which could
+be TIR, Relax or any other ones developed by third-party vendors, is first
+translated to an intermediate Doc node tree, and then the Doc tree is mapped to a target
+language, for example, Python, C++ IRBuilder API, or Rust.
+
+**Stage 1 [TVM IR => Doc]**. On the first stage, the printer needs to take care
+of translating a TVM IR to Doc tree. As an example, `tir.For` is translated to
+`ForDoc` without having to worry about the underlying language. Note that some
+complicated nodes in TVM IR, for example, `PrimFunc`, could be translated to
+multiple IR elements, including `FunctionDoc` and a few `StmtDoc`.
+
+During the translation from IR to Doc tree, it is possible that some statement
+influences the syntax of its children or vice verse, especially for syntactic
+sugars and declaring undefined variables in IR fragment printing. Therefore, a
+generic data structure `Frame` is introduced to allow retrieval and
+manipulation the relevant context information.
+
+**Stage 2. [Doc => target language]**. On the second stage, Doc tree is then
+honestly translated to the target language in text format. For example, when
+the target language is python, `ForDoc` is translated to python’s for loop
+syntax:
+
+```python
+for ... in ...:
+  ...
+```
+
+When the target language becomes python IRBuilder, `ForDoc` is translated to:
+
+```cpp
+with T.For(...):
+  ...
+```
+
+For generality, the Doc tree is designed to select minimal elements that exist
+in languages used in developing TVM. A full spec of the Doc could be found in
+the next section.
+
+## Distributed Registration
+
+As a major engineering challenge for TVMScript to scale to multiple IRs, the
+existing printing logic has to be engineered, maintained and re-engineered in a
+single file, which has brought significant confusion for developing multi-level
+IRs for TVM Unity.
+
+Inspired by the pass infrastructure, as well as the ReprPrinter in TVM, we
+propose to develop the infrastructure to enable distributed registration, and
+further allows printer for different levels of IR to be registered in separate
+translation units, and in the meantime keeps the capability to be mixed
+together at various level, for example, Relax uses TIR expression in its
+function bodies, and TIR calls back to Relax function.
+
+## Diagnostics and Reproducibility
+
+Existing error reporting mechanisms have not taken IR structure and
+reproducibility into consideration. Usually it reports a single line error
+message without providing necessary context of how the IR looks like during
+compilation. For example, when comparing whether two TIRs are structurally
+equivalent, the system may report:
+
+```cpp
+ValueError: StructuralEqual check failed, caused by lhs:
+{slow_memory_3_var: buffer(slow_memory_3_buffer_var, 0x501bf80), fast_memory_2_var: buffer(fast_memory_2_buffer_var, 0x501bd80), placeholder_3: buffer(placeholder_5, 0x50138a0), placeholder_2: buffer(placeholder_4, 0x5012b60), T_subtract: buffer(T_subtract_1, 0x5014390)}
+and rhs:
+{}
+```
+
+which lacks necessary information for users to understand where the mismatch
+is.
+
+As a recent effort, structural error reporting in TIR scheduling provides
+relevant and reproducible context, as demonstrated below:
+
+```cpp
+@tvm.script.ir_module
+class Module:
+    @tir.prim_func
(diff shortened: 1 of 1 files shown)


rfcs/0074-tvmscript-unified-printer.md
```diff
@@ -0,0 +1,475 @@
...
+- For each IR node that has diagnostic message, trace back through its parent
+  until it reaches to an IR node in the map collected in previous step. Then it
+  can produce a map from Doc node to diagnostic message.
+- In the 2nd stage, diagnostic message will be printed as doc is being printed
+  into target language
+
+# Reference-level explanation
+
+## Doc Spec
+
```
wrongtest, line 262: How to represent the type and nested types in the doc nodes ?

yelite: Types will be represented by `ExprDoc`. For example, `Tensor((2, 3), "float32")` will be 

```
CallDoc(
  IdDoc("Tensor"),
  [
    TupleDoc([
       LiteralDoc(2),
       LiteralDoc(3)
    ]),
    LiteralDoc("float32")
  ]
)
```

rfcs/0074-tvmscript-unified-printer.md
```diff
@@ -0,0 +1,475 @@
...
+TVMScriptPrinter.
+
+We extend the idea of the existing Doc class at
+[src/printer/doc.h#L67](https://github.com/apache/tvm/blob/main/src/printer/doc.h#L67)
+to allow better consistency and scalability. An IR, which could
+be TIR, Relax or any other ones developed by third-party vendors, is first
+translated to an intermediate Doc node tree, and then the Doc tree is mapped to a target
+language, for example, Python, C++ IRBuilder API, or Rust.
+
+**Stage 1 [TVM IR => Doc]**. On the first stage, the printer needs to take care
```
areusch, line 154: just to spell it out, what's the rationale for translating into Doc tree if we have to then do another step of translating the Doc tree into a target frontend language? will the Doc tree be reused e.g. in the parser? are there common operations we can handle in translating into the Doc tree?

rfcs/0074-tvmscript-unified-printer.md
```diff
@@ -0,0 +1,475 @@
...
+honestly translated to the target language in text format. For example, when
+the target language is python, `ForDoc` is translated to python’s for loop
+syntax:
+
+```python
+for ... in ...:
+  ...
+```
+
+When the target language becomes python IRBuilder, `ForDoc` is translated to:
```
areusch, line 176: just curious, do we have an example of using a different target language other than Python?

rfcs/0074-tvmscript-unified-printer.md
areusch: okay but i'm not quite sure i follow how we are splitting up the printing operations here when we are considering supporting multiple frontend languages. is it like:
1. Doc tree represents a sort of AST that captures things like scope nesting which may affect indent, without specifying how to serialize ForDoc
2. Doc tree -> language translation step prints ForNode in the language proper?

how do we handle e.g. Python indentation when translating from Doc -> Python? is it assumed that the IRModule -> doc translation always sets indentation that's compatible with Python?

rfcs/0074-tvmscript-unified-printer.md
```diff
@@ -0,0 +1,475 @@
...
+example, python, C++). 
+
+Currently TVMScript priner is designed specifically for TIR, and printing
+multiple dialects together was not a design goal at that time. Therefore,
+supporting Relax requires ad-hoc hack around the system (for
+instance, [relax#149](https://github.com/tlc-pack/relax/pull/149) added support
+of printing `T.cast` and `T.max` in an ad-hoc way, without reusing the printing
+code for TIR). The unified printer in this RFC addresses this issue by having a
+unified approach for printing IR tree to TVMScript. Engineers will be able to
+implement a fully-fledged printer for Relax, TIR and any potential IR in the
```
sunggg, line 54: Is there any requirement for third-party IR to be supported by TUNIP?

rfcs/0074-tvmscript-unified-printer.md
```diff
@@ -0,0 +1,475 @@
...
+## Distributed Registration
+
+As a major engineering challenge for TVMScript to scale to multiple IRs, the
+existing printing logic has to be engineered, maintained and re-engineered in a
+single file, which has brought significant confusion for developing multi-level
+IRs for TVM Unity.
+
+Inspired by the pass infrastructure, as well as the ReprPrinter in TVM, we
+propose to develop the infrastructure to enable distributed registration, and
+further allows printer for different levels of IR to be registered in separate
```
sunggg, line 196: I'm not quite sure if I understand this correctly. Can you give me any example?

sunggg (commented):
Thank you for the great RFC!
Just a few questions.

rfcs/0074-tvmscript-unified-printer.md
junrushao1994: Yes. Example target languages could be python/C++ IRBuilder we will develop later on for TVMScript metaprogramming (which is not part of the RFC though). Conceptually it's something like:

```python
with Builder():
  with T.prim_func(name="main"):
    A = T.arg(T.Buffer((128, 128), "float32"))
    B = T.arg(T.Buffer((128, 128), "float32"))
    with T.grid(128, 128, 128) as i, j, k:
      with T.block(name="block"):
        vi = T.axis.S(128, i)
        vj = T.axis.S(128, j)
        vk = T.axis.R(128, k)
```

The initial point of having two-stage translation though is to match the existing behavior, not anything new

YuchenJin (approved):
LGTM! The unified representation across multiple dialects and better error reporting will greatly help the development of Relax and future IRs. 

Thank you @yelite and @junrushao1994 for the design and RFC!

rfcs/0074-tvmscript-unified-printer.md
yelite: I added a section for the rationale of translating into Doc tree in RFC. Briefly speaking, two-stage printing is more scalable if printer needs to support more than one IR, because it separates the logic to handle IR semantics with the logic to handle target language syntax. 

<details>
  <summary>Code Example</summary>
  
To print buffer region (things like `A[1:2, 3, 4]`) in the current TVMScript printer,
```cpp
Doc TVMScriptPrinter::PrintBufferRegion(const BufferRegionNode* op) {
  Doc doc;
  if (op->region.size() == 0) {
    doc << Print(op->buffer) << "[()]";
  } else {
    doc << Print(op->buffer) << "[";
    for (size_t i = 0; i < op->region.size(); ++i) {
      if (i != 0) doc << ", ";
      const auto& range = op->region[i];
      if (!is_one(range->extent)) {
        doc << Print(range->min) << " : " << Print(ana_.Simplify(range->min + range->extent));
      } else {
        doc << Print(range->min);
      }
    }
    doc << "]";
  }
  return doc;
}
```

while in the unified printer with two-stage printing

```cpp
ExprDoc PrintBufferRegion(tir::BufferRegion buffer_region, IRDocsifier p) {
  Array<Doc> indices;

  for (const Range& range : buffer_region->region) {
    if (tir::is_one(range->extent)) {
      indices.push_back(p->AsExprDoc(range->min));
    } else {
      indices.push_back(p->AsExprDoc(range));
    }
  }

  return p->AsExprDoc(buffer_region->buffer)->Index(indices);
}
```

</details>

yelite: @areusch To answer your questions,

> is it assumed that the IRModule -> doc translation always sets indentation that's compatible with Python?
The IR -> Doc translation doesn't know about indentation. It just sets the right values to Doc without knowing any detail on target language syntax. For example, ForDoc is defined as
```cpp
class ForDocNode : public StmtDocNode {
 public:
  ExprDoc lhs{nullptr};
  ExprDoc rhs{nullptr};
  Array<StmtDoc> body;
}
```

In the Doc -> language translation, indentation is added when it processes the `ForDoc->body`.

rfcs/0074-tvmscript-unified-printer.md
yelite: This means the printer for TIR and Relax can be defined in different `cc` files and doesn't know the existence of each other. For example, the folder structure of TIR printer and Relax printer will be like
```bash
src/script/printer/
├── core # Core infra, which is IR-agnostic
├── tir # TIR dialect 
│   ├── expr.cc
│   └── ...
└── relax # Hypothetical Relax dialect (not part of our RFC)
    └── ...
```

An implication is that printer for an IR can be loaded as dynamic library, which is a building block to support third-party IR.

sunggg: Interesting. Thank you for the clarification! Cool work!

junrushao1994 (approved):

areusch:
very sorry for the delay here.

@junrushao1994 i guess i'm still a little fuzzy on the rationale for why we want to have two ways to parse TVMScript. I know this is a bit bigger than just this RFC, but with this RFC we're further pushing the cart down this path so I'd like us to be explicit about why.

Relay has a single roundtrippable serialization format, as do most languages. I think we benefit from this in that we only have one set of tests to maintain.

With TIR, my understanding is that we used Python as an AST parser since language constructs in TIR somewhat closely matched to Python. I'm not sure I see why we should want to have two different ways to do this--shouldn't we just have a single way to serialize TIR and parse it that always works? It seems like we could still expose this parser to various frontends via PackedFunc.

junrushao1994:
> Relay has a single roundtrippable serialization format, as do most languages. I think we benefit from this in that we only have one set of tests to maintain.

To clarify, Relay has two roundtrippable serialization formats: text format and json. people use text format for readability, and in most usecases, go for json reliable serialization.

> why we want to have two ways to parse TVMScript

First of all, the RFC focuses on the printer, so would be great if we talk about the parser in the incoming RFC, but to clarify, we are not having two ways to parse TVMScript in the same language. Instead, for metaprogramming capability, one may need to interleave host language-based IRBuilder with the parser.

> language constructs in TIR somewhat closely matched to Python.

To clarify, TIR is a DSL independent of any existing language. It's somewhat close to HalideIR but has been evolving for years with lots of new features, for example, blocks. Python syntax is an add-on for usability but TIR design is in no ways matched to python.

Anyways, I'm happy to talk more on the upcoming metaprogramming RFC.

Hzfengsy (approved):
LGTM. Thanks @yelite !!!

areusch:
> people use text format for readability, and in most usecases, go for json reliable serialization.

what do you mean by "reliable" here? if they're truly roundtrippable, aren't they both reliable? just trying to understand :)

> Instead, for metaprogramming capability, one may need to interleave host language-based IRBuilder with the parser.

Is the goal then that you'd be able to serialize an IRModule which contains fragments that need to be repeated via meta-programming? then, if you did this using a host language's IRBuilder, you could quickly write some meta-programming steps in that host language to expand the program.

>To clarify, TIR is a DSL independent of any existing language. ... Python syntax is an add-on for usability but TIR design is in no ways matched to python.
I agree with your statement, but I thought the repr form of TIR wasn't so easy to work with and thus the motivation for TVMScript. While I agree that TIR is technically separate, if TVMScript is a core way in which TIR is used, I'd argue we should treat them conceptually as joined (e.g. TVMScript as the recommended roundtrip text format for TIR). What are your thoughts there?

junrushao1994:
Thanks @areusch for your response!

> if TVMScript is a core way in which TIR is used, I'd argue we should treat them conceptually as joined (e.g. TVMScript as the recommended roundtrip text format for TIR). What are your thoughts there?

Let's phrase it this way: TVMScript serves as a frontend of easily constructing and manipulating TIR, Relax and any third-party IRs. It offers readability, type checking, metaprogramming, etc. In terms of serialization, my standing is that reflection-based JSON serialization is still the best way to go as it's the least error-prone if we don't need readability and manipulability.

> what do you mean by "reliable" here? if they're truly roundtrippable, aren't they both reliable? just trying to understand :)

In practice we did notice that there could be some ad-hoc issues (which are definitely fixable, not a big deal, so I'm not saying it's not good) that probably be overlooked when implementing the parser. For example, @mbs-octoml fixed a dtype issue very recently (https://github.com/apache/tvm/pull/11224). Instead, reflection-based JSON serialization mechanism in TVM is definitely less bug-prone because it doesn't need to handle edge cases - the downside is that it's almost unreadable though.

> Is the goal then that you'd be able to serialize an IRModule which contains fragments that need to be repeated via meta-programming? then, if you did this using a host language's IRBuilder, you could quickly write some meta-programming steps in that host language to expand the program.

Yeah with metaprogramming we are able to write bunch of code that require host-language control. The next RFC is all about metaprogramming so stay tuned :-)

junrushao1994:
To summarize offline discussion with @areusch:

Q: Is this going to unify current fragmented printing formats?
A: Yes for TIR. After following the standard deprecation procedure (see Section "Upgrade Plan"), TVMScript will be the only frontend (i.e. user-facing printer/parsing) for TIR, while the original non-roundtrippable text format will eventually be deprecated. For serialization/deserialization, reflection-based JSON format is preserved without change, which is basically like protobuf for reliably data sharing.

Q: Is JSON serialization going to be deprecated?
A: No. It is a separate issue. TVM's JSON serialization is conceptually similar to protobuf, which is designed to be reliably persist data, while not necessarily human-readable, which is different than our goal in this RFC (frontend).

Q: Is this going to introduce new functionalities that make TVM more fragmented?
A: No. It's a refactoring of existing TIR TVMScript printer. With future deprecations done, it actually reduces the fragmentation on TIR side.

Q: Why is this called "unified" printer?
A: It provides a unified infrastructure that allows IR developers to extend to other IRs, for example, Relax, without interfering with each other. More concretely speaking, as pointed out in the folder structure in the RFC text:

```
src/script/printer/
├── core # Core infra, which is IR-agnostic
│   ├── ir_docsifier.cc
│   └── ...
├── tir # TIR dialect 
│   ├── expr.cc
│   ├── stmt.cc
│   └── ...
└── relax # Hypothetical Relax dialect (not part of our RFC)
    └── ...
```

TIR developers only need to maintain the `tir` folder, and relax developers only need to maintain `relax` folder. The IRModule that contains both TIR and other dialects is allowed to be printed with this infrastructure.

areusch:
thanks for summarizing @junrushao1994 . to provide some additional perspective:
- I'd really like us to get to a place where we have 1 reliable and readable format to serialize TIR. i don't think we should build all of this infrastructure just to provide a "best-effort" serialization--anything expressable in TIR should roundtrip in any serialization format we provide. there's no reason we have to sacrifice reliability for readability, either--it's just a matter of putting effort into the unit tests.
- right now i believe we have 3 TIR formats: `repr()`, TVMScript, and JSON. This RFC looks to provide infra that allows for generation of more formats e.g. so that Python doesn't have to be the frontend for TIR, or so we can leverage other APIs or frontend languages to write TIR. i'm okay with adopting this design to separate indentation from serialization, but if we do follow through and introduce an IRBuilder serialization, that would count as a fourth way to serialize TIR--doing that is only going to make it harder to achieve the previous goal because it splits our human and unit-test attention across multiple formats. i don't see a good reason for doing that laid out in the RFC.
- this PR does provide considerations for Relax, although the language spec hasn't been RFC'd or introduced yet. i don't want to block the PR over that, but it's plain that the eventual goal is to leverage this infrastructure for Relax when it lands. then that gets us into a scenario where we have one parser/serializer for Relay and another for everything else. that's also more complex than it needs to be. admittedly, Relax and TIR benefit more from meta-programming than Relay, so it makes sense that the old Relay parser/serializer isn't a great fit here. but if we're going to discuss Relax here, it would be great to see this design considering how it could replace all of the printers in the codebase, not just overhaul the TIR one. anyhow, this discussion can be left for a future time when the Relax merge plan is clearer.

areusch (approved):

junrushao1994:
Hey @areusch thanks for elaborating your points and these are all definitely great questions to me!

> right now i believe we have 3 TIR formats: repr(), TVMScript, and JSON. This RFC looks to provide infra that allows for generation of more formats e.g. so that Python doesn't have to be the frontend for TIR, or so we can leverage other APIs or frontend languages to write TIR

Oh I think I get your concern here!!

No, we are not building more and more and more frontends to cause fragmentation. Here is the thing - we only need two forms for serialization:
- JSON for protobuf-like reliable serialization;
- TVMScript python syntax.
And we defragment the system by retiring `repr()` for TVMScript. Therefore, as a python user, the only interface you touch is TVMScript python syntax after the RFC landed.

Imagine you are a rust user who doesn't want to use python, our proposal makes it possible to develop a frontend in pure rust. And of course, it's not going to be our priority, i'm just stating the possibility.

> but if we do follow through and introduce an IRBuilder serialization, that would count as a fourth way to serialize TIR

Ah no, IRBuilder is not serialization, sorry for the confusion. Let's imagine a common usecases in company's prod env where python is strictly prohibited (e.g. on-device training), but our developers wrote most of the operators in python, which is the reality. How do we quickly migrate those operators to C++? Our RFC provides an answer to that - just write a codegen from Doc to C++ IRBuilder :-)  Of course again, the RFC aims to open possibilities instead of implementing those functionalities.

> this PR does provide considerations for Relax, although the language spec hasn't been RFC'd or introduced yet. i don't want to block the PR over that, but it's plain that the eventual goal is to leverage this infrastructure for Relax when it lands

Yeah it's definitely fair to say the PR considers Relax a lot. On the other hand, Relax doesn't need this RFC to use TVMScript. Take the current codebase as an example, it's not using the design of this proposal and it's totally fine given it's assuming there are only 2 IRs. Instead, I'm trying to convey is that we build an open infra for any vendors to integrate with their own IRs, which can be represented and compiled in the same IRModule with Relax, TIR, etc.

areusch:
@junrushao1994 ah thanks for clarifying! a few follow-ups then:

> Imagine you are a rust user who doesn't want to use python, our proposal makes it possible to develop a frontend in pure rust. And of course, it's not going to be our priority, i'm just stating the possibility.

Does this mean you can import the JSON-based serialization using only libtvm.so (e.g. rust could import JSON)? If so, then, in the hypothetical prod scenario:

>Let's imagine a common usecases in company's prod env where python is strictly prohibited (e.g. on-device training), but our developers wrote most of the operators in python, which is the reality. How do we quickly migrate those operators to C++?

, what's the motivation for someone to use IRBuilder instead of just serializing the TVMScript to JSON via parse/print and then ingesting the JSON in their prod env using the JSON parser?

> it's not using the design of this proposal and it's totally fine given it's assuming there are only 2 IRs. Instead, I'm trying to convey is that we build an open infra for any vendors to integrate with their own IRs, which can be represented and compiled in the same IRModule with Relax, TIR, etc.

That makes sense, thanks for clarifying :)

junrushao1994:
@areusch Thanks for following up!

> what's the motivation for someone to use IRBuilder instead of just serializing the TVMScript to JSON via parse/print

JSON is mostly for serializing an IR after it's constructed (which users cannot manipulate), and the TVMScript format is for users to construct the IR themselves. For example, to implement a ReLU operator, TVMScript users could write a python/rust code, while it's pretty infeasible for them to write JSON with bare hand.

junrushao1994:
Hey I'm happy to discuss more, and let's keep this RFC open until the end of this week

areusch:
@junrushao1994 ok, just want to clarify one question here:
>  TVMScript users could write a python/rust code, while it's pretty infeasible for them to write JSON with bare hand.
what functionality does JSON provide that TVMScript doesn't do right now as it's designed? Is there a reason to have two formats to begin with? imo it is not worth having two formats here unless there's a clear need for structural analysis of TIR outside of TVM.

i have heard some folks in the community interested in converting TIR to JSON (I think @SebastianBoblestETAS and @MichaelJKlaiber if memory serves), but I also think that the use cases they're interested in could be accomplished if we had a user-friendly, robust Python interface to the TIR.

junrushao1994:
Thanks @areusch for pointing me to the thread! Definitely happy to read the [discussion](https://discuss.tvm.apache.org/t/export-tir-to-json/12329), and glad to see that @vinx13 unblocks the `SaveJSON` method for NDArrays :-) completely agreed that TVMScript could be the usecase which provides readability if this is the case that @SebastianBoblestETAS and @MichaelJKlaiber wanted. 

> what functionality does JSON provide that TVMScript doesn't do right now as it's designed? Is there a reason to have two formats to begin with? imo it is not worth having two formats here unless there's a clear need for structural analysis of TIR outside of TVM.

Yeah JSON doesn't provide any extra functionality, but it's pretty stable and reliable protobuf-like serialization, which could be faster to parse/dump if we only want to save the IRs or send them around between processes without the need of readability and manipulatability.

For general users, our suggestion should be like, just use TVMScript to write TIR and forget about JSON. For advanced developers, for example, we want to store TIR in tuning database, cache TIR somewhere for later usage, JSON is definitely the "to-go" approach.

yelite:
> what functionality does JSON provide that TVMScript doesn't do right now as it's designed? Is there a reason to have two formats to begin with? imo it is not worth having two formats here unless there's a clear need for structural analysis of TIR outside of TVM.

A good analog is the ProtoBuf [text format](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/text_format.cc) vs. wire format (binary format). People write proto in text format when they need to create proto manually, while the binary format is used in RPC payload for its portability and efficiency.

Back to our own use case, TVMScript is the choice if user want to construct IR graph manually, but it's not portable because the TVMScript parser has a strong dependency on Python, especially after the Metaprogramming RFC (#79). On contrast, the JSON format will be portable because it only depends on the IR node definition in C++.

PR #78, opened 2022-06-09
Add Target Features RFC

Mousius:


areusch:
cc @kparzysz-quic @tqchen @jwfromm @masahi @junrushao1994 @comaniac

rfcs/0078-target-features.md
```diff
@@ -0,0 +1,178 @@
...
+```python
+isa = arm_isa.IsaAnalyzer(target)
+if isa.has_dsp_support:
+    do_dsp_stuff()
+```
+
+The `Target` can be directly inspected:
+
+```python
+if target.features.dsp:
```
areusch, line 55: should we adopt any namespacing? if so how would we represent that with this syntax?

areusch (commented):
@Mousius added a question. would you like to discuss at a Community Meeting?

Mousius:
> @Mousius added a question. would you like to discuss at a Community Meeting?

Happy to do a quick update on where this went 😸

Mousius:
> If we add another member to `Target`, how will this interact with the [target parser proposal](https://github.com/apache/tvm-rfcs/pull/71)?

The hope is to use the target parser to parse out the features as it'd just be an additional field we can set:

```c++
target_json.Set("features", features_map);
```

rfcs/0078-target-features.md
Mousius: I'm hopeful for the initial pass we can work with `target.features.X` rather than further namespacing - most of what I've seen in other compilers is a flat array seems to work.

areusch (approved):
let's call this accepted since we're implementing it :)

PR #81, opened 2022-06-23
[USMP] Update RFC with constants pools

manupak:
This commit introduces the notion of constant memory pools
and removes the need to define access on each of the targets.

### DIFF
--- a/rfcs/0009_Unified_Static_Memory_Planning.md
+++ b/rfcs/0009_Unified_Static_Memory_Planning.md
@@ -461,19 +461,43 @@ Inputs :
 * IRModule containing 
     * AoT TIR PrimFunc (the control function describing the call graph to operators)
     * All Operator Functions
-    * Each tir.allocate in the IRModule annotated with candidate pools ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md))
+    * Each tir.allocate in the IRModule annotated with candidate workspace pools ([Using the annotation field of tir.allocate](https://github.com/apache/tvm-rfcs/blob/c447cbfbd5abceaa7623a0f90cc492784e6f0c0b/rfcs/0023-adding-annotation-field-to-tir.allocate.md))
 
 
         ```
+        /*!
+         * \brief base class pool of memory accessible by one or more targets.
+         * This is extended to represent workspace pools and constant pools.
+         * (See below)
+         */
         struct PoolInfoNode : public Object {
-        String  pool_name;
-        Integer size_bytes;
-        Integer alignment;
-        Integer pool_offset;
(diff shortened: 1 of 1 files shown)


manupa-arm:
cc : @areusch

rfcs/0009_Unified_Static_Memory_Planning.md
```diff
@@ -461,19 +461,43 @@ Inputs :
...
+            Map<Target, Integer> target_burst_bytes;
+            /*! \brief The targets linked to the pool */
+            Array<Target> targets;
         }
+
+        struct WorkspacePoolInfoNode : public PoolInfoNode {
+        };
         ```
-        The input IRModule is expected to have "candidate_memory_pools" annotation populated with a orderered list of PoolInfo objects. The ordering will indicate to the planner the order of preference each allocate will be pinned to each Pool. The core compiler will run a pass to assign each tir.allocate with candidate_memory_pools based on the target each PrimFunc is executed, prior to invoking the USMP.
+        The input IRModule is expected to have "candidate_memory_pools" annotation populated with a orderered list of WorkspacePoolInfo objects. The ordering will indicate to the planner the order of preference each allocate will be pinned to each Pool. The core compiler will run a pass to assign each tir.allocate with candidate_memory_pools based on the target each PrimFunc is executed, prior to invoking the USMP.
```
areusch, line 500: not sure if we should wade into the ordering/prioritization problem here. i mean we could, but it does seem like a bit of a can of worms. we could also solve this by customizing the allocator. just curious if you have a use case in mind?

areusch (approved):
thanks @manupa-arm , just a question here which i might have missed before

rfcs/0009_Unified_Static_Memory_Planning.md
manupa-arm: thanks @areusch for taking a look.

The ordering/prioritization is not in the scope of this change.

In fact, it is something already supported : https://github.com/apache/tvm/blob/395e91ff54543864a90240d18c8efd8c277c758b/tests/python/unittest/test_tir_usmp_algo.py#L366-L413

However, the USMP will only try to solve the allocation problem not the multi-objective problem of memory and performance. In other words, the scheduling stage is responsible for assigning the right pool for each allocate node. However, if the scheduling does not care about what pools it gets allocated, USMP will try to do a prioritized/ordered consumption of the pools to assign all allocated nodes to memory pools.

areusch: ok, that makes sense to me

areusch: only request--shall we document the above somewhere? maybe in a file-level docstring for USMP?

manupa-arm: Sure, happy to do that in a follow up

areusch:
i have no concerns here, @manupa-arm feel free to submit this when you're ready to start

PR #84, opened 2022-06-29
[RFC] Name mangling in IRModules

gigiblender:
[Rendered](https://github.com/gigiblender/tvm-rfcs/blob/mangling-rfc/rfcs/0077_name_mangling_ir_modules.md)


gigiblender:
> Thanks @gigiblender ! a few points for consideration:
> 
>     * Ideally NameSupply should interact well with our current linkage spec, e.g. if a function have an attribute `global_symbol`, it means the name is "final"(because external users are expecting to look it up with that name), likely we need to have global. Anything that does not have a `global_symbol` attr can subject to change.
> 
>     * Would be useful to initialize a NameSupply from an existing IRModule, so it does not need to be carried through out passes and is only needed for pass that regenerate global vars.
> 
> 
> In addition to that, it would be really nice to get some smart naming resolve mechanism.
> 
> For example, it would be really nice to have the following effect (by parsing the suffix integer separately from prefix string
> 
> ```python
> x = get_next_name("fun12", existing_names=["func12", "func13", "func4"])
> assert x == "func14"
> ```
> 
> This is to avoid the case where we have "func12_0_0_0_0" when multiple set of prefixes are called and they usually looks confusion.

Thank you for feedback @tqchen! I agree with your suggestions and amended the text with further clarifications.

areusch:
cc @mbs-octoml @Mousius @manupa-arm @MichaelJKlaiber @grant-arm @kparzysz-quic

rfcs/0077_name_mangling_ir_modules.md
```diff
@@ -0,0 +1,102 @@
...
+The `NameSupply` will be implemented as a class in TVM following the usual design, extending `Object` and `ObjectRef`, respectively.
+As mentioned, the `NameSupply` will contain an internal map of `String -> GlobalVar` used to perform lookups by `UniqueGlobalFor`. Additionally, a map of `String -> Int` will be used to provide unique names for `FreshGlobal`.
+
+Additional refactorings to the compiler will be performed as part of this RFC. For example, the `GlobalVar` deduplication in `TECompilerImpl::LowerInternal` will be removed by changing the signature of `tvm::LowerSchedule`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- There might be cases when creating new `GlobalVars` without using a `NameSupply` might be needed. For example, in `IRModule::FromExprInContext`, the expression might already be annotated with a global symbol.
+We must be sure that the `NameSupply` interacts well with the `global_symbol` attribute.
```
mbs-octoml, line 89: I think since global_symbol is already endemic and a string we just rely on UniqueGlobalFor to recover the GlobalVar.

rfcs/0077_name_mangling_ir_modules.md
```diff
@@ -0,0 +1,102 @@
...
+
+# **Rationale and alternatives**
+An alternative way to unify name supplying and mangling in IR modules is to create a compiler pass that does the job. 
+The benefit of this is that the implementation would not be very intrusive. The downside is that it could only address name mangling inside IRModules. 
+There are cases when GlobalVars are created outside an IRModule, only to be later assigned to it. 
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Currently, the `NameSupply` cannot be accessed through FFI.
```
mbs-octoml, line 100: Suggest just making it an ObjectRef/Object. Then it can be a first class field of IRModule.

kparzysz-quic:
A couple of thoughts...

- We should have a common facility that represents a scope in which names need to be unique.  Conceptually it could be a set of "known" strings (that are presumed used), and something (like a private counter) to generate a unique suffix.  The use would be `new_name = GenerateName("foo")`.  If `foo` had already been generated, `new_name` would be set to `"foo4"` (for example).  This should not produce global variables (or any IR objects), just names.
- This RFC still doesn't settle the question whether "name_hint" is a final name or not.  I'm open to requiring that global symbols are given final names from the beginning, not hints.

kparzysz-quic (dismissed):
If nothing else, this should not generate new IR, only names.

kparzysz-quic:
`IRModule` already has functions to create new globals.  With the name-generating `NameSupply` added to the module, these functions could use that implicitly.

gigiblender:
> A couple of thoughts...
> 
>     * We should have a common facility that represents a scope in which names need to be unique.  Conceptually it could be a set of "known" strings (that are presumed used), and something (like a private counter) to generate a unique suffix.  The use would be `new_name = GenerateName("foo")`.  If `foo` had already been generated, `new_name` would be set to `"foo4"` (for example).  This should not produce global variables (or any IR objects), just names.
>
> 
>     * This RFC still doesn't settle the question whether "name_hint" is a final name or not.  I'm open to requiring that global symbols are given final names from the beginning, not hints.

Thanks for the comments @kparzysz-quic. I agree that we could have the `NameSupply` methods return Strings and make sure that `GlobalVars` are created using strings provided by a NameSupply (eventually one contained within the IRModule). This way, the `NameSupply` could be used for other name mangling scenarios where GlobalVars are not needed.

Regarding the `name_hint` being a final name. I am too of this opinion (it should be a final name) and believe that it already works this way. If I am not wrong, the only edge case is where we use the `global_symbol` attribute instead.

@mbs-octoml  @tqchen any thoughts?

rfcs/0077_name_mangling_ir_modules.md
gigiblender: Agreed

rfcs/0077_name_mangling_ir_modules.md
```diff
@@ -0,0 +1,102 @@
...
+The changes are internal to TVM. 
+The code writer is expected to avoid as much as possible calling the constructor of `GlobalVar` and instead use a `NameSupply` to generate one. 
+The `NameSupply` is constructed by passing a `String prefix` (usually a module name) that is to be prepended to the `GlobalVars` generated: `auto name_supply = NameSupply::NameSupplyWithPrefix(mod_name);`
+
+A `NameSupply` can be attached to an `IRModule`. The `IRModule` acts a carrier for the `NameSupply` through compiler passes. 
+The attached `NameSupply` can then be retrieved from the `IRModule` when, for example, a pass generates new `GlobalVars`.
+
+The `NameSupply` contains two methods to provide a `GlobalVar`:
+
+- `GlobalVar UniqueGlobalFor(String name, bool add_prefix)` performs a cache lookup and returns a `GlobalVar`. If a miss occurs, a new `GlobalVar` is created, inserted into the cache, and returned. The `add_prefix` boolean defaults to `true`. If it is `false`, then the `prefix_` field will not be added to the `GlobalVar`.
```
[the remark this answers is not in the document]
gigiblender: I think that's certainly something worth exploring. I can add that during the implementation if it proves necessary.

areusch:
@tqchen i think one goal we should have with this RFC is to codify things like
> When a function does not have global_symbol attribute, it means the LinkageType is private

the purpose is not to codify it so that we can't change it, but just so that we are all on the same page here. perhaps we can incorporate definitions like this into the RFC text and into helper functions?

areusch:
i've put this RFC on the meeting agenda for this week's Community Meeting

tqchen:
> the purpose is not to codify it so that we can't change it, but just so that we are all on the same page here. perhaps we can incorporate definitions like this into the RFC text and into helper functions?

I agree

gigiblender:
Adding a [draft PR](https://github.com/apache/tvm/pull/12066) to this discussion.

rfcs/0077_name_mangling_ir_modules.md
```diff
@@ -29,20 +30,31 @@ This RFC aims to unify the creation of unique `GlobalVars` and refactor the curr
...
+The code writer is expected to avoid as much as possible calling the constructor of `GlobalVar` and instead use a `GlobalVarSupply` to generate one:
+- The `GlobalVarSupply` is constructed by passing a `NameSupply` that is used to generate unique strings for the `name_hint` member of `GlobalVars`.
+Alternatively, builder methods such as `GlobalVarSupply::EmptySupply` can be used to construct an empty `GlobalVarSuppyl` with an empty `NameSupply`.
+- The `NameSupply` can be constructed by passing a `String prefix` (usually a module name) that can then be prepended  to the `String`s generated: `auto name_supply = NameSupply::NameSupplyWithPrefix(mod_name);`
+
+A `GlobalVarSupply` can be derived from an existing `IRModule` or an array of `IRModules` by calling `tvm::BuildGlobalVarSupply(modules)`. 
+When a pass generates new `GlobalVars`, a `GlobalVarSupply` can be created from the current `IRModule`. Then, `GlobalVarSupply::FreshGlobal`
+can be used to guarantee uniqueness of new `GlobalVars`.
+```
+GlobalVarSupply var_supply = tvm::BuildGlobalVarSupply(mod);
```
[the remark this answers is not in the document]
gigiblender: That sounds right.

areusch:
we discussed this at the [community meeting](https://discuss.tvm.apache.org/t/next-tvm-community-meeting-july-13/13120/2) yesterday. here are notes:

- @areusch: one of suggestions raised was to make it possible to reconstruct a NameSupply from an IRModule. we might need to make it possible to initialize it from multiple IRmodules since we currently lower Relay to many separate IRModules in the compiler.
- @kparzysz-quic : what's the use of having multiple modules, unless they're built for separate devices. do we need `mod_name` as a field?
  - `mod_name` is not to distinguish different IRModules at compile time; more to distinguish separate models at runtime via namespace prefix.
- @kparzysz-quic : what is the connection between the name that we attach to a global and the name that actually ends up in an object file? there is a strong need to have a degree of control over the names that end up in the object file and their visibility. how are we addressing this need? i think there is a lack of this in TVM now. we don't have a way
  - @tqchen: good point. we should document the linkage type. right now there is an implicit convention: the `name_hint` is not final (it's just an internal-to-compiler concept) and there's a GlobalVar which enforces that there is external linkage type with a fixed name. e.g. if we are splitting the compilation flow, then there is an expectation that when calling into a symbol, the name won't change. another rule: if a function has attribute `kGlobalSymbol`, you cannot change it or even delete it.
  - @manupa-arm : when we say we want to have control: do we want this to extend to Relay? if a Relay module contains a GlobalVar, should that be emitted with exactly that name?
    - @tqchen : if an IRModule has a function with attr `kGlobalSymbol` we should respect that. but as a rule of thumb we should delay attaching this attr til as late of possible, so we should try to avoid that symbol.
    - @areusch : we also don't have this ability in scheduling; it would be hard to attach that attribute to a Relay function e.g. main because we prefix lowered functions
    - @tqchen: we could special-case an AOT entry function when the compiler takes special care around a symbol.
    - filed https://github.com/apache/tvm/issues/12098 to document the convention.
 - @areusch some context for this RFC: in the C++ runtime, we previously mostly cared that a GlobalVar name could be passed to Module.GetFunction() and then that function would find that implemented PrimFunc. When we implemented AOT in the C runtime, we changed to expecting to be able to directly call those names as C symbols. This caused a bit of a problem when we moved to implement both the C and C++ AOTExecutor wrappers, where metadata lookups could be different because of name mangling. To solve that, we intended to move in a direction where the names of all symbols were fixed at birth, thus NameSupply.
   - @kparzysz-quic: who is typically the caller of these functions?
   - @areusch typically an executor
   - @kparzysz-quic can we modify the names of the global? shouldn't matter so long as everyone agrees

areusch:
@gigiblender is this ready for a last review?

gigiblender:
> @gigiblender is this ready for a last review?

@areusch yes

areusch (approved):

areusch:
thanks @gigiblender , the RFC has been accepted! please create a [tracking issue](github.com/apache/tvm/issues) and mention this PR in that issue to link it.

PR #87, opened 2022-08-09
Amend RFC0070 with DeclBuffer TVMScript syntax updates

vinx13:
As a followup on the previous discussion https://github.com/apache/tvm-rfcs/pull/70#issuecomment-1149308146, this PR updated the DeclBuffer RFC with TVMScript related changes.

cc @Lunderberg @wrongtest-intellif @Hzfengsy @junrushao1994 @cyx-6 @tqchen

### DIFF
--- a/rfcs/0070-introducing-decl-buffer.md
+++ b/rfcs/0070-introducing-decl-buffer.md
@@ -149,17 +149,17 @@ with flattened indices.
 * New statement `T.decl_buffer` will be introduced. It has the same interface as `T.buffer_decl`.
 ```python
 def decl_buffer(
-    shape,
-    dtype="float32",
-    data=None,
-    strides=None,
-    elem_offset=None,
-    scope="global",
-    align=-1,
-    offset_factor=0,
-    buffer_type="default",
-    axis_separators=None) -> Buffer:
-    ...
+    shape: Sequence[Union[PrimExpr, int]],
+    dtype: str = "float32",
+    data: Var = None,
+    strides: Optional[Sequence[int]] = None,
+    elem_offset: Optional[int] = None,
+    scope: str = "global",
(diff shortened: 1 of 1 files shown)


Hzfengsy (approved):

rfcs/0070-introducing-decl-buffer.md
```diff
@@ -146,9 +146,36 @@ Specifically, the updated flow of buffer flattening using `DeclBuffer` will be:
 with flattened indices.
 
 ## TVM script updates
+* New statement `T.decl_buffer` will be introduced. It has the same interface as `T.buffer_decl`.
+```python
+def decl_buffer(
+    shape: Sequence[Union[PrimExpr, int]],
+    dtype: str = "float32",
+    data: Var = None,
```
areusch, line 154: just one note--usually this is first if positional args are used (e.g. in the example above), is the order of these kwargs correct?

areusch (approved):
thanks @vinx13 !

rfcs/0070-introducing-decl-buffer.md
vinx13: I'm keeping the order the same as the existing `T.buffer_decl`

areusch: ok, but in the example above (line 135) you have:
```
A_flattened = T.decl_buffer(A.data, (256,), "float32")
```

this says to me that:
- `shape` == `A.data`
- `dtype` == `(256,)`
- `data` == `"float32"`

could you keep the RFC consistent?

vinx13: Thank you for pointing out, I've updated the RFC

areusch:
lgtm!

PR #97, opened 2023-01-08
[RFC] Further Unify Packed and Object in TVM Runtime

tqchen:
This RFC proposes to further unify our PackedFunc and Object in TVM Runtime. The proposal builds on top of our past lessons and recont lessons from related project such as [matxscript](https://github.com/bytedance/matxscript)

The key improvements include: unifying type_code, solidifying AnyValue support for both stack and object values, open doors for small-string and NLP-preprocessing, and enable universal container.

[rendered](https://github.com/tqchen/tvm-rfcs/blob/unify-packed/rfcs/0097-unify-packed-and-object.md)

### DIFF
--- a/rfcs/0097-unify-packed-and-object.md
+++ b/rfcs/0097-unify-packed-and-object.md
@@ -11,51 +11,51 @@ This RFC proposes to further unify our PackedFunc and Object in TVM Runtime. Key
 
 ## Motivation
 
-FFI is one of the main components of the TVM. We use PackedFunc convention to safely type-erase values and pass things around. In order to support a general set of data structures both for compilation purposes, we also have an Object system, which is made to be aware in the Packed API. 
+FFI is one of the main components of the TVM. We use PackedFunc convention to safely type-erase values and pass things around. In order to support a general set of data structures both for compilation purposes, we also have an Object system, which is made to be aware in the Packed API.
 
 Object supports reference counting, dynamic type casting, and checking as well as structural equality/hashing/serialization in the compiler.
 Right now, most of the things of interest are Object, including containers like Map, Array. PackedFunc itself, Module, and various IR objects.
-Object requires heap allocation and reference counting, which can be optimized through pooling. They are suitable for most of the deep learning runtime needs, 
+Object requires heap allocation and reference counting, which can be optimized through pooling. They are suitable for most of the deep learning runtime needs,
 such as containers, as long as they are infrequent.
-In the meantime, we still need to operate with values on the stack. Specifically, when we pass around int, and float values. 
+In the meantime, we still need to operate with values on the stack. Specifically, when we pass around int, and float values.
 It can be wasteful to invoke heap allocations/or even pooling if the operations are meant to be low cost. As a result, the FFI mechanism also serves additional ways to be able to pass **stack values** directly around without object.
 
 This post summarizes lessons from us and other related projects and needs around the overall TVM FFI and Object system. And seek to use these lessons to further solidify the current system. We summarize some of the needs and observations as follows:
 
 ### N0: First class stack small string and AnyValue
 
-Data preprocessing is an important part of ML pipeline. Preprocessing in NLP involves strings and containers. Additionally, when translating programs written by users (in python), there may not be sufficient type annotations. 
+Data preprocessing is an important part of ML pipeline. Preprocessing in NLP involves strings and containers. Additionally, when translating programs written by users (in python), there may not be sufficient type annotations.
 
 The programs below comes from real production scenario code from matxscript in NLP Preprocessing:
 
 ```cpp
-// This can be part of data processing code translated 
+// This can be part of data processing code translated
 // from user that comes without type annotation
 AnyValue unicode_split_any(const AnyValue& word) {
   List ret;
   for (size_t i = 0; i < word.size(); ++i) {
      AnyValue res = word[i];
-     ret.push_back(res);   
+     ret.push_back(res);
   }
   return ret;
 }
 // This is a better typed execution code
-// Note that word[i] returns a UCS4String container to match python semantics 
+// Note that word[i] returns a UCS4String container to match python semantics
 // Use UCS4String stores Unicode in a fixed-length 4 bytes value to ease random
-// access to the elements. 
+// access to the elements.
 List<UCS4String> unicode_split(const UCS4String& word) {
   List<UCS4String> ret;
   for (size_t i = 0; i < word.size(); ++i) {
      UCS4String res = word[i];
-     ret.push_back(res);   
+     ret.push_back(res);
   }
   return ret;
 }
 ```
-We would like to highlight a few key points by observing the above programs: 
+We would like to highlight a few key points by observing the above programs:
 - Need a base AnyValue to support both stack values and object.
     - This is to provide a safety net of translation.
-- The AnyValue needs to accommodate small-string(on stack) to enable fast string processing. Specifically, note that the particular example creates a `UCS4String res` for every character of the word. If we run heap allocation for each invocation, or even do reference countings, this can become expensive. The same principle also generalizes to the need to accommodate fast processing of other on-stack values. 
+- The AnyValue needs to accommodate small-string(on stack) to enable fast string processing. Specifically, note that the particular example creates a `UCS4String res` for every character of the word. If we run heap allocation for each invocation, or even do reference countings, this can become expensive. The same principle also generalizes to the need to accommodate fast processing of other on-stack values.
 
 
 While it is possible to rewrite the program through stronger typing and get more efficient code. It is important to acknowledge the need to efficient erased runtime support (with minimum overhead), especially given many ML user comes from python.
@@ -64,7 +64,7 @@ While it is possible to rewrite the program through stronger typing and get more
 
 In the above example, it is important to note that the container `List` should hold any values. While it is possible to also provide different variant of specialized containers(such as `vector<int>`), to interact with a language like python, it would be nice to have a single universal container across the codebase. We also experienced similar issues in our compilation stack. As an example, while it is possible to use Array to hold IR nodes such as Expr, we cannot use it to hold POD int values, or other POD data types such as DLDataType.
 
-Having an efficient universal container helps to simplify conversions across language as well. For example, a list from python will be able to be turned into a single container without worrying about content type. The execution runtime will also be able to directly leverage the universal container to support all possible cases that a developer might write. 
+Having an efficient universal container helps to simplify conversions across language as well. For example, a list from python will be able to be turned into a single container without worrying about content type. The execution runtime will also be able to directly leverage the universal container to support all possible cases that a developer might write.
 
 ### N2: Further Unify POD Value, Object and AnyValue
 
@@ -76,7 +76,7 @@ Both TVMRetValue and Object leverages a code field in the beginning of the data 
 There are two interesting regimes of operation that comes with ObjectRef and AnyValue.
 
 - R0: On one hand, if we are operating on the regime of no need for frequent stack value operations. It is desirable to simply use Object. Because object is more compact on register (the size of ptr, which costs 8 bytes on modern 64 bit machines and 4 bytes on 32 bit machines), it can obtain underlying container pointers easily for weak references
-    
+
     ```cpp
     void ObjectOperation(ObjectRef obj) {
       if (auto* IntImmNode int_ptr = obj.as<IntImmNode>()) {
@@ -84,10 +84,10 @@ There are two interesting regimes of operation that comes with ObjectRef and Any
       }
     }
     ```
-    
+
 - R1: On the other hand, when we operate on frequent processing that is also not well-typed (as the `unicode_split` example). It is important to also support a AnyValue that comes with stack value support.
 
-As a point of reference, python use object as base for everything. But that indeed creates the overhead for str, int (which we seek to eliminate). Java and C# support both stack values, and their object counter part. 
+As a point of reference, python use object as base for everything. But that indeed creates the overhead for str, int (which we seek to eliminate). Java and C# support both stack values, and their object counter part.
 Right now we have both mechanism. It would be **desirable to further unify the Object and AnyValue** to support both R0 and R1. Additionally, it would be nice to have automatic conversions if we decide that two mechanisms are supported. Say a caller pass in a boxed int value, the callee should be able to easily get int out from it(or treat it as an int) without having to do explicit casting. So the same routine can be implemented via either R0 or R1 that is transparent to the caller.
 
 - This is also important for compilers and runtimes, as different compiler and runtime might have their own considerations operating under R0/R1.
@@ -99,7 +99,7 @@ We have the following design goals:
 - G0: Automatic switching between object focused scenario and stack-mixed that requires AnyValue.
 - G1: Enable efficient string processing, specifically small-string support for NLP use-cases.
 - G2: Enable efficient universal container (e.g common container for List/Array that stores everything).
-  - Note that it does not prevent us to create specalized code such as `List<String>` as java do, except that 
+  - Note that it does not prevent us to create specalized code such as `List<String>` as java do, except that
     they still share the same underlying container.
   - Array will share the same container with List to avoid conversion cost.
 - G3: Reduce concept duplication(type_code) and provide an unify approach for POD values and object values(including boxing and unboxing)
@@ -112,13 +112,13 @@ AnyValue unicode_split_any(const AnyValue& word) {
   for (size_t i = 0; i < word.size(); ++i) {
      // efficient small string support
      AnyValue res = word[i];
-     ret.push_back(res);   
+     ret.push_back(res);
   }
   return ret;
 }
 
 // Unify object and POD value handling
-// passing an boxed int object to int function and get out int 
+// passing an boxed int object to int function and get out int
 // automatically without conversion
 int MyIntFunc(AnyValue x) {
   int xval = x;
@@ -149,11 +149,11 @@ struct Object {
   // 4 bytes type code
   // This is a common header with AnyPODBase_
   int32_t type_code;
-  // 4 bytes ref counter 
+  // 4 bytes ref counter
   RefCounterType<int32_t> ref_counter;
   // 8 bytes deleter
   typedef void (*FDeleter)(Object* self);
-  FDeleter deleter; 
+  FDeleter deleter;
   // Rest of the sections.
 };
 
@@ -188,7 +188,7 @@ struct AnyView: public AnyPodBaseValue_ {
 
 // An any value with extra padding data
 // can be used to store larger small str
-template<int num_paddings> 
+template<int num_paddings>
 struct AnyPad : public AnyValue {
   union {
     char v_pad_bytes[num_paddings * 8];
@@ -198,7 +198,7 @@ struct AnyPad : public AnyValue {
 };
 ```
 
-This is a design that outlines the key terms 
+This is a design that outlines the key terms
 
 - T0: Object: the intrusive ptr managed object, used by most containers
     - This is the same as the current object, we list here for clarity.
@@ -208,10 +208,10 @@ This is a design that outlines the key terms
 - T3: AnyPad: an any value that have larger padded size to accomodate on stack values.
     - When the initial value defaults to null. Both AnyValue and AnyPad, can choose to fill the small_len to be the size of total bytes available. This can help us to be able to pass small string back in C API (without template), by looking at `AnyValue*` ’s small_len field to decide the maximum bytes allowed.
 
-**Discussions**  The default size of AnyValue is 16 bytes. This means that for small string, we can use extra 8 bytes to store the string part(7 bytes if we need a tail `\0`). If we go with UCS4, we can store two extra UCS4 Char without the tail `\0`. The extra space may not be sufficient for some of the small string needs (as a reference matxscript adopts extra padding of 8 bytes to accommodate small string unicode). AnyPad serves as another variation of AnyValue that contains extra stack space. AnyPad is intended to be used interchangeably in any places that AnyValue appears. See also followup sections on conversions function signatures on how that works. One interesting future direction point here is that future compilers can choose to try different AnyPad in code generation and autotune the padding default to the scenario that best fits the application.
+**Discussions**  The default size of AnyValue is 16 bytes. This means that for small string, we can use extra 8 bytes to store the string part(7 bytes if we need a tail `\0`). If we go with UCS4, we can store two extra UCS4 Char without the tail `\0`. The extra space may not be sufficient for some of the small string needs (as a reference matxscript adopts extra padding of 8 bytes to accommodate small string unicode). AnyPad serves as another variation of AnyValue that contains extra stack space. AnyPad is intended to be used interchangeably in any places that AnyValue appears. See also followup sections on conversions function signatures on how that works.
 
 ```cpp
-// This can be part of data processing code translated 
+// This can be part of data processing code translated
 // from user that comes without type annotation
 AnyValue unicode_split_any(const AnyValue& word) {
   List ret;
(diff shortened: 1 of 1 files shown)


tqchen:
cc @apache/tvm-committers for visibility

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+    - Note that this implies (by intention) change the ABI of packed func. We will update compiler/runtime to do so.
+    - We will introduce adapter to support the TVMArgs during the transition, but favors moving to new state to reduce overall complexity.
+- M1: Introduce new string support (with small string)
+- M2: Introduce universal container
+
+We believe the overall milestones are positive given the net gain obtained to enable pre-processing and stronger interpolation with ML ecosystem and the community as well. It also opens doors to bring in features in projects like matx so we can enable efficient NLP preprocessing together with ML workload in the same pipeline.
+
+Additionally, Unifying FFI and Object would bring further unification and reduction in our overall code complexity while leveling up the extensibility, so it serves as a strong improvement to the overall quality of the project. 
+
+## Drawbacks
```
leandron, line 590: Can you comment on _how_ specifically the proposed change impacts the TVM runtime, e.g. is this a breaking change in some aspect?

leandron (commented):
Hi @tqchen et al. This change seems positive as consolidating the generic object type. I've got a few minor points and I guess my main question which asking for a slight expansion on how the change is specifically expected to impact TVM runtimes.

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+D3b may have overhead for small-str and POD. D3a and D3c are small-str friendly.
+
+### D4: PackedFunc Convention
+
+In this section, we revisit the PackedFunc convention under the new context. The update to the C++ PackedFunc API will run as follows:
+
+- TVMArgs now stores ptr of AnyView, TVMArgs[i] will now return AnyView
+- Alias: TVMRetValue = AnyValue, TVMArgValue = AnyView
+
+These changes are invisible to the users as long as they use the same source library. We will immiedately gain the ability to do universal switching when defining PackedFunc.
```
Mousius, line 463:
```suggestion
These changes are invisible to the users as long as they use the same source library. We will immediately gain the ability to do universal switching when defining PackedFunc.
```

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+
+- **D4a:** Current C API:
+    
+    ```cpp
+    int TVMCPackedFunc(PackedFuncHandle handle,
+                       int num_args, int* type_codes, TVMValue* values,
+                       int* out_tcode, TVMValue* out_value);
+    ```
+    
+    - This would require packing unpacking the typecode. Note that small string passing wont work because of the lack of the seq_len padding field.
```
Mousius, line 480:
```suggestion
    - This would require unpacking the typecode. Note that small string passing wont work because of the lack of the seq_len padding field.
```

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+    
+    // TVMAnyView and TVMAnyValue follows the same layout as AnyPodBaseValue_
+    int TVMCPackedFunc(PackedFuncHandle handle,
+                       int num_args, TVMAnyView* args, 
+                       TVMAnyValue* out);
+    ```
+    
+    This approach combines code and value into a single entity, this would mean a change of ABI convention in generated code. This approach makes it possible to directly return a small-str without boxing it (however if it is faced in python frontend, likely we still need to box it to str).
+    
+    We will introduce adapters to support the old calling convention, which constructs TVMAnyView on the stack then pass things over in the transitioning period.
```
Mousius, line 493: What's the planned naming conventions for this and how do they eventually get transitioned to just being `TVMCPackedFunc` again?

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+
+### D0a and D0b: Default size of AnyValue
+
+The default size of AnyValue
+
+- **D0a**: 16 bytes
+    - AnyView also cost 16bytes. Use AnyPad<n> in locals when necessary
+- **D0b**: 24 bytes
+
+Both choices are likely OK and won’ have big impact. Because we support AnyPad<n> natively in locals (and can also pass AnyPad<n> around) by taking address of it, we would recommend start with D0a — this also keeps the overall call stack cost consistent with the current design.
```
Mousius, line 512:
```suggestion
Both choices are likely OK and won’t have big impact. Because we support AnyPad<n> natively in locals (and can also pass AnyPad<n> around) by taking address of it, we would recommend start with D0a — this also keeps the overall call stack cost consistent with the current design.
```

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+
+- **D1a**: Stores type_code if the object is in S1
+- **D1b**: Stores all type_code (S1-S4)
+- **D1c**: Store type of only static segment of types. (S1-S3 but not S4).
+
+The current state starts with D1a.  The tradeoff here is again not as critical. One thing to consider is the object code lookup cost when the frontend only recognizes part of the type. Likely starting from D1a is a reasonable pt.
+
+### D1section: Any.type_code section convention
+
+When to store non-object type code in AnyValue.type_code. We hae the following code segmentation organization
```
Mousius, line 526:
```suggestion
When to store non-object type code in AnyValue.type_code. We have the following code segmentation organization 
```

Mousius (changes requested):
Hi @tqchen et al,

Looks like a good unification which will hopefully solve a bunch of issues, one specific one I'd like to highlight for you is where we're getting the wrong types out of the type system in tvmc (`IntImm` is both `int` and `bool`):
https://github.com/apache/tvm/blob/main/python/tvm/driver/tvmc/target.py#L37-L41
Am I assuming correctly that by using AnyValue and allowing it to map to the correct Python type we'll be able to properly infer the values here?

I've posted a few questions alongside a general suggestion that we orient this towards improving TVM with a reference to `matxscript` kept in Prior Art for reference rather than through-out. Otherwise most of my comments are minor spelling mistakes - it might be worth running this through a spell checker?

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+Object requires heap allocation and reference counting, which can be optimized through pooling. They are suitable for most of the deep learning runtime needs, 
+such as containers as long as they are infrequent.
+In the meantime, we still need to operate with values on stack. Specifically, when we pass around int, float values. 
+It can be wasteful to invoke heap allocations/or even pooling if the operations is meant to be low cost. As a result, the FFI mechanism also serves additional ways to be able to pass **stack values** directly around without object.
+
+This post summarizes lessons from us and other related projects and needs around the overall TVM FFI and Object system. And seek to use these lessons to further solidify the current system. We summarize some of the needs and observations as follows:
+
+### N0: First class stack small string and AnyValue
+
+**Lesson from matxscript:** Data preprocessing is an important part of ML pipeline. Pre-processing in NLP involves strings and containers. Additionally, when translating programs written by users (in python), there may not be sufficient type annotations. We can common get the one of the programs below
```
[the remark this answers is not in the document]
tqchen: This particular lesson do come from matxscript more significantly :) Including the unicode example. So I think it is approperiate to cite in both places

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+     UCS4String res = word[i];
+     ret.push_back(res);   
+  }
+  return ret;
+}
+```
+
+- Need a base AnyValue to support both stack values and object.
+    - This is to provide a safety net of translation.
+- The AnyValue needs to accommodate small-string(on stack) to enable fast string processing. Specifically, note that the particular example creates a `UCS4String res` for every character of the word. If we run heap allocation for each invocation, or even do reference countings, this can become expensive.
```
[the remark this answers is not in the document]
tqchen: Yes This does generalizes to need to support stack values. 
However, that also depends on whether such scenarios are useful in the settings we consider. In this particular case. We have not yet seen needs on other stack values yet. But the design naturally generalizes to stack values

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+     AnyValue res = word[i];
+     ret.push_back(res);   
+  }
+  return ret;
+}
+// This is a better typed execution code
+// Note that word[i] returns a UCS4String container to match python semantics 
+// Use UCS4String stores unicode in a fixed-length 4 bytes value to ease random
+// access to the elements. 
+List unicode_split(const UCS4String& word) {
```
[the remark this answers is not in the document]
tqchen: Yes, the code below, however, was mainly used for demonstration purposes about what can happen

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+  return ret;
+}
+```
+
+Both AnyValue and AnyView also have direct correspondence in the current codebase (TVMRetValue and TVMArgValue). We will use `AnyValue` and `AnyView` for consistency throughout this document. 
+
+**Default size of AnyValue** Any variant of AnyPad can be used as default size of AnyValue. For example, we list the following design choices
+
+- **D0a** Default to AnyPad<0> aka 16 bytes. The advantage is smaller size overall in default parameter passing.
+- **D0b** Default to AnyPad<1> aka 24 bytes. According to matx’s experience, AnyPad<1> serves well for bytedance’s internal NLP processing needs. However that was also before we had the extra AnyPad proposal. It is now possible to have AnyValue default to 16 bytes, while still create AnyPad during intermediate execution.
```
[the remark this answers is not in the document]
tqchen: I think it is useful to put the reference under the context, so we understand the design tradeoffs here

rfcs/0097-unify-packed-and-object.md
tqchen: The naming convention is being as proposed TVMAnyValue. The particular signature is a C signature for the frontend API.

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+- M1: Introduce new string support (with small string)
+- M2: Introduce universal container
+
+We believe the overall milestones are positive given the net gain obtained to enable pre-processing and stronger interpolation with ML ecosystem and the community as well. It also opens doors to bring in features in projects like matx so we can enable efficient NLP preprocessing together with ML workload in the same pipeline.
+
+Additionally, Unifying FFI and Object would bring further unification and reduction in our overall code complexity while leveling up the extensibility, so it serves as a strong improvement to the overall quality of the project. 
+
+## Drawbacks
+
+The design proposal would involve changes in our runtime system. This is however, a positive step to further solidy and reduce the overall amount of concepts in the code-base, further unify packed and object, simplify and solidify our implementation along side of AnyValue and Object. Finally, it also opens doors for NLP applications and cross project interactions.
```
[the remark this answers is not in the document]
tqchen: This does affects and cover all exsting tvm runtimes. We intended to do a refactor for all of them for consistency

rfcs/0097-unify-packed-and-object.md
tqchen: Please see M0 section in the Phasing part. WIll also add a reference here:

- Note that this implies (by intention) change the ABI of packed func. We will update compiler/runtime to do so.
- We will introduce adapter to support the TVMArgs during the transition, but favors moving to new state to reduce overall complexity.

More details can be found in D4

rfcs/0097-unify-packed-and-object.md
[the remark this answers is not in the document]
tqchen: will add a note here

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+    - When the initial value defaults to null. Both AnyValue and AnyPad, can choose to fill the small_len to be the size of total bytes available. This can help us to be able to pass small string back in C API (without template), by looking at `AnyValue*` ’s small_len field to decide the maximum bytes allowed.
+
+**Discussions**  The default size of AnyValue is 16 bytes. This means that for small string, we can use extra 8 bytes to store the string part(7 bytes if we need a tail `\0`). If we go with UCS4, we can store two extra UCS4 Char without the tail `\0`. The extra space may not be sufficient for some of the small string needs (as a reference matxscript adopts extra padding of 8 bytes to accomodate small string unicode). AnyPad serves as another variation of AnyValue that contains extra stack space. AnyPad is intented to be used interchangably in any places that AnyValue appears. See also followup sections on conversions function signatures on how that works. One interesting future direction point here is that future compilers can choose to try different AnyPad in code generation and autotune the padding default to the scenario that best fit the application.
```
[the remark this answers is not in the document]
tqchen: To make sure the reader can get most of the information, we beleive it is helpful to also put these information in context of discussion, along with the related sections. This is also a common practice in scientific writings

rfcs/0097-unify-packed-and-object.md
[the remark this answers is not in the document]
tqchen: Rephrased here

rfcs/0097-unify-packed-and-object.md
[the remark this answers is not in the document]
tqchen: added a note about this

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+  List ret;
+  for (size_t i = 0; i < word.size(); ++i) {
+     UCS4String res = word[i];
+     ret.push_back(res);   
+  }
+  return ret;
+}
+```
+
+- Need a base AnyValue to support both stack values and object.
```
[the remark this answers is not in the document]
tqchen: Added additional explainations

rfcs/0097-unify-packed-and-object.md
tqchen: Added an expanded explaination

rfcs/0097-unify-packed-and-object.md
tqchen: Thanks for the suggestions, updated accoridngly

rfcs/0097-unify-packed-and-object.md
tqchen: Thanks for the suggestions, updated accoridngly

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+
+Having an efficient universal container helps to simplify conversions across language as well. For example, list from python will be able to be turn into a single container without worrying about content type. The execution runtime will also be able to directly leverage the universal container to support all possible cases that a developer might write. 
+
+### N2: Further Unify POD Value, Object and AnyValue
+
+TVM currently do have an AnyValue. Specifically `TVMRetValue` is used to hold managed result for C++ PackedFunc return can serve as any value. Additionally, if the value is object. `ObjectRef` serves as a nice way that comes with various mechanisms, including structural equality hashing.
+If we create Boxed Object for each stack values, e.g. Integer to represent int. We will be able to effectively represent every value in Object as well.
```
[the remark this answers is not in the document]
tqchen: Added a reference

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+
+- This is also important for compilers and runtimes, as different compiler and runtime might have their own considerations operating under R0/R1.
+
+## Guide-level explanation and Design Goals
+
+We have the following design goals:
+
+- G0: Automatic switching between object focused scenario and stack-mixed that requires AnyValue.
+- G1: Enable efficient string processing, specifically small-string support for NLP use-cases.
+- G2: Enable efficient universal container (e.g. Array that stores everything).
```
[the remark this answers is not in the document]
tqchen: The intention is to have both Array and List follow the same container, but still enables specialization

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+    - If T contains other objects, check and convert.
+
+The conversion from Object to AnyValue(which can then be converted to AnyView) can have two possibilities:
+
+- **D2a:** Simply keep object as they are when writing to AnyValue ****
+    - This simplifies conversion from Object to AnyValue. But when we convert Any into POD values, we will need to check whether if it is Boxed.
+- **D2b:** Always unbox to the POD value if object is a boxed value.
+    - This simplifies conversion of AnyValue into POD, since there is no need to check for boxed values.
+
+We encourage D2b when possible, this is because such conversion can be simplified in assignment. It also can help to simplify compiler side logic which only looks at POD type code but cannot handle the ObjectBoxing.
```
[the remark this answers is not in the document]
tqchen: Updated per suggestion

tqchen:
Thanks @leandron @Mousius for helpful suggestions in readability, we have incorporated most of them.  We would like still keep discussions of prior work design(in this case matx) both in prior Art as well as in the section of respective context to give clarity and context for the reader (this is also a common practice in technical scientific writing). Also updated the text with more clarifications on some of the questions asked.

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,677 @@
...
+  };
+  const uint32_t zero_ = 0;            // for c_str
+  int32_t category_or_small_len_ = 0;  // small_len: >= 0; large: -2, 
+};
+```
+
+Key elements include:
+
+- There is a zero field to enable `\0` paddings for small-str
+- The category_or_small_len field is stored in the end, to accommodate the zero padding
```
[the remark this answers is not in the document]
tqchen: updated the text here. category means whether it is a small or big string

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,712 @@
...
+    We will introduce adapters to support the old calling convention, which constructs TVMAnyView on the stack then pass things over in the transitioning period. PackedFunc will continue to support the existing TVMArgs and TVMRetValue signature, which adapts 
+    to the new calling convention.
+    ```c++
+    class PackedFunc {
+     public:
+      // old convention
+      void CallPacked(TVMArgs args, TVMRetValue* rv);
+      // new convention
+      void CallPacked(int num_args, AnyView* args, AnyValue* out)
+    }
```
Mousius, line 510: Some rogue whitespace here.
```suggestion
    }
```

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,712 @@
...
+                       int num_args, TVMAnyView* args, 
+                       TVMAnyValue* out);
+    ```
+    
+    This approach combines code and value into a single entity, this would mean a change of ABI convention in generated code. This approach makes it possible to directly return a small-str without boxing it (however if it is faced in python frontend, likely we still need to box it to str).
+    
+    We will introduce adapters to support the old calling convention, which constructs TVMAnyView on the stack then pass things over in the transitioning period. PackedFunc will continue to support the existing TVMArgs and TVMRetValue signature, which adapts 
+    to the new calling convention.
+    ```c++
+    class PackedFunc {
```
Mousius, line 504: This works for C++, what about the C runtime and `TVMFuncCall` (https://github.com/apache/tvm/blob/main/include/tvm/runtime/c_runtime_api.h#L323-L324)?

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,712 @@
...
+    - When the initial value defaults to null. Both AnyValue and AnyPad, can choose to fill the small_len to be the size of total bytes available. This can help us to be able to pass small string back in C API (without template), by looking at `AnyValue*` ’s small_len field to decide the maximum bytes allowed.
+
+**Discussions**  The default size of AnyValue is 16 bytes. This means that for small string, we can use extra 8 bytes to store the string part(7 bytes if we need a tail `\0`). If we go with UCS4, we can store two extra UCS4 Char without the tail `\0`. The extra space may not be sufficient for some of the small string needs (as a reference matxscript adopts extra padding of 8 bytes to accommodate small string unicode). AnyPad serves as another variation of AnyValue that contains extra stack space. AnyPad is intended to be used interchangeably in any places that AnyValue appears. See also followup sections on conversions function signatures on how that works. One interesting future direction point here is that future compilers can choose to try different AnyPad in code generation and autotune the padding default to the scenario that best fits the application.
```
Mousius, line 211: The future reference is unnecessary to motivate the current work and you've copied it to Future Possibilities, therefore I still advocate for removing it here.

rfcs/0097-unify-packed-and-object.md
```diff
@@ -0,0 +1,712 @@
...
+
+- This is also important for compilers and runtimes, as different compiler and runtime might have their own considerations operating under R0/R1.
+
+## Guide-level explanation and Design Goals
+
+We have the following design goals:
+
+- G0: Automatic switching between object focused scenario and stack-mixed that requires AnyValue.
+- G1: Enable efficient string processing, specifically small-string support for NLP use-cases.
+- G2: Enable efficient universal container (e.g common container for List/Array that stores everything).
```
Mousius, line 101: It would be better to standardise on the container name rather than duplicate it and use up both names?

Mousius (changes requested):

rfcs/0097-unify-packed-and-object.md
tqchen: This is mainly considering the state of transition. Array currently in TVM is immutable(closer to tuple), where as List is muttable. So likely we will need to start with two container names (that shares the same underlying object). Then we can consider aliasing, consolidation or keep two versions based on the need.

rfcs/0097-unify-packed-and-object.md
tqchen: Great question. Updated the text to contain this part.

 Transition of `TVMFuncCall` also happens in two steps.
    - First step the frontend facing APIs such as `TVMFuncCall` will be kept the same, providing an adapter to call into
      the new convention under the hood.
    - Then we will update the frontend implementation, compiler, and runtime to match the new proposed convention.

rfcs/0097-unify-packed-and-object.md
tqchen: I removed that per suggestion

tqchen:
@Mousius to specifically answer your questions on IntImm for bool and int. Likely this is due to the way we use container and checkers. Right now IntImm is indeed designed to hold for int and bool. And we will need to specific type checking on dtype which might be missed in such a case. Differentiating bool and int in the special section will likely resolve the issue if we believe it is important to have special casing for bool

rfcs/0097-unify-packed-and-object.md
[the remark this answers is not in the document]
tqchen: updated

tqchen:
@leandron @Mousius please take another look, i believe all the outstanding comments are addressed

cyx-6 (approved):
LGTM! We are so excited to embrace a more unified and robust runtime object system in TVM. I believe this system will not only help address some current issues, but also be a solid foundation of our TVM Unity. And hope to have some contribution with our new runtime object system implementation. :)

tqchen:
ping @Mousius @leandron

Mousius:
@tqchen Ack. I'll take another look this week.

rfcs/0097-unify-packed-and-object.md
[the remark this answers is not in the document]
Mousius: Agree to disagree 😸

Mousius (approved):

Hzfengsy:
The comments so far seem have been addressed, would love to see if we have additional comments, and move forward on this

Hzfengsy:
Thanks for everyone's input. We are going to merge in 24 hours if there are no additional comments.

Hzfengsy:
This RFC is merged now. Thanks @tqchen for the proposal and the reviews from @leandron @Mousius @cyx-6

PR #101, opened 2023-06-01
[RFC] TIR Language Specification

slyubomirsky:
[Rendered view.](https://github.com/slyubomirsky/tvm-rfcs/blob/tir-spec/rfcs/0101-tir-spec.md)

This RFC proposes including a language specification ([draft included](https://github.com/slyubomirsky/tvm-rfcs/blob/tir-spec/rfcs/assets/0101/spec.md)) for TIR in TVM's documentation.

Many thanks to all who reviewed and commented on earlier drafts (see the acknowledgements in the RFC text), especially @junrushao, @vinx13, and @Lunderberg who were my primary guides to TIR during the initial draft process.

### DIFF
--- a/rfcs/assets/0101/spec.md
+++ b/rfcs/assets/0101/spec.md
@@ -143,6 +143,7 @@ Attrs ::= Attrs(contents: {str: Object*})
 ```
 
 *Note that attributes and annotations can contain arbitrary TVM objects as values. These objects are used only at compile time.
+
 **Refers to one of the base classes in the TVM object representation. In practice, `ObjectRef`s are usually TIR AST nodes. Which ones are appropriate depend on the specific attributes (only those listed under the semantics have any visible effects; the rest are used only at compile time).
 
 Additionally, at run time, `PrimFunc`s take in parameters corresponding to buffers via the `DLPack` library's `DLTensor` class, defined in [dlpack.h](https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h):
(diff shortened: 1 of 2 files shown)


slyubomirsky:
Requesting reviews from some in the community who are knowledgeable about TIR: @Hzfengsy @tqchen @junrushao @Lunderberg @vinx13 @cyx-6 @MasterJH5574

I would also welcome suggestions on other TIR contributors whose attention should be drawn to this RFC.

Perhaps it might make sense to set up a channel on the TVM Discord to allow for some real-time discussion.

slyubomirsky:
@kparzysz-quic @yzh119 Perhaps you might also be interested in reviewing (feel free to tag others as well).

slyubomirsky:
I've added an update for recently added inter-`PrimFunc` calls, please have a look!

rfcs/assets/0101/spec.md
```diff
@@ -522,7 +528,7 @@ Unlike `PrimExprs`, statements do not return values. Instead, they operate by mo
...
         5. One further condition that `PrimFunc`s expect of their `DLTensor` arguments: No two `DLTensor` arguments are permitted to alias each other. 
-        6. Next, `body` is executed. The `PrimFunc` produces outputs by mutating values in buffers passed as the inputs; these changes can be observed by the caller via the `DLTensor` representations passed in step i.
+        6. Next, `body` is executed. The `PrimFunc` produces outputs by mutating values in buffers passed as the inputs; these changes can be observed by the caller via the `DLTensor` representations passed in step i. If a `tir.ret` builtin was executed, the `PrimFunc` returns the value that was passed to `tir.ret`; otherwise, the `PrimFunc` returns a void value. Upon returning (whether after encountering a call to `tir.ret` or the end of `body`), the current scope is popped and all values allocated within the `PrimFunc` are freed.
```
Lunderberg, line 531: Looking at it, I'm not sure on the "all values allocated within the PrimFunc are freed" is true.

If there is an `Allocate` node, it gets lowered to a matched pair of to the PackedFuncs registered as `"tir.TVMBackendAllocWorkspace"` and `"tir.TVMBackendFreeWorkspace"`.  If the body of the `Allocate` node contains an explicit return through `T.ret`, or an implicit return through an `Assert` node, then the free wouldn't occur.

Though, that's probably best considered a bug, and not intended behavior.

slyubomirsky: Since you don't think it's the intended behavior, I think we should call that a bug then. Happy to hear others' thoughts.

PR #105, opened 2023-09-15
[RFC][SYCL] add sycl backend RFC

DongBaiYue:
This RFC is to add a new backend language——SYCL. Previously I also created an RFC topic in[ the forum](https://discuss.tvm.apache.org/t/rfc-sycl-sycl-backend-for-tvm/15678).

### DIFF
--- a/rfcs/0105-sycl-backend.md
+++ b/rfcs/0105-sycl-backend.md
@@ -104,7 +104,7 @@ This RFC only adds the SYCL backend to TVM, no other features will be affected. 
 
 The added codegen and runtime should be compatible with the existing TensorIR infra.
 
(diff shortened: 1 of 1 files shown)


Hzfengsy:
@tqchen @masahi Please take a look if you are interested.

tqchen:
I think SYCL codegen is great, would be good to make the descripton centered around TensorIR, so we can leverage latest infra and prepare for support like foundational models

DongBaiYue:
> I think SYCL codegen is great, would be good to make the descripton centered around TensorIR, so we can leverage latest infra and prepare for support like foundational models

Yes, SYCL codegen takes TensorIR as input and is compatible with TensorIR infra.

tqchen (approved):

PR #108, opened 2024-04-11
[RFC] Add NNEF frontend

agoston-mc:
A RFC to add a Neural Network Exchange Format frontend to TVM relay. 
[Link to discuss](https://discuss.tvm.apache.org/t/rfc-0108-add-nnef-frontend/17088)

### DIFF
--- a/0108-add-nnef-frontend.md
+++ b/0108-add-nnef-frontend.md
@@ -1,0 +1,98 @@
+- Feature Name: `Relay NNEF frontend`
+- Start Date: 2024-04-11
+- RFC PR: [apache/tvm-rfcs#0108](https://github.com/apache/tvm-rfcs/pull/0108)
+- GitHub Issue: [apache/tvm#0000](https://github.com/apache/tvm/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add the Khronos Neural Network Exchange Format (NNEF) as a frontend to TVM.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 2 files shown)


tqchen:
Thanks for the proposal, as a community we recently moves towards the relax IR for latest genAI workloads, additionally, it is unclear how much adoption NNEF have as of now versus ONNX and other formats

tqchen:
Thanks for the note. We are in the process of revamping docs. The latest set of emerging model optimizations like LLMs will be based on relax. https://github.com/apache/tvm/tree/main/python/tvm/relax/frontend/onnx likely is a good reference there

agoston-mc:
We have updated the PR with Relax frontend, but we have also kept the Relay, as an option, thinking it could be useful to have both, because we noticed performance differences during testing.

We observed that Relax by the default build pipeline is significantly slower than Relay. (On CPU we observed 2 orders of magnitude slower runs, while on GPU 3-5x slower, the models we tested were mobilenet and resnet variants, all static models) We observed the same with the ONNX Relax frontend, so we suspect the issue is with the compilation, not with the frontends. Is this a normal situation with the current state of development of Relax? 
By using Meta Schedule with a custom pipeline (with only a ValidateOps transformation), we managed to match or surpass the speed of Relay, but in many cases using the 'zero' or 'default_build' pipelines did not improve the performance.
What is the recommended workflow to be able to reach the performance of Relay reliably (at least on static models)? 

Anyways this does not affect the frontend code in the PR so we could move on with that as the frontend is ready to be submitted to TVM. We are just curious for debug/measurement reasons, as we were surprised by the results.

tqchen:
I think the main reason here was relay default incorporate autotuning by default, while Relax dos not.  The main rationale as of now is we would like to choose to decouple metaschedule tuning from the flow. 

That does not mean metaschedule cannot be applied, we do encourage users to apply metaschedule for traditional applications. In the build flow, the meta-schedule can then get applied by composing together with default flow.

The zero pipeline as of now mainly focus on some of some extra out of box improvement for latest LLM models and can expand to more in future

gyenesvi:
Thanks for the info about the schedules and differences, it makes sense.

As for moving on, what would be the next step now? Do you need any other info from us for reviewing?

tqchen (approved):

tqchen:
Leaving it open for another week in case others want to chime in, otherwise LGTM

gyenesvi:
Great, thank you!

